PR #1, opened 2016-03-03
Upgrade the logic for parsing flags and add test.

Carrotman42:
Don't require -dir unless -fuse or -instances_metadata is used, or if
  -instances contains a unix socket.

Don't allow -instances_metadata unless running on GCE.

Give a nicer error if windows users try to use -instances.

Give a nicer message for systems which don't support FUSE and forgot to
  set -instances.

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -35,14 +35,14 @@ import (
 // local connections.  Values received from the updates channel are
 // interpretted as a comma-separated list of instances.  The set of sockets in
 // 'dir' is the union of 'instances' and the most recent list from 'updates'.
-func WatchInstances(dir string, instances []instanceConfig, updates <-chan string) (<-chan proxy.Conn, error) {
+func WatchInstances(dir string, cfgs []instanceConfig, updates <-chan string) (<-chan proxy.Conn, error) {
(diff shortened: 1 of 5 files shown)


cmd/cloud_sql_proxy/proxy.go
```diff
@@ -150,31 +137,177 @@ func listenInstance(dst chan<- proxy.Conn, dir, instance string) (net.Listener,
...
+			ret.Network = spl[0]
+			addr, ok := loopbackForNet[spl[0]]
+			if !ok {
+				return ret, fmt.Errorf("invalid %q: unrecognized network %v", instance, spl[0])
+			}
+			ret.Address = fmt.Sprintf("%s:%s", addr, spl[1])
+		case 3:
+			// User provided a host and port; use that.
+			ret.Network = spl[0]
+			ret.Address = fmt.Sprintf("%s:%s", spl[1], spl[2])
```
jasonjei: Got it. It looks like this makes #2 redundant with this change, right?  For maximum compatibility, would you want to do protocol:port(optional :address)? Or do you think it is more intuitive to do protocol:address:port? Thanks for adding in the option to specify what addresses to listen! Jason

Carrotman42: I think it's more intuitive to do network:host:port since host:port is the most common order when doing networking things.

PR #2, opened 2016-03-07
Added option to indicate which address to listen

jasonjei:
In addition to being able to use `./cloud_sql_proxy -dir=/cloudsql -instances=PROJECT:REGION:INSTANCE=tcp:PORT` functionality, I think it is important to be able to indicate which **ADDRESS** to listen with a `-instances=PROJECT:REGION:INSTANCE=tcp:PORT:ADDRESS` option.

The main use case is using with Docker and GKE. I'd like to be able to use TCP to simplify Docker Google Container Engine deployments and simply have my existing services look for the cloud SQL proxy service. 

So now, I can advertise a proxy service with the port 3306 open in my Kubernetes setup. The proxy uses 127.0.0.1, which I think is a good default, but for having a proxy available for use by other Kubernetes services, I'd need to advertise **0.0.0.0**.

The alternative options are not good. To use CloudSQL today with Google Container Engine, I would have to have my existing Docker containers to run an additional process (which is against the Docker philosophy of one container one process).  Alternatively, one GKE user has proposed using a Docker container called `cloudsqlip` which polls running GKE nodes every few seconds and adds new entries into the ACL via IP, which I also think is undesirable. 

I'd really prefer not to modify my existing Docker containers running on Google Container Engine to include the cloud_sql_proxy binary, so running an additional Docker container to access CloudSQL seems like a decent option. Isn't that the Docker way of doing things?

Let me know what you think.

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -122,18 +122,24 @@ func listenInstance(dst chan<- proxy.Conn, dir, instance string) (net.Listener, 
 	var path string
 	var l net.Listener
 	if eq := strings.Index(instance, "="); eq != -1 {
-		spl := strings.SplitN(instance[eq+1:], ":", 2)
+		spl := strings.SplitN(instance[eq+1:], ":", 3)
 		if len(spl) == 1 {
 			return nil, fmt.Errorf("invalid format in %q; expected 'project:instance=tcp:port'", instance)
 		}
 
 		instance = instance[:eq]
 
+		host := "127.0.0.1"
+		if len(spl) == 3 {
+			host = spl[2]
+		}
+
 		var err error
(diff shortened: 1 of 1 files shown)


Carrotman42:
Thanks for the pull request! I've actually implemented something similar in my pull request [0]. I don't have time to go through it right now, but hopefully the request will be merged by the end of the week.

[0] https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/1

Carrotman42:
I've merged the other pull request that does the same thing, so I'll close this one.

Thanks again for the contribution! If you have any other features/bugfixes/anything I'd be happy to accept them.

jasonjei:
No problem, thanks for making this utility and making the source available before the update was pushed. Sure beats managing ACLs!

PR #5, opened 2016-03-16
Switch sample to YAML.

dlorenc:
cc @jeffmendoza

Jeff, I switched the exmaples to YAML. I couldn't get "kubectl create secret" to work though:

$ kubectl create secret generic foobar --from-file=README.md
Error: unknown flag: --from-file

I'm on kubectl 1.1.7, maybe that's the problem?


jeffmendoza:
Ah, yes, sorry I think that is a 1.2 feature.

LGTM

dlorenc:
Thanks! I'm looking forward to that feature, creating these secret files by hand isn't much fun.

cc @Carrotman42

PR #14, opened 2016-04-05
Use the 'goauth' library to get oauth tokens instead of doing it myself

Carrotman42:
This change automatically supports gcloud credentials as well as removes three mini-libraries that existed before I knew that library did everything I wanted for me.

(previously there was a disabled-by-constant way to do verbose logging, but I never used it so it doesn't really matter.)


Carrotman42:
I signed it! I work for Google already

PR #24, opened 2016-04-19
Add usage for specifying another host

minodisk:


minodisk:
I signed it!

README.md
```diff
@@ -57,6 +57,10 @@ JSON file. You may also set the GOOGLE_APPLICATION_CREDENTIALS environment varia
     ./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:3306 &
     mysql -u root -h 127.0.0.1
 
+    # For specifying another host:
+    ./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:0.0.0.0:3306 &
+    mysql -u root -h 0.0.0.0
```
Carrotman42: I don't think specifying `-h 0.0.0.0` is going to work. Maybe you meant 127.0.0.1 still (or the actual IP of the machine running the proxy?)

README.md
```diff
@@ -57,6 +57,10 @@ JSON file. You may also set the GOOGLE_APPLICATION_CREDENTIALS environment varia
     ./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:3306 &
     mysql -u root -h 127.0.0.1
 
+    # For specifying another host:
```
Carrotman42: Can you add a warning here that care should be taken when running this command? If executing on a system without a correctly configured firewall this will could potentially allow anything on the internet to access the database.

README.md
minodisk: `mysql -u root -h 0.0.0.0` works in this container: https://hub.docker.com/r/syoya/cloudsql-proxy/~/dockerfile/
And this container can be accessed from another container (like application container). When I specified `=tcp:127.0.0.1:3306`, the container wasn't accessible.

But this is hard to understand, so I will rewrite this line `mysql -u root -h [machine-ip]`.

README.md
minodisk: It was so right, I'll add a notification about the risk.

Carrotman42:
Thanks for the pull request! If you have any more documentation (or code!) suggestions I'll be happy to include them.

PR #25, opened 2016-04-19
Fix typo in accountErrorSuffix

petercm:
Correct the error message describing the use of the -credential_file
parameter. Previously it incorrectly suggested -credential**_s_**_file


Carrotman42:
Thanks!

PR #27, opened 2016-04-19
Respect credential_file and token args on GCE

### LINKED ISSUE #26: Unable to use -credential_file on GCE
This was working for me last week. I suspect it could be related to the scope check fix in #21 

The check for `onGCE` in `checkFlags` comes before anything would look at the tokenFile provided in `-credential_file` (or a `-token` or `GOOGLE_APPLICATION_CREDENTIALS` for that matter). So if the default service account does not have the `sqlservice.admin` scope those other auth methods will never be checked.

I'm running in Kubernetes and would prefer to only give access to the pods that need it.

petercm:
Skip permissions check for GCE VM service account when a credential_file
or token were passed as arguments. Fixes #26

This was the least invasive change I could make to address this. The content check on tokenFile and token is now done here and in `authenticatedClient(ctx context.Context)` but I'm not sure what the best way to refactor this would be.

I tested on a default GCE instance and got the expected error when passing no credential file and successfully connected when passing the file.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -176,6 +176,11 @@ func checkFlags(onGCE bool) error {
 		}
 		return nil
 	}
+
+	if *token != "" || *tokenFile != "" || os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != "" {
+		return nil
+	}
+
 	scopes, err := metadata.Scopes("default")
 	if err != nil {
 		if _, ok := err.(metadata.NotDefinedError); ok {


cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -176,6 +176,11 @@ func checkFlags(onGCE bool) error {
 		}
 		return nil
 	}
+
+	if tokF, tok := *tokenFile, *token; tokF != "" || tok != "" {
```
Carrotman42: Setting the GOOGLE_APPLICATION_CREDENTIALS environment variable is supposed to be equivalent to setting -credential_file. Can you also check if `os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")` is not ""?

Also a style nit: there's no need to store to local variables here. Just compare the values directly:

if *token != "" || *tokenFile != "" || os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != "" {

Carrotman42:
Whoops, sorry about that. Thanks for the report and the pull request! Just a small comment and I'll merge it in.

petercm:
All set

Carrotman42:
Thanks again!

Issue #28, opened 2016-04-19
Include Dockerfile for cloudsql-docker


psamaan:
I can't seem to find the Dockerfile or image components for the Docker image compiled from this repo and available at "b.gcr.io/cloudsql-docker/gce-proxy", so those of us using this on GCE as the only way to access Cloud SQL securely are unable to build an image from specific commits so we can for instance rollback to previous non-production breaking changes. Kindly share the Dockerfile or otherwise maintain a version system for the images corresponding to the commits on GitHub.

Carrotman42:
Ah yes, I'm surprised I actually didn't do that already.

By the way, sorry for breaking your production with the latest release :/ If you have a test environment that you would be able to test the master branch please let me know. I could help set anything up if it meant that releases would be more stable.

psamaan:
Thank you. It's okay about the latest release, I ended up hopping on a host running a working container, finding its orphaned image and tagging and pushing it to our own repo to use instead, so it worked out. I'll see what I can do about the testing environment. Thanks for the quick merge.

Carrotman42:
PTAL at https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/29; I decided against publishing a standalone Dockerfile which was 'FROM golang' and compiled the proxy on its own, but that would force end users to download the container containing the compiler and standard library even though it's unnecessary.

Let me know if you know of another way to create a standalone Dockerfile which does not end up basing itself off of another container (other than 'scratch'). Also feel free to make any comments on that pull request if you have anything to add/request.

Carrotman42:
Please reopen if https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/29 doesn't fully solve your problem or if you have more feature requests for building.

Carrotman42:
For version 1.05 I did a docker push without updating the 'latest' tag. You can try out the new version by pulling b.gcr.io/cloudsql-docker/gce-proxy:1.05

I will update the latest tag after a bit of time.

PR #29, opened 2016-04-19
Add support for building a docker container in build.sh

Carrotman42:
As requested in https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/28, work the logic for creating the docker container into the build script.

I could technically commit a standalone Dockerfile which does everything, but afaict the downside would be that the Go compiler and standard library would end up packaged together inside the container along with the proxy, making the distribution size pretty large. This way the container can be built from scratch, and can contain a single, statically-linked binary.

### DIFF
--- a/cmd/cloud_sql_proxy/build.sh
+++ b/cmd/cloud_sql_proxy/build.sh
@@ -6,15 +6,27 @@
 # This script simply allows a convenient way to cross compile and build a docker
 # container.
 #
-# With no arguments, this script will build a binary marked with "development"
+# With no arguments, this script will build a binary marked with "development",
+# otherwise the binary version will be annotated with the string provided.
+#
+#    # Build a binary labeled with 'development'
+#    ./build.sh
+#
+#    # Build a binary labeled with 'beta'
+#    ./build.sh beta
 #
 # Specifying 'release' as the first argument to this script will cross compile
 # for all supported operating systems and architectures. This requires a version
-# identifier to be supplied as the second argument.
+# identifier to be supplied as the second argument:
 #
-# Specifying 'docker' as the first argument to this script will build a
-# container, tagging it with the second argument. The tag will also be used as
-# the version identifier.
+#    # Build a binary for each of the supported platforms labeled with '0.001'
+#    ./build.sh release 0.001
+#
+# Specifying docker as the first argument to this script will build a
+# container, tagging it with the identifier in the second argument.
+#
+#    # Build a docker container named 'cloud-sql-proxy:my-tag'
+#    ./build docker my-tag
 
 files=$(git status -s)
 if [[ $? != 0 ]]; then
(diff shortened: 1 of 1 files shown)


Carrotman42:
I signed it!

cmd/cloud_sql_proxy/build.sh
```diff
@@ -1,39 +1,97 @@
...
-      echo "   Compile $OUT"
-      CGO_ENABLED=0 GOOS=$OS GOARCH=$ARCH go build -ldflags "-X 'main.versionString=$VERSION'" -a -installsuffix cgo -o $OUT .
+      build
     done
   done
-fi
+  ;;
+"docker")
+  if [[ "$2" == "" ]]; then
+    echo >&2 "Must provide a version number to use as the second parameter"
```
petercm: What you're using this for in `docker build` (val to -t) is really the name rather than version number. What you end up with if you pass `1.0` for example is:
```
REPOSITORY                                          TAG                 IMAGE ID            CREATED             SIZE
1.0                                                  latest              26b86071bec3        46 seconds ago      10.75 MB
```

It might be better to ask for both name and version number with version defaulting to the git sha and then use version as the tag. Eg `docker build -t "$2:$(git_version)"`. You can get around this right now by passing a version in the name:tag syntax but it makes some of the echos at least look off. Eg `build.sh docker cloudsql-proxy:e9ce9b2bbef0fbe33bf755b535e049b16513df13` results in: 

```
REPOSITORY                                          TAG                                        IMAGE ID            CREATED             SIZE
cloudsql-proxy         e9ce9b2bbef0fbe33bf755b535e049b16513df13   87a668e55dfa        42 seconds ago      10.75 MB
```

Carrotman42:
I just had the script hard-code the name to 'cloud-sql-proxy' and let the version decide the tag. Let me know if you think giving defaults to the 'version' or the ability to change the name would be useful.

I think I'm going to rewrite this script in Go (I'm much better at Go than Bash), but in the meantime feel free to open issues about any features you want in the existing script.

petercm:
If you rewrite in Go having the ability to override the hard-coded name might be nice (its a bit easier to modify the bash script to prefix with gcr.io/myproject/ vs. modifying and recompiling). Otherwise letting the arg decide the tag works for me.

On a related note how often is gcr.io/cloudsql-docker/gce-proxy rebuilt and/or will it start using this script to produce tagged images?

Carrotman42:
I'll start pushing the binaries with tags so that you can depend on a specific version, upgrading whenever you feel like it. I can stick to a concrete release cycle if you want, but I suspect updates to this client will be spurious rather than consistent so depending on the frequency there could be a no-op release. So far I've updated the image basically when I know there is a set of features released which would be useful for the average/first-time user.

I'm not sure what to do about the 'latest' tag. Part of me wants to update it on each new release because people who care about using a stable version can rely on a specific tag. Thoughts? Maybe I should release a new version and update the 'latest' tag after a week or so.

petercm:
I've been using my own build to get around the credential file issue. Most of the time I think as long as there's an easy way to revert when there's an issue the spurious releases are probably fine. The 'latest' tag thing [can be](https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375#.2rb99cpbl) [tricky](http://container-solutions.com/docker-latest-confusion/) as a production dependency. It _is_ probably good to keep one, I just wish it was easier to know when there was a new release and opt-in or out.

PR #33, opened 2016-04-26
adding port 0 call to validNets listen test

ptinsley:
Changing the listen call to use port 0 to grab an available port, the current code doesn't work on alpine linux in a container. This small change fixes it and I don't see a reason it would break other use cases assuming the Go source code is correct:

>If laddr has a port of 0, ListenTCP will choose an available port.

Example of this breaking:

>cloud_sql_proxy -dir=/cloudsql -instances=PROJECT:us-central1:INSTANCE=tcp:3306 -credential_file=/account.json
>2016/04/26 20:20:56 errors parsing config:
invalid "PROJECT:us-central1:INSTANCE=tcp:3306": unrecognized network tcp


Carrotman42:
Thanks for the merge request! I swear I checked whether passing 0 was okay.

PR #39, opened 2016-07-20
Update README.md for clearer instructions

frankvanrest:
- Only Second Generation is supported for CloudSQL Proxy
- On GCE you need to target `b.gcr.io/cloudsql-docker/gce-proxy:1.05` because of this Issue: GoogleCloudPlatform#35


Carrotman42:
Thanks much for the doc updates!

PR #50, opened 2016-10-15
Create kubernetes walk-through

apelisse:


Carrotman42:
Very sorry this fell off my radar. I'll commit it now and we can loop back around to see if things need to be updated.

PR #56, opened 2016-12-22
Minor readme typo correction

moizjv:
Changing `-credential_file` to `-credentials_file`


moizjv:
hmm, may be i need to edit the error message.

PR #64, opened 2017-03-06
Add support for Unix Socket connections to Postgres databases

ankushagarwal:
Use .s.PGSQL.5432 as the socket filename

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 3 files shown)


ankushagarwal:
I signed it!

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -228,7 +229,28 @@ func parseInstanceConfig(dir, instance string) (instanceConfig, error) {
...
+		if err != nil {
+			return instanceConfig{}, err
+		}
+		in, err := sql.Instances.Get(spl[0], spl[2]).Do()
+		if err != nil {
+			return instanceConfig{}, err
+		}
+		if strings.HasPrefix(strings.ToLower(in.DatabaseVersion), "postgres") {
+			path := filepath.Join(dir, instance)
+			if _, err := os.Stat(path); os.IsNotExist(err) {
```
Carrotman42: nit:

	path := filepath.Join(dir, instance)
	if _, err := os.Stat(path); !os.IsNotExist(err) {
		return instanceConfig{}, err
	}
	if err := os.MkdirAll(path, 0755); err != nil {
		return instanceConfig{}, err
	}
	ret.Address = filepath.Join(path, ".s.PGSQL.5432")

Carrotman42:
Sheesh, the big red X and O are intimidating. I just wanted one little change, github!

cmd/cloud_sql_proxy/proxy.go
ankushagarwal: Updated.

PR #66, opened 2017-03-08
Bug fix: Create the directory first and then check using os.IsNotExist

ankushagarwal:
!os.IsNotExist(err) returns true even when err == nil.


Carrotman42:
fwiw lgtm

Issue #61, opened 2017-02-07
Create official Helm chart


jayjun:
The official [docs](https://cloud.google.com/sql/docs/container-engine-connect) recommend the "sidecar pattern" for deploying to Google Container Engine. It's painstaking to follow the instructions currently, having to hand wire secrets up and carefully splice lines of code together from the sample YAML.

It'll be nice to have an official `google/cloudsql-proxy` Helm chart.

Install with a single `helm install google/cloudsql-proxy -f values.yaml` command.

In addition, I would like to suggest instructions for a standalone pod/service hosting `cloud_sql_proxy` that listens on `tcp:0.0.0.0:3306`, also installed using a Helm chart. Yes, it's less secure than listening on localhost, but for those who have secured their cluster properly, decoupling Cloud SQL proxy from the application pod greatly simplifies deployment.

Carrotman42:
I am unfamiliar with Helm entirely. Is there a chance you would be able to send a pull request?

jayjun:
@Carrotman42 Honestly I'm a Helm beginner too. Though I doubt this is a suitable repo to host the chart?

Anyway, I took a shot and submitted a PR for further discussion.

kevchentw:
There is a chart for it on https://github.com/kubernetes/charts/tree/master/stable/gcloud-sqlproxy

kevchentw:
Maybe we can add it on readme?

jayjun:
Oops, I closed the pull request (#67) already, but left this issue open.

I’m closing this issue now. Maybe @kevchentw can submit another pull request to adjust README.

kevchentw:
ok. I will send a PR latter

PR #67, opened 2017-03-09
Add Helm chart (#61)

jayjun:
This chart creates a standalone Kubernetes pod and service hosting Cloud SQL proxy. In its own namespace, if desired.

It exposes the proxy to the whole cluster by starting it with `tcp:0.0.0.0:3306` set, instead of `tcp:3306` which is only suitable for the "sidecar" pattern. This is significantly different from all the other methods recommended in the guide, and a potential security risk.

Also, I'm not sure if this repo is the right place to host this.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+name: cloudsql-proxy
+description: Deploys a Google Cloud SQL proxy as a Kubernetes service
+version: 0.1.0
+maintainers:
+  - name: Tan Jay Jun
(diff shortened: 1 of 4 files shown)


jayjun:
Link to another way of doing it. https://github.com/rimusz/charts/tree/master/stable/gcloud-sqlproxy

jayjun:
Helm charts should be kept in separate repos so they can be added via `helm repo add`.

rimusz’s chart is practically identical to my pull request, so I’m closing this.

hfwang:
Thanks for looking into this all the same!

PR #71, opened 2017-03-16
Add a cloudsqlpostgres dialer

ankushagarwal:


proxy/dialers/postgres/hook.go
```diff
@@ -0,0 +1,61 @@
...
+	sql.Register("cloudsqlpostgres", &drv{})
+}
+
+type drv struct{}
+
+type dialer struct{}
+
+func (d dialer) Dial(ntw, addr string) (net.Conn, error) {
+	// Drop the port number at the end (:5432) from the address
+	addr = regexp.MustCompile(":[0-9]+$").ReplaceAllString(addr, "")
```
Carrotman42: Compile regular expressions once per binary execution.

proxy/dialers/postgres/hook.go
```diff
@@ -0,0 +1,61 @@
...
+	// lib/pq converts the instance to the format '[project:region:instance]'
+	// format, so we need to drop the first and last character before we
+	// pass it to thr proxy dialer
+	if len(addr) < 2 {
+		return nil, fmt.Errorf("Failed to parse instance name: %q", addr)
```
Carrotman42: errors should start with lowercase letters

proxy/dialers/postgres/hook.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	addr = addr[1 : len(addr)-1]
```
Carrotman42: Do you mind checking explicitly for [ and ] characters before you cut them off? If the postgres driver ever shifts and stops putting them there this might start being super confusing.

Consider using a regexp group to remove both the port number and the [ ] with less manual logic.

proxy/dialers/postgres/hook.go
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package postgres adds a 'cloudsqlpostgres' driver to use when you want
+// to access a Cloud SQL Database via the go database/sql library.
+// It is a wrapper over the driver found at github.com/lib/pq.
+// To use this driver, you can look at an example in
+// postgres_test package in the hook_test.go file
+package postgres
```
Carrotman42: can you write a helper like in mysql? People like when libraries only take 1 line to call :) https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/proxy/dialers/mysql/hook.go#L40

ankushagarwal: This takes only one line to call

db, err := sql.Open("cloudsqlpostgres", "host=project:region:instance user=postgres dbname=postgres password=password sslmode=disable")

proxy/dialers/postgres/hook.go
ankushagarwal: Done

proxy/dialers/postgres/hook.go
ankushagarwal: Done

proxy/dialers/postgres/hook.go
ankushagarwal: Done

proxy/dialers/postgres/hook.go
Carrotman42: Doing that also requires a blank import of this package, which is tedious. It's nice to have a function to get the *sql.DB in this package since it will be a simpler explicit import (which are easier to understand and harder to forget). Plus, then your helper can call Ping and help prevent the giant misconception that sql.Open is proactive rather than lazy-dialing.

Also, that's a fairly annoying line to write. We want this library to be easy for people who have not used this driver before. It's nice to give them a Go function to call which will construct the DSN correctly rather than asking them to read a bunch of docs outside of this repo.

Definitely not asking you to remove what you have already, what you have is good to have. But having convenience functions is nice for people using our library, since I'd like this to be a convenience function.

mhindery:
Any update on this?

proxy/dialers/postgres/hook.go
ankushagarwal: I would argue that this way is better because this is exactly how the lib/pq driver is used. 

```
import _ "github.com/lib/pq"
db, err := sql.Open("postgres", "host=ipaddress user=postgres dbname=postgres password=password")
```

But if you insist, I can add the helper functions later on.

PR #75, opened 2017-03-30
Fix the issue on project ids which prefixed by organization

oto0:
Some cloud project ids have organization as prefix, separated with colon. It makes problem on parsing configuration that passes by -instances parameter. 

The observer error is :
 errors parsing config:
	googleapi: Error 403: The client is not authorized to make this request., notAuthorized

This fix handles this scenario and merge organization to project id again.


Carrotman42:
Thanks for raising the issue! I'd rather move forward with https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/76 though, as it shares code that already exists for this job.

PR #76, opened 2017-03-30
Fix bad handling of domain-scoped projects introduced in 1.07

Carrotman42:
Instead of reimplementing the tricky logic, just reuse the (unit tested) logic that exists already.


proxy/cloudsqlutil/cloudsqlutil.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
ankushagarwal: Rename package to just `util`?

Carrotman42: Done

PR #77, opened 2017-03-30
Use alpine docker base image to get certificates

apelisse:
Currently, one has to map a volume with certificates to the image.

We could simply fix that by using alpine and install certificates
directly in the image.


apelisse:
I wonder if you would consider that. It could be much more convenient, especially when using kubernetes.

apelisse:
Updated the documentation!

Carrotman42:
The simplification is much appreciated!

PR #83, opened 2017-04-13
Expanded the dialer test to provide additional examples of how to authenticate

digiexchris:
Expanded the dialer test to include being able to test using provided credentials as well as the default credentials. This should also provide a little bit more example code for people using the proxy in a go program without using default credentials

### DIFF
--- a/tests/dialers_test.go
+++ b/tests/dialers_test.go
@@ -59,6 +59,8 @@ func TestDialer(t *testing.T) {
 		t.Fatal("Must set -db_name")
 	}
 
+	//if token/tokenFile are not set, the github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/mysql library
+	//will automatically attempt to find default authentication via proxy.InitDefault */
 	if *token != "" || *tokenFile != "" {
 		client, err := clientFromCredentials()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


digiexchris:
@googlebot  I signed the CLA

digiexchris:
@googlebot  I signed the CLA

tests/dialers_test.go
```diff
@@ -44,6 +59,15 @@ func TestDialer(t *testing.T) {
 		t.Fatal("Must set -db_name")
 	}
 
+	if *token != "" || *tokenFile != "" {
```
Carrotman42: Would you mind adding a comment here explaining that if token/tokenFile are not set, the `github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/mysql` library will automatically attempt to find default authentication via `proxy.InitDefault`?

tests/dialers_test.go
```diff
@@ -59,3 +83,38 @@ func TestDialer(t *testing.T) {
...
+
+		all, err := ioutil.ReadFile(f)
+		if err != nil {
+			return nil, fmt.Errorf("invalid json file %q: %v", f, err)
+		}
+		cfg, err := goauth.JWTConfigFromJSON(all, SQLScope)
+		if err != nil {
+			return nil, fmt.Errorf("invalid json file %q: %v", f, err)
+		}
+		t, _ := cfg.TokenSource(ctx).Token()
```
Carrotman42: Instead of pulling out the oauth2 package explicitly in this branch, how about using the Client method on the JWT config?

    client = cfg.Client(ctx)

tests/dialers_test.go
digiexchris: You bet! Change made.

tests/dialers_test.go
digiexchris: Ah, that's way better thanks. I've made that change!

Carrotman42:
Thanks for the pull request! The next review will hopefully go faster than this one :)

Issue #78, opened 2017-04-04
cannot fetch token: Post https://accounts.google.com/o/oauth2/token: x509: failed to load system roots and no roots provided


norbert-yoimo:
I'm trying to connect to a postgresql cloudsql instance, from a kubernetes pod, but when I try to establish a connection with `psql -h 127.0.0.1`, I get:

```
psql: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

Looking at the logs of the cloudsql pod, I see:

```
2017/04/04 13:15:53 New connection for "myproject-157612:europe-west1-c:myproject"
2017/04/04 13:15:53 couldn't connect to "myproject-157612:europe-west1-c:myproject": Post https://www.googleapis.com/sql/v1beta4/projects/myproject-157612/instances/myproject/createEphemeral?alt=json: oauth2: cannot fetch token: Post https://accounts.google.com/o/oauth2/token: x509: failed to load system roots and no roots provided
```

I'm using cloudsqlproxy 1.07, with the following deployment spec:

```
- image: gcr.io/cloudsql-docker/gce-proxy:1.07
    name: cloudsql
    command: ["/cloud_sql_proxy", "-dir=/cloudsql", "-credential_file=/secret/cloud-sql.json", "-instances=myproject-157612:europe-west1-c:myproject=tcp:5432"]
    volumeMounts:
        - name: cloudsql
          mountPath: /cloudsql
        - name: sql-proxy-secret
          mountPath: /secret/
```

Am I supposed to mount some SSL certificates inside the cloud sql pod?

Carrotman42:
Please try the 1.09 release, it should include those certificates in the
docker container.

norbert-yoimo:
That fixed it, thank you.

duzy:
Hi, I'm using the proxy lib (instead of cloud_sql_proxy) in my Go program. But I'm having exactly the same issue as this. All my guess is the cert files are missing, but not sure what to do for a fix!

```
2017-04-11T11:04:38.746339777Z 2017/04/11 11:04:38.746134 create table contributor: error: Post https://www.googleapis.com/sql/v1beta4/projects/xxx/instances/lingual-work-vocab/createEphemeral?alt=json: x509: failed to load system roots and no roots provided
```

Carrotman42:
Are you creating your own docker container? If so, you have to include
certificates for yourself. Look at the Dockerfile in this repo for the base
[...]

PR #85, opened 2017-04-21
Update README.md

jerairrest:
Update example to new version that does not have an error. Based on https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/78


jerairrest:
I signed it!

Carrotman42:
Sorry this slipped by without me noticing! Thanks for the fix.

PR #100, opened 2017-06-16
Add Quiet Flag

mykolasmith:
Hello,

This pull requests adds a `-quiet` flag. I am using `cloud_sql_proxy` in production on Google Container Engine, [per these instructions](https://cloud.google.com/sql/docs/postgres/connect-container-engine), and the constant "Connection established" logging pollutes our log streams (and will eventually incur additional costs to Stackdriver Logging). Please consider merging this PR so that we may silence logging.

Thanks!

Issue #44, opened 2016-09-07
Make logging configurable


MDrollette:
When using the proxy with PHP (and no persistent connections) the chattiness of the proxy logs is overwhelming. Since I am running the proxy on Container Engine, those logs are being shipped to Cloud Logging and the amount of "new connection" and "connection closed" messages is significant.

It would be nice to disable logging of new and closed connections.

Carrotman42:
Sure, that makes a lot of sense.

But please note: you really /should/ be pooling connections. Particularly with the proxy, dialing new database connections is significantly less efficient than using existing connections.

OlivierBoucher:
I could submit a pull request if you can tell me which logging library you would like to use.

Carrotman42:
As for what library to use, I would prefer to continue using the standard
logging library in Go. But, it'd be best to abstract it so that users of
the Go library can swap in their own favorite logging library if they feel
like it.

My initial thought is to create a new package `
github.com/GoogleCloudPlatform/cloudsql-proxy/log` which contains exported
variables `Infof`, `Verbosef`, etc which simply forward to the stdlib's
`log.Infof`. In the main file, a "--verbose" flag should be defined, and an
init function should overwrite the definition of `log.Verbosef` to first
check the flag value and then drop the log if the flag is false.

OlivierBoucher:
Thanks, will submit a PR soon

OlivierBoucher:
@Carrotman42 Can you tell me if the current interface suits you? I'm pretty sure it would be compatible with most logging libraries. Just give me a clear and I'll replace references to the stdlib logger in the entire repo

mykolasmith:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/100

Carrotman42:
@OlivierBoucher sorry I didn't see your comment from a day ago, and I'm also not exactly sure which interface you mean.

After receiving #100  I went ahead and made the changes I suggested in https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/44#issuecomment-306547267. This will be sent out later today.

Carrotman42:
Setting -verbose=false will now turn off the new/closed connection logs. Note that errors which are not EOF will still be logged when in non-verbose mode since these may be useful for debugging client-/server-side problems. Those who would like to disable all output can use the -quiet flag submitted in #100

PR #101, opened 2017-06-16
Add -verbose flag to allow users to turn off verbose output

Carrotman42:
The flag defaults to true to keep old functionality (and to ease debugging when first setting it up), but people running the proxy in a server environment will likely want to turn off verbose output.

At some point we could consider setting the default to false.

Addresses #44 along with #100


hfwang:
LGTM

PR #104, opened 2017-06-23
Set the UserAgent used when contacting the sqladmin API

Carrotman42:
We use the version of the proxy for the officially released proxy binary and "custom cloud_sql_proxy version >= 1.10" for any code that is custom-built.


proxy/certs/certs.go
```diff
@@ -34,9 +34,43 @@ import (
...
+		UserAgent:    defaultUserAgent,
+	})
+}
+
+// RemoteOpts are a collection of options for NewCertSourceOpts. All fields are
+// optional.
+type RemoteOpts struct {
+	// Host specifies the base path for the sqladmin API. If left blank, the
+	// default from the autogenerated sqladmin library is used.
+	Host string
```
laixer: Out of scope of this CL: This ought to be called something other than 'host'. Maybe APIBasePath?

PR #109, opened 2017-07-28
Fix gce_test.go to run on everyone's computer.

Carrotman42:
For reasons (I think) related to cached .a files on my computer, I was actually able to import a main package into other packages for a while. Either the .a files got cleaned or the recent Go compiler fixed the unintended support, but in any case now everyone can correctly run the gce_test.

### DIFF
--- a/tests/gce_test.go
+++ b/tests/gce_test.go
@@ -42,7 +42,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tests/gce_test.go
```diff
@@ -138,6 +147,28 @@ func (p *process) Close() error {
 	return p.sess.Close()
 }
 
+func compileProxy() (binaryPath string, _ error) {
+	// Find the 'build.sh' script by looking for it in cwd, cwd/.., and cwd/../..
+	var buildSh string
+	for parents := 0; parents < 2; parents++ {
+		cur := filepath.Join(strings.Repeat(".."+string(filepath.Separator), parents), buildShLocation)
```
hfwang: Not a big deal, but how about having a slice for base path that starts as an empty slice and then appending .. to it with each loop iteration, that way you aren't doing your own filepath.Join via strings.Repeat?

Carrotman42: Done

PR #110, opened 2017-08-01
add install instruction for Kubernetes

kevchentw:


kevchentw:
I signed it!

README.md
```diff
@@ -78,6 +78,13 @@ I'm open to adding more drivers, feel free to file an issue.
 
 ## To use from Kubernetes:
 
+### Using Deployment and Service
+Follow this [instruction](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/Kubernetes.md)
+
+### Using Helm
+Follow this [instruction](https://github.com/kubernetes/charts/tree/master/stable/gcloud-sqlproxy)
```
Carrotman42: Looking through this page I see that the tag of the newest version of the Proxy is already out of sync (1.10 is already released). How should we make sure that page stays up to date?

Carrotman42: 'latest' is definitely not a suggested configuration since that allows the Proxy code to be changed at any time (e.g. by me when I update that tag). Not only could that introduce instability to those who rely on the code (longer startup times on new versions), if there is a bug in the latest version I would singlehandedly (accidentally) break any users affected by that bug. This has already happened in the past and I'd very much prefer it not happen again.

Ideally, users of the Proxy would use a specific tag and update that tag along with whatever rollout procedures they have for their own application, making sure that it is validated before it is pushed out to production. I think it's best to link to the [releases page](https://github.com/GoogleCloudPlatform/cloudsql-proxy/releases) in that readme and advise people to update the `values.yaml` to stay up to date with the new versions of the Proxy as they are released.

I'm hesitant to use `latest` in the default `values.yaml` in the repository, but it's probably the best way to keep things up to date in that repo unfortunately. Maybe we should just add a note in the `values.yaml` about not using the `latest` tag in production to avoid getting a bug accidentally pushed to them.

kevchentw: I have added some note in the example YAML. 
I think it is better than adding in readme or description, since a lot of people just copy and paste :)

Carrotman42:
Thanks for the pull request!

PR #111, opened 2017-08-03
Retry the SSH in gce_test for up to 100 seconds.

Carrotman42:
Often VMs cannot immediately be SSH'd to after creation. A retry fixes this, and we might as well do it in the test instead of forcing people to retry the test itself


hfwang:
LGTM.

Oh GCE. You always have a funny definition of "ready".

Issue #99, opened 2017-06-14
Feature: flag to set client.RefreshCfgThrottle


imtaco-lh:
Default client.RefreshCfgThrottle is one minute, so when cloudsql failover, proxy might wait for more than one minute to connect the new master if first connection retry fail.

Is possible to add a flag to set RefreshCfgThrottle manually ?

Carrotman42:
I think this is a good idea, although I'd like to note that there probably should be some sort of forced lower bound to help people not shoot their API quota in the foot in the case that their application does not exponentially back off.

PR #112, opened 2017-08-03
Add --refresh_config_throttle for setting the period between API calls.

Carrotman42:
Fixes https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/99


Carrotman42:
Note that there shouldn't be a new certificate required after a failover, since when the replica becomes the new master the certificates will be updated so that it looks like the old master. Users will have to reconnect, but shouldn't need a new cert.

In any case, I think the default value of 1 minute is a little high and we may want to decrease it to like 20-30 seconds (especially because we recently increased default API quota for customers) but for now exposing the flag should help out power users for now. Hopefully the flag documentation and the minimum value will help most people realize that this is not a flag to be used lightly.

PR #113, opened 2017-08-09
Invalid Kubernetes config

japgolly:
Running the example yields this error:
```
Error from server (Invalid): error when creating "/dev/fd/63": DaemonSet.extensions "cloudsqlproxy" is invalid: spec.template.spec.containers[0].ports[0].name: Invalid value: "sqlproxy-port-database1": must be no more than 15 characters
Error from server (Invalid): error when creating "/dev/fd/63": Service "sqlproxy-service-database1" is invalid: spec.ports[0].targetPort: Invalid value: "sqlproxy-port-database1": must be no more than 15 characters
```


japgolly:
`I signed it!`

Carrotman42:
Thanks for the pull request!

Issue #119, opened 2017-08-23
Readiness and liveness probes in Kubernetes


wayneashleyberry:
I'm struggling to setup a health check for the proxy container on Kubernetes. Some documentation or guidance around this would be great. We're having to delete pods quite regularly due to an app not being able to connect.

Tried [defining a tcp liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-tcp-liveness-probe) which didn't quite do the trick.

AthenaShi:
So far we don't expose such thing unfortunately...

As a work around you can check by run cmd: 
netcat -U /cloudsql/$INSTANCE_CONNECTION_STRING

As for the "an app not being able to connect" issue, if you think it might due to cloud sql proxy, please send an email with following information to cloud-sql@google.com so I may be able to help out looking at some internal data.

- with a link to this page.
- describe the issue you have.
- the full name of your cloud sql instance.
- the command your run
- the full logs

rosskevin:
When trying to [deploy as recommended in kubernetes](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/Kubernetes.md) with spinnaker, it seems the lack of the probe causes the `highlander` strategy to tear down the node.

I'll try the `netcat`, but it would be convenient if there was a typical set of endpoints without having the INSTANCE_CONNECTION_STRING in more than one place.

rosskevin:
`netcat` is not available on the stock image, so this:

```yaml
livenessProbe:
  exec:
    command: ["netcat", "-U", "/cloudsql/coverbid:us-central1:db=tcp:3306"]
  initialDelaySeconds: 5
  timeoutSeconds: 10
```

fails with 

```
Liveness probe failed: rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:247: starting container process caused "exec: \"netcat\": executable file not found in $PATH" 
```

rosskevin:
I misunderstood netcat's location and options on the image.

I attempted the following:

```yaml
livenessProbe:
  exec:
    command: ["nc", "/cloudsql/foo:us-central1:db=tcp:3306"]
  initialDelaySeconds: 5
  timeoutSeconds: 3
```

and it failed with:

```
Liveness probe failed: nc: bad address '/cloudsql/foo:us-central1:db=tcp:3306'
```

Has anyone else gotten `nc` to work as a probe?

[...]

PR #123, opened 2017-10-09
Add netcat for internal liveness probe

rosskevin:
If running in a kubernetes cluster standalone, `netcat` needs to be available to execute an internal liveness probe as mentioned in #119.


rosskevin:
My misunderstanding, `netcat` is available as `nc`.

Issue #94, opened 2017-06-09
Feature: Refresh SSL cert before it expires in the background


Jille:
We should refresh the SSL certificate in the background before it expires. This will avoid the latency spike every hour (as mentioned in #87).

patrykmikula:
Any progress on this issue ?

AthenaShi:
I've created an internal ticket and it will be prioritized.

kurtisvg:
It looks like #160 addresses this.

lychung83:
fixed by merge of https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/222

PR #142, opened 2018-01-17
Automatically refresh SSL certs before they expire

jellybeanfiend:
fixes issue #94


proxy/proxy/client.go
```diff
@@ -225,6 +228,15 @@ func (c *Client) Dial(instance string) (net.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Refresh cert 5 minutes before it expires.
+	timer := time.NewTimer(cfg.Certificates[0].Leaf.NotAfter.Sub(time.Now()) - refreshCertBuffer)
+	go func() {
```
Carrotman42: I think this makes more to put in refreshCfg since we don't want to start more than one goroutine at a time for refresh (and refreshCfg is the one with the mutex).

proxy/proxy/client.go
```diff
@@ -225,6 +228,15 @@ func (c *Client) Dial(instance string) (net.Conn, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Refresh cert 5 minutes before it expires.
+	timer := time.NewTimer(cfg.Certificates[0].Leaf.NotAfter.Sub(time.Now()) - refreshCertBuffer)
```
Carrotman42: You should definitely verify that the delay value passed to the time package is not negative since the cert info is from a remote server, or else the code will probably panic if that server ever serves a strange-looking cert.

Carrotman42: Since you're not reusing the time.Timer, it's simpler to use `time.After` instead.

proxy/proxy/client.go
```diff
@@ -225,6 +228,15 @@ func (c *Client) Dial(instance string) (net.Conn, error) {
...
+	go func() {
+		<-timer.C
+		logging.Verbosef("Cert for instance %s will expire soon, refreshing now.", instance)
+		c.refreshCfg(instance)
```
Carrotman42: Due to the way refreshCfg/cachedCfg works (refreshCfg grabs the write-lock of `c.cfgL`, and cachedCfg grabs the read-lock), calling this in a background goroutine is not sufficient to prevent delays of Dialing. Previously the delay would happen during the first dials after the cert expires. Now, any dials that happen 5 minutes before the cert expires will block just the same. You'll have to do something fancier with the locks to get concurrent Dials during the time refreshCfg is contacting the backend.

jellybeanfiend: Ah, I see what you mean now. Is there any reason to hold the lock for the entirety of the method? What about only using the read lock to get the instance data from cfgCache and only lock when updating cfgCache in the deferred function? Sorry if it's a dumb question, still trying to get up to speed on the code.

proxy/proxy/client.go
jellybeanfiend: In refreshCfg it would still look something like:
go func() {
    <- time.After(blah)
    c.refreshCfg(instance)
}() (in addition to the mutex changes mentioned in the other comment) Otherwise I'm not sure what you mean. Could you elaborate?

Carrotman42: Yep, just like that. The main difference is that you can be sure that this happens exactly once if you're holding the redialing mutex while setting this off. Otherwise you might get a growing number of goroutines sleeping and then refreshing the config with a corresponding unbounded increase of RPCs to Cloud SQL API backends. That'd be a fun problem to have to mitigate for sure ;)

proxy/proxy/client.go
Carrotman42: No dumb questions, if the code is hard to understand I'm the one to blame :)

The write lock is being held the entire call to refreshConfig to prevent concurrent dials. For example, if two application threads try to open a new connection to a database which has expired config at the same time, we only want to contact the Cloud SQL API on one of the corresponding goroutines while the other waits for the cert to be updated. Right now the lock is very coarse-grained: only one goroutine is allowed to contact the API at a time and no goroutines are allowed to read the config during this period. It was designed this way because we figured that most of the time there was one or two instances that the proxy was going to connect to, so this simpler, coarser locking structure was advantageous to code clarity without really affecting performance (since other goroutines probably had to wait for their cert to be updated anyway).

As you suggest, I agree from a thread-safety point of view it is safe to just unlock the write lock during the calls to `c.Certs.Local` and `c.Certs.Remote`, but then those two threads would try to contact the backend at the same time instead of redialing on their own. Since there are potentially hundreds of application threads that could try to do a redial at the same time I don't think it will scale very well. I think we need to add a bit more depth to the locking strategy. Here's some requirements I can think of:

   1) Readers must be able to read cached config while goroutines are contacting the API
   2) Only one goroutine may be contacting backend at a time (at least, per instance)
   3) When a reader finds the cached config is invalid/expired AND there is already a goroutine contacting the backend, this reader should wait for and use the response that goroutine gets (rather than contacting the backend itself).

Given those requirements I whipped up something that I think could work, but you should definitely look it over to make sure it isn't witchcraft (i.e. I haven't even compiled it). And if you have another implementation you can think of that you like more free to use it. You can ping me over chat/email if you need more clarification or anything.

    type Client struct {
       // A bunch of other stuff
       cfgCache map[string]*cacheEntry
       cfgL     sync.RWMutex
    }

    type cacheEntry struct {
        timestamp time.Time
        // The err, addr, and cfg fields of this entry must not be read before this channel has been closed.
        ready chan struct{}
        // If err is not nil, the addr and cfg are not valid.
        err  error
        addr string
        cfg  *tls.Config
     }

    func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, err error) {
       throttle := c.RefreshCfgThrottle
       if throttle == 0 {
          throttle = DefaultRefreshCfgThrottle
       }

       c.cfgL.Lock()

       // cfgCache is lazy-init
       if c.cfgCache == nil {
          c.cfgCache = make(map[string]*cacheEntry)
       }

       old, oldok := c.cfgCache[instance]

       if oldok && time.Since(old.timestamp) < throttle {
          c.cfgL.Unlock()
          logging.Errorf("Throttling refreshCfg(%s): it was only called %v ago", instance, time.Since(old.lastRefreshed))
          // Refresh was called too recently, just reuse the result after it is ready
          <-old.ready
          return old.addr, old.cfg, old.err
       }

       ent := &cacheEntry{
          timestamp: time.Now(),
          ready: make(chan struct{}),
       }
       c.cfgCache[instance] = ent
       c.cfgL.Unlock()

       defer func() {
          ent.err = err
          ent.addr = addr
          ent.cfg = cfg
          // Now that we've written the response in the entry, close the channel to
          // signal that the cache entry is ready.
          close(ent.ready)
       }()

       mycert, err := c.Certs.Local(instance)
       if err != nil {
          return "", nil, err
       }

       scert, addr, name, err := c.Certs.Remote(instance)
       if err != nil {
          return "", nil, err
       }
       certs := x509.NewCertPool()
       certs.AddCert(scert)

       cfg = &tls.Config{
          ServerName:   name,
          Certificates: []tls.Certificate{mycert},
          RootCAs:      certs,
       }
       return fmt.Sprintf("%s:%d", addr, c.Port), cfg, nil
    }

proxy/proxy/client.go
jellybeanfiend: Moved to refreshCfg.

proxy/proxy/client.go
jellybeanfiend: Done!

jellybeanfiend: Done.

proxy/proxy/client.go
jellybeanfiend: What about using two separate locks: one to protect the number of concurrent API calls and one to protect the cache? The former would be held through the entirety of refreshCfg to allow for only one goroutine to make the API calls at a time, and the latter would allow other goroutines to access the cache while the cert is refreshed.

I don't think this would meet criteria number 3, but if the idea is to refresh the cert a few minutes before it expires, I don't think readers will typically need to wait for the cert to be refreshed.

PTAL at my changes. I feel like I'm probably missing something. :)

proxy/proxy/client.go
```diff
@@ -78,9 +81,12 @@ type Client struct {
...
 	// The cfgCache holds the most recent connection configuration keyed by
 	// instance. Relevant functions are refreshCfg and cachedCfg. It is
-	// protected by cfgL.
+	// protected by cacheL.
 	cfgCache map[string]cacheEntry
-	cfgL     sync.RWMutex
+	cacheL   sync.RWMutex
+
+	// cfgL prevents multiple goroutines from contacting the Cloud SQL API at once.
+	cfgL sync.Mutex
```
Carrotman42: Can we call this "refreshL" since it guards refreshCfg? (or refreshCfgL if you prefer)

proxy/proxy/client.go
```diff
@@ -158,24 +164,28 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, err
...
 		return old.addr, old.cfg, old.err
 	}
 
-	if c.cfgCache == nil {
-		c.cfgCache = make(map[string]cacheEntry)
-	}
-
 	defer func() {
+		c.cacheL.Lock()
 		c.cfgCache[instance] = cacheEntry{
```
Carrotman42: We should enhance the cache update logic a bit. If there is an existing cache entry which is valid (no error, and the cert isn't expired) we should avoid overwriting that cache entry with a new one if we have an error getting a new one. We might as well allow the user to use the other cert for 5 more minutes and the normal "redial when expired" mechanism will work as expected. (we don't need to set off another background goroutine or retry in the background, I think an error here will be rare enough.)

proxy/proxy/client.go
```diff
@@ -195,13 +205,24 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, err
...
 		RootCAs:      certs,
 	}
+
+	// Refresh cert 5 minutes before it expires.
+	timeToRefresh := cfg.Certificates[0].Leaf.NotAfter.Sub(time.Now()) - refreshCertBuffer
+	if timeToRefresh > 0 {
+		go func() {
+			<-time.After(timeToRefresh)
+			logging.Verbosef("Cert for instance %s will expire soon, refreshing now.", instance)
+			c.refreshCfg(instance)
```
Carrotman42: I think checking for an error here would be a good idea since it is still an error which users may care about. They'd also receive an error the next time they tried to dial, but I figure we might as well print it out now.

proxy/proxy/client_test.go
```diff
@@ -183,3 +192,43 @@ func TestMaximumConnectionsCount(t *testing.T) {
...
+		},
+		RefreshCfgThrottle: 500 * time.Millisecond,
+	}
+
+	// Call Dial to cache the cert.
+	if _, err := c.Dial(instance); err != errFakeDial {
+		t.Errorf("unexpected error: %v", err)
+	}
+
+	cached, ok := c.cfgCache[instance]
```
Carrotman42: We should strive to write all of our code to be free of race conditions, especially tests: when we run them we can enable the race detector to check for problems, and ideally any problems found would be in the SUT rather than the unit tests themselves!

I suggest running the tests under the race detector (add `-race` to `go test`) until it doesn't report problems.

(and if there happen to be problems already, pleas file an issue! :] we don't run these tests regularly under the race detector so it's possible another regression has slipped in)

proxy/proxy/client_test.go
```diff
@@ -183,3 +192,43 @@ func TestMaximumConnectionsCount(t *testing.T) {
...
+	if !ok {
+		t.Error("expected instance to be cached")
+	}
+
+	// Wait for cert to expire.
+	time.Sleep(timeToExpire + time.Second)
```
Carrotman42: In general we should avoid sleeping in tests, it tends to add flakiness to otherwise deterministic tests. But in this case we're waiting for something in the background to finish so I think it's OK. But to cut down on flakiness, I suggest looping a few times (with short sleeps in the middle) in case the threads are getting scheduled in an interesting order.

proxy/proxy/client.go
jellybeanfiend: Done.

proxy/proxy/client.go
jellybeanfiend: Done, I think it's enough to check oldok.

proxy/proxy/client.go
jellybeanfiend: Done.

proxy/proxy/client_test.go
jellybeanfiend: Done.

jellybeanfiend:
I ran a test to make sure my fix actually works. I ran two proxy clients, one with and without the fix, and measured how long it took to connect to a mysql instance with both of them several times per millisecond for 5 hours. Here's the graph of connection durations (blue is without fix, red is with fix): https://screenshot.googleplex.com/FjoTYD1jks6 Seems to work properly!

Carrotman42:
Do you mind resharing the picture with labeled axis (I'm not exactly sure what I'm looking at), and can you attach the image directly to your comment so that everyone can view it?

jellybeanfiend:
Yep, sorry about that!

with_and_without_fix
beforefix
withfix

PhiloNL:
@Carrotman42 any idea when this will be merged? 😄  The SSL refresh sometimes times out causing 'MySQL server has gone away' errors because the proxy can't connect to the server. Happens once or twice a month.

proxy/proxy/client.go
```diff
@@ -150,32 +156,39 @@ func (c *Client) handleConn(conn Conn) {
...
 		return old.addr, old.cfg, old.err
 	}
 
-	if c.cfgCache == nil {
-		c.cfgCache = make(map[string]cacheEntry)
-	}
-
 	defer func() {
+		if err != nil && oldok {
+			return
```
Carrotman42: I don't think this logic is correct and could cause a problem when there is a consistent error trying to refresh the cert. If `oldok == true` but the cached cert is expired, we *want* to overwrite the cache entry with a new lastRefreshed timestamp so that subsequent callers to refreshCfg get throttled above rather than becoming a thundering herd to the cloud SQL API.

Unfortunately there's no unit test to check for that, in the past I was just careful :) There probably should be, but not necessarily for this CL.

proxy/proxy/client.go
```diff
@@ -195,13 +208,26 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, err
...
+			}
+		}()
+	}
+
 	return fmt.Sprintf("%s:%d", addr, c.Port), cfg, nil
 }
 
 func (c *Client) cachedCfg(instance string) (string, *tls.Config) {
-	c.cfgL.RLock()
+	c.cacheL.Lock()
```
Carrotman42: I think the cacheL should still be a RWMutex so that multiple concurrent dialers (with a valid, cached cert) can call this method in parallel, otherwise there's needless contention.

Carrotman42:
Sorry @PhiloNL I don't know: I don't have commit access to this repo any more.

hfwang:
Sorry for the noise. I thought carrotman42 had LGTM-ed.

@jellybeanfiend PTAL at the comments

PR #145, opened 2018-01-24
cmd: use INSTANCES env var

ceocoder:
in spirit of keeping manifest files same across different projects,
it'd be good to pass some config options via environment variables.
with this change, one can pass either flag -instances or INSTANCES
environment variable to get same results

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -63,7 +63,8 @@ Unix socket-based connections.`)
 	instances = flag.String("instances", "", `Comma-separated list of fully qualified instances (project:region:name)
 to connect to. If the name has the suffix '=tcp:port', a TCP server is opened
 on the specified port to proxy to that instance. Otherwise, one socket file per
-instance is opened in 'dir'. Not compatible with -fuse`)
+instance is opened in 'dir'. You may use INSTANCES environment variable
+for the same effect. Using both will use value from flag, Not compatible with -fuse`)
 	instanceSrc = flag.String("instances_metadata", "", `If provided, it is treated as a path to a metadata value which
 is polled for a comma-separated list of instances to connect to. For example,
 to use the instance metadata value named 'cloud-sql-instances' you would
@@ -137,6 +138,9 @@ Connection:
             -instances=my-project:my-region:my-instance=tcp:3306
 
      When connecting over TCP, the -instances parameter is required.
+
+     Supplying INSTANCES environment variable achieves the same effect.  One can
+     use that to keep k8s manifest files constant across multiple environments
 
   -instances_metadata
      When running on GCE (Google Compute Engine) you can avoid the need to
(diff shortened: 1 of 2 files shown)


ceocoder:
@AthenaShi should I create an issue for this first?  this is more of a feature request and solution in one.

ceocoder:
That works,  do you think just joining by a `,` should be fine?

ceocoder:
@Carrotman42 done.  wdyt?

ceocoder:
@Carrotman42 updated [README.md](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/145/files#diff-04c6e90faac2675aa89e2176d2eec7d8) as well, left Kubernetes.md as is for sake of brevity.    Would it be possible to have a new release with this feature built this week?

ceocoder:
gentle nudge for this, please let me know if there is anything we can do to get this merged.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -217,6 +231,7 @@ func userAgentFromVersionString() string {
 const accountErrorSuffix = `Please create a new VM with Cloud SQL access (scope) enabled under "Identity and API access". Alternatively, create a new "service account key" and specify it using the -credential_file parameter`
 
 func checkFlags(onGCE bool) error {
+
```
AthenaShi: Remove this newline please.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -138,6 +140,18 @@ Connection:
 
      When connecting over TCP, the -instances parameter is required.
 
+     Supplying INSTANCES environment variable achieves the same effect.  One can
+     use that to keep k8s manifest files constant across multiple environments.
+     Supplying both INSTANCES and -instances will result in joining of two inputs.
```
AthenaShi: I would suggest let the "-instances" flag overrides the "INSTANCES" env if both specified (which in my opinion shouldn't).  If somebody specify by flag, one might not want it to be affected by the env variable.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -395,6 +410,10 @@ func main() {
 		projList = gcloudProject()
 	}
 
+	if envInstances := os.Getenv("INSTANCES"); envInstances != "" {
+		instList = append(instList, stringList(envInstances)...)
```
timtadh: It seems a bit weird to join the environment variable with the values from the command line. Are their examples of other common tools with this behavior. IIRC most things have the command line value override the environment variable.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
ceocoder: @AthenaShi that was the original goal, I changed it to this based on @Carrotman42 's note.  Will revert if this indeed desired.

ceocoder: @AthenaShi can't find that comment on github (miss gerrit), pasting it from my email,

> Kevin Malachowski <notifications@github.com> 
> How about appending the two instead of over writing? It'd be in the same spirit as how -instances and -instance_metadata can be used together.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
ceocoder: done

cmd/cloud_sql_proxy/cloud_sql_proxy.go
ceocoder: done.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
ceocoder: yep, fixed.

timtadh:
:+1: from me.

PR #148, opened 2018-02-23
Gracefully handle Temporary errors from Accept()

Jille:
This should prevent the Cloud SQL proxy from dying once it e.g. hits the
max filedescriptor limit.


Carrotman42:
Or, blindly sleeping an amount of time would work too. I guess we don't need a back off, we just shouldn't consume a whole CPU core during the duration of the transient error.

Carrotman42:
And print out the error too? It would be useful to help debug a problem.

AthenaShi:
Any update on this?

Jille:
I've added a small sleep (for the case the Accept() was < 10ms).

The errors were already printed out before my if statements.

Carrotman42:
Sorry, missed the print out. Lgtm

AthenaShi:
The "nap" looks good ;)

PR #160, opened 2018-04-23
Automatically refresh SSL certs before they expire, part 2

jellybeanfiend:
For issue #94, Follow up of https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/142 which was reverted due to some unaddressed comments.

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -32,10 +32,13 @@ const (
 	keepAlivePeriod           = time.Minute
 )
 
-// errNotCached is returned when the instance was not found in the Client's
-// cache. It is an internal detail and is not actually ever returned to the
-// user.
-var errNotCached = errors.New("instance was not found in cache")
+var (
+	// errNotCached is returned when the instance was not found in the Client's
+	// cache. It is an internal detail and is not actually ever returned to the
+	// user.
+	errNotCached      = errors.New("instance was not found in cache")
+	refreshCertBuffer = 5 * time.Minute
+)
 
 // Conn represents a connection from a client to a specific instance.
 type Conn struct {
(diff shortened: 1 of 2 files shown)


proxy/proxy/client_test.go
```diff
@@ -183,3 +192,54 @@ func TestMaximumConnectionsCount(t *testing.T) {
...
+	}
+
+	c.cacheL.Lock()
+	cached, ok := c.cfgCache[instance]
+	c.cacheL.Unlock()
+	if !ok {
+		t.Error("expected instance to be cached")
+	}
+	waitTil := time.After(timeToExpire + (10 * time.Millisecond))
+loop:
```
Carrotman42: Maybe this came from a previous review comment of mine, but: this loop doesn't do anything different than simply executing this sleep:
   
    time.Sleep(timeToExpire + 10*time.Millisecond)

I somewhat recall asking for a sleep+retry loop. If that's what you were attempting to implement here, you should be checking the cfgCache multiple times and sleep+retrying if the config you were looking for was not found/refreshed. That will reduce flakiness by giving the background goroutines some time to process while still reducing the amount of time the test takes in the common case.

proxy/proxy/client.go
```diff
@@ -195,16 +207,33 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, err
 		Certificates: []tls.Certificate{mycert},
 		RootCAs:      certs,
 	}
+
+	// Refresh cert 5 minutes before it expires.
+	timeToRefresh := cfg.Certificates[0].Leaf.NotAfter.Sub(time.Now()) - refreshCertBuffer
```
Carrotman42: nit: please use `mycert` here instead of `cfg.Certificates[0]`

proxy/proxy/client.go
```diff
@@ -150,31 +156,37 @@ func (c *Client) handleConn(conn Conn) {
...
-	}
-
 	defer func() {
-		c.cfgCache[instance] = cacheEntry{
-			lastRefreshed: time.Now(),
-
-			err:  err,
-			addr: addr,
-			cfg:  cfg,
+		if err == nil || (oldok && isExpired(old.cfg)) {
```
Carrotman42: Still not all the cases :) You still want to update the cache in the case that the first call to refreshCfg for an instance returned an error: in that case, `oldok` would be false and `err` would be non-nil.

kurtisvg:
@jellybeanfiend Are you planning on continuing work on this PR?

hfwang:
jellybeanfiend has mentioned they will not be continuing to work on this. I'll find someone internally to pick this back up.

PR #165, opened 2018-05-17
Publicly expose the driver so it can be wrapped

luna-md:
This makes the Driver struct for postges public, this allows tools like https://github.com/luna-duclos/instrumentedsql to wrap it


luna-md:
Closing this PR to resubmit under my personal time and github account

PR #166, opened 2018-05-17
Publicly expose the driver so it can be wrapped

luna-duclos:
This lets tools such as https://github.com/luna-duclos/instrumentedsql wrap the driver and provide instrumentation.

### DIFF
--- a/proxy/dialers/postgres/hook.go
+++ b/proxy/dialers/postgres/hook.go
@@ -32,10 +32,10 @@ import (
 )
 
 func init() {
-	sql.Register("cloudsqlpostgres", &drv{})
+	sql.Register("cloudsqlpostgres", &Driver{})
 }
 
-type drv struct{}
+type Driver struct{}
 
 type dialer struct{}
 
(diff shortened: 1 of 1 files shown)


Carrotman42:
IMO it would be better if `github.com/lib/pq` were updated to allow custom dialers. Then this package wouldn't have to register a new driver just to allow a custom dialer, and this specialized logic could be made to look more like mysql/hook.go.

(although since we've already released a version where we register a database driver called `cloudsqlpostgres` we may be stuck supporting this thing forever anyway...)

luna-duclos:
Any update on this ?

broady:
Hey Luna,

I messaged you on the Gopher slack about this PR. Maybe we can chat there in realtime (probably tomorrow, since I think it's late where you are...)

> it would be better if github.com/lib/pq were updated to allow custom dialers.

It doesn't seem to have gotten much traction:
https://github.com/lib/pq/issues/470

Is there any reason not to expose this Driver except that it'd be better to do it a different way in lib/pq (if they'd even make that change)?

Carrotman42:
> Is there any reason not to expose this Driver except that it'd be better to do it a different way in lib/pq (if they'd even make that change)?

That's my concern I suppose. I was trying to avoid the need to maintain it here. But if people want it and the other repo refuses to maintain it for us, I suppose we could exert the effort to make it better for our consumers.

Carrotman42:
That, plus the database/SQL driver namespace is global, and I don't like adding to the global namespace of possible drivers if we don't have to. I'd rather piggyback off someone else's name to avoid collisions.

broady:
@Carrotman42 any further objection?

Let's get this merged, if not. I think it's fair to say that adding support custom dialer registration to lib/pq is a long shot.

Carrotman42:
Yeah just go ahead - I don't need to be a stick in the mud here for no great reason.

hfwang:
Thanks for your patience Luna!

PR #171, opened 2018-06-01
fix GoogleCloudPlatform/cloudsql-proxy/157

lychung83:
improved error message when gcloud is missing.


lychung83:
Need more clean up.

PR #174, opened 2018-06-08
Add flag to specify IP types to use when connecting

mandyh2018:
Add the user input flag "ip_address_type" to choose IP address for a given instance


hfwang:
The similar change that emgu did for the java socket factory allowed for a list of ip types that way you could use the proxy with instances that had a mix of different ip types.

(The not yet working PR for java is at https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/78)

proxy/certs/certs.go
```diff
@@ -215,9 +217,29 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
 	}
+
+	//Add the option to choose which IP address to use based on user input ip_address_type
+	//Assume IP address type order is not fixed, we need to find the corresponding IP address by iterating IPAddresses
+	userIpAddressType := flag.Lookup("ip_address_type").Value.String()
+	userChosenIpAddress := ""
+	errorMessageIpAddress := ""
+	for _, eachIpAddressOption := range data.IpAddresses {
+		errorMessageIpAddress += "(TYPE: " + eachIpAddressOption.Type + " IP: " + eachIpAddressOption.IpAddress + ") "
+		if(strings.ToLower(eachIpAddressOption.Type) == strings.ToLower(userIpAddressType)) {
```
hfwang: Consider adding "PUBLIC" as an alias for "PRIMARY": see listIpTypes in https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/78/commits/fda79905410f489bcafed5eaed6a8731965ef355

hfwang: nit, I dont think you need the parens around the condition.

proxy/certs/certs.go
```diff
@@ -215,9 +217,29 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
 		logging.Errorf("WARNING: specifying the correct region in an instance string will become required in a future version!")
 	}
-	if len(data.IpAddresses) == 0 || data.IpAddresses[0].IpAddress == "" {
+
+	if len(data.IpAddresses) == 0 {
 		return nil, "", "", fmt.Errorf("no IP address found for %v", instance)
 	}
+
+	//Add the option to choose which IP address to use based on user input ip_address_type
+	//Assume IP address type order is not fixed, we need to find the corresponding IP address by iterating IPAddresses
```
hfwang: space after //

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -82,6 +82,8 @@ can be removed automatically by this program.`)
 	tokenFile = flag.String("credential_file", "", `If provided, this json file will be used to retrieve Service Account credentials.
 You may set the GOOGLE_APPLICATION_CREDENTIALS environment variable for the same effect.`)
 
+	ipAddressType = flag.String("ip_address_type", "Primary", "Default to be primary. Options: Primary, private")
```
hfwang: Nit, given that the enum values in the API are all uppercase, mind making the default "PRIMARY"? (or "PUBLIC", which may make more sense long term)

mandyh2018:
Just update the code changes based on the review comments. Thanks!

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -82,6 +82,8 @@ can be removed automatically by this program.`)
 	tokenFile = flag.String("credential_file", "", `If provided, this json file will be used to retrieve Service Account credentials.
 You may set the GOOGLE_APPLICATION_CREDENTIALS environment variable for the same effect.`)
 
+	ipAddressType = flag.String("ip_address_type", "PRIMARY", "Default to be PRIMARY. Options: PRIMARY / PUBLIC, PRIVATE")
```
hfwang: Can this instead be a list, so if we specified say PRIVATE,PRIMARY then if the instance has a private IP, it would use that, otherwise it would use the primary IP?

proxy/certs/certs.go
```diff
@@ -215,9 +217,35 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
 	}
-	if len(data.IpAddresses) == 0 || data.IpAddresses[0].IpAddress == "" {
+
+	if len(data.IpAddresses) == 0 {
 		return nil, "", "", fmt.Errorf("no IP address found for %v", instance)
 	}
+
+	// Add the option to choose which IP address to use based on user input ip_address_type
+	// Assume IP address type order is not fixed, we need to find the corresponding IP address by iterating IPAddresses
+	flagTypeName := "ip_address_type"
```
hfwang: Instead of needing to mess with the flags like this, I think it would be better to have a field in the RemoteCertSource struct that specified the ip address type preference.

It could then be set by the flag in sidecar mode, as well as settable via code in the library mode.

proxy/certs/certs.go
```diff
@@ -215,9 +234,44 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
 		logging.Errorf("%v", err)
 		logging.Errorf("WARNING: specifying the correct region in an instance string will become required in a future version!")
 	}
-	if len(data.IpAddresses) == 0 || data.IpAddresses[0].IpAddress == "" {
+
+	if len(data.IpAddresses) == 0 {
 		return nil, "", "", fmt.Errorf("no IP address found for %v", instance)
 	}
+
+	// Find the first matching IP address type between user input types and the types that the instance has
```
hfwang: Can you break this into a separate helper function?

proxy/certs/certs.go
```diff
@@ -215,9 +234,44 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
+	IPAddrInUse := ""
+	isFound := false
+	for _, eachIPAddrTypeByUser := range s.IPAddrTypes {
+		for _, eachIPAddrTypeOfInstance := range data.IpAddresses {
+			if strings.ToUpper(eachIPAddrTypeOfInstance.Type) == strings.ToUpper(eachIPAddrTypeByUser) {
+				IPAddrInUse = eachIPAddrTypeOfInstance.IpAddress
+				isFound = true
+				break
```
hfwang: We usually prefer to return early, this will skip all the breaks and conditionals and remove the need for isFound

proxy/certs/certs.go
```diff
@@ -215,9 +234,44 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
+			IPAddrTypesOfInstance += "(TYPE: " + eachIPAddrTypeOfInstance.Type + " IP: " + eachIPAddrTypeOfInstance.IpAddress + ") "
+		}
+
+		IPAddrTypeOfUser := ""
+		for _, eachIPAddrOpt := range s.IPAddrTypes {
+			IPAddrTypeOfUser += eachIPAddrOpt + ","
+		}
+		IPAddrTypeOfUser = strings.TrimSuffix(IPAddrTypeOfUser, ",")
+
+		return nil, "", "", fmt.Errorf("User input IP address type (%v) does not match the instance %v, the instance's IP addresses are {%v}", IPAddrTypeOfUser, instance, IPAddrTypesOfInstance)
```
hfwang: Instead of accumulating the strings and joining with commas manually, you can just use Sprintf("%v") (which fmt.Errorf uses as well), for example:

https://play.golang.org/p/igOzq3tOvQ4

tests/common_open_source.go
```diff
@@ -71,4 +71,4 @@ func compileProxy() (string, error) {
 	}
 
 	return "cloud_sql_proxy", nil
-}
+}
```
hfwang: please revert this change?

proxy/certs/certs.go
```diff
@@ -215,9 +234,44 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
...
 		logging.Errorf("WARNING: specifying the correct region in an instance string will become required in a future version!")
 	}
-	if len(data.IpAddresses) == 0 || data.IpAddresses[0].IpAddress == "" {
+
+	if len(data.IpAddresses) == 0 {
 		return nil, "", "", fmt.Errorf("no IP address found for %v", instance)
 	}
+
+	// Find the first matching IP address type between user input types and the types that the instance has
+	IPAddrInUse := ""
```
hfwang: nit: this should be named ipAddrInUse

hfwang:
Also, it looks like there are still conflicts :(

proxy/certs/certs.go
mandyh2018: Done.

proxy/certs/certs.go
mandyh2018: Done.

proxy/certs/certs.go
mandyh2018: Good idea! For data.IpAddresses, I still use for loop since it is a slice of pointers. Couldn't find a better way for now. Any comment is welcome!

proxy/certs/certs.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/util"
 	"google.golang.org/api/googleapi"
 	sqladmin "google.golang.org/api/sqladmin/v1beta4"
+	"strings"
```
hfwang: yet more nitpicking: import this in the top block of imports please.

proxy/certs/certs.go
```diff
@@ -185,6 +204,27 @@ func parseCert(pemCert string) (*x509.Certificate, error) {
 	return x509.ParseCertificate(bl.Bytes)
 }
 
+// Find the first matching IP address by user input IP address types
+func findIpAddr(s *RemoteCertSource, data *sqladmin.DatabaseInstance, instance string) (ipAddrInUse string, err error) {
```
hfwang: Now that this is extracted into a function, it'd be nice to add tests for it ;)

Also, IMO you may as well make this a function that is part of the RemoteCertSource struct.

mandyh2018: Sure. Will add the test later.

proxy/certs/certs.go
mandyh2018: Sure. Thanks for all these comments. They are really helpful!

hfwang:
LGTM!

PR #178, opened 2018-06-12
Add configurable delay before termination

flori:
With this modification and called with -term-delay=30s cloudsql proxy waits 30 seconds before terminating. This is useful, if it is used as a kubernetes sidecar container and the application server in the main container is still doing some work requiring a database connection before gracefully terminating.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -30,9 +30,11 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/logging"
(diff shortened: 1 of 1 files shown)


flori:
I signed it!

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -373,6 +377,24 @@ func gcloudProject() []string {
 	return []string{data.Core.Project}
 }
 
+func setupTermDelay() {
+	terminated := make(chan bool, 0)
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+
+	go func() {
+		<-terminated
```
hfwang: dumb question, why have two goroutines here? Why not just have this line read

`<-sigs`?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -373,6 +377,24 @@ func gcloudProject() []string {
...
+		time.Sleep(*termDelay)
```
hfwang: This is rather pessimistic, it'd be nice if the client proxy shut off as soon as there were no connections...

Instead of just always sleeping for termDelay, what do you think of adding a ticker that checks if client.ConnectionsCounter == 0, and if so, calls os.Exit(0) early? (If you go that route, take care to also remove the `if c.MaxConnections > 0` that only updates the ConnectionsCounter if needed.

hfwang: Sorry, while I'm hoping for unicorns: we should also call `close(connSrc)` here as well, so that the proxy stops accepting new connections at this point.

hfwang:
flori: I piled a bunch of feature requests on top of your PR. If you feel like I'm asking for too much, let me know!

PR #188, opened 2018-07-02
add support for watching instances via file

night:
The company I work for uses confd for configuration management, and we wanted a way to reload the instances list without restarting the process. This seems to work well enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ cloud_sql_proxy takes a few arguments to configure:
   will be removed from `-dir` as well (unless it was also specified in
   `-instances`), but any existing connections to this instance will NOT be
   terminated.
+* `-instances_file="/path/to/instances/file"`: Specify a file to load and watch for instances configuration changes. It's a comma separated list similar to the `-instances` param value.
 
 Note: `-instances` and `-instances_metadata` may be used at the same time but
 are not compatible with the `-fuse` flag.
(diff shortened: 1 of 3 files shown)


night:
I signed it!

kurtisvg:
@night Your CLA check is still failing - would you please double check the instructions listed in the post and make sure you didn't miss any steps?

night:
@kurtisvg I see the CLA listed under `Corporate Agreements` at https://cla.developers.google.com/clas and the emails match up .. possible there is some kind of delay/bug? I posted right after it was signed.

kurtisvg:
@night Thanks! The bot rechecks when you comment (you may have commented before the change had propagated to the bot the first time)

hfwang:
@night I've been mulling this over for awhile, and I'm a bit hesitant to take this. While I understand why this feature is valuable to you, it feels like the number of ways that you can specify and configure the proxy client is constantly ballooning and that has made things harder to maintain.

I'm curious, for your use case, is it possible to use either fuse, or the instance metadata (which is used for GCE VMs) to solve your problem?

If not, would you mind opening an issue as a way to guage demand and see if other folks are interested in this functionality?

kurtisvg:
We are going to close this for now, please feel free to open an issue to better gauge support if you feel strongly about this feature.

PR #195, opened 2018-08-03
Feature/grpc connection

kirbyquerby:
add ability to connect to proxy server via gRPC on port 3308.


hfwang:
(For context to curious onlookers: this is the result of an intern project. We are experimenting with tunneling database traffic through gRPC instead of a TLS tunnel.)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -501,16 +505,41 @@ func main() {
...
+			IgnoreRegion:   !*checkRegion,
+			UserAgent:      userAgentFromVersionString(),
+			IPAddrTypeOpts: ipAddrTypeOptsInput,
 		}),
 		Conns:              connset,
 		RefreshCfgThrottle: refreshCfgThrottle,
-	}).Run(connSrc)
+	}
+
+	if *usegRPC {
```
hfwang: this branch is pretty unbalanced. Could you either push this into proxyClient.Run or maybe add a proxyClent.RunGrpc(connSrc) instead?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -501,16 +505,41 @@ func main() {
...
+		logging.Infof("Using gRPC\n")
+
+		for conn := range connSrc {
+			rpcProxy, err := grpcproxy.ObtainProxyConnection(grpcproxy.AuthConfig{
+				ProxyClient: proxyClient,
+				Conn:        conn,
+				Port:        grpcPort})
+			if err != nil {
```
hfwang: throwing a fatal error if any connection fails seems pretty harsh. Consider just logging the error?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -501,16 +505,41 @@ func main() {
...
+				log.Fatal(err)
+			}
+
+			logging.Infof("Making new grpc tunnel")
+			err = rpcProxy.CreateTunnel(conn.Conn)
+			if err != nil {
+				logging.Infof("ohno %v", err)
```
hfwang: This error is not very descriptive :)

proxy/certs/certs.go
```diff
@@ -182,6 +182,7 @@ func (s *RemoteCertSource) Local(instance string) (ret tls.Certificate, err erro
 		return err
 	})
 	if err != nil {
+		logging.Infof("createEphemeral failure \n")
```
hfwang: generally speaking, we try not to log anything, and this is sort of like the log and throw antipattern (here's an article about the java equivalent: https://community.oracle.com/docs/DOC-983543#logAndThrow)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hfwang: We talked about this in person a little, the broader point here is that this proxyClient and the grpcproxy should have a more parallel structure. One way to achieve that may be to extract the config retrieving part.

kurtisvg:
Merging into an experimental branch for now.

Issue #128, opened 2017-10-17
Feature Request: Perform a graceful shutdown upon SIGTERM


thrnio:
The method documented in [Connecting from Google Container Engine](https://cloud.google.com/sql/docs/mysql/connect-container-engine) for using Cloud SQL Proxy from GKE is to run the proxy as a sidecar container in your pod alongside your application. For normal use, this works well, but as [this StackOverflow question describes](https://stackoverflow.com/questions/41679364/kubernetes-stop-cloudsql-proxy-sidecar-container-in-multi-container-pod-job), there are problems when trying to do the same in a Kubernetes Job.

The main issue is that for a Job to finish, all the containers in its Pod need to finish and return 0, so when using Cloud SQL Proxy as a sidecar container, even after the application container finishes, the proxy container stays around and Kubernetes believes the Job is not finished. 

What would be helpful is for there to be some way to tell Cloud SQL Proxy to shut itself down cleanly after the application has finished. Perhaps something like a configurable timeout where the proxy shuts down after having no active connections for a period of time, or even some way for the application to explicitly tell the proxy it has finished before the application itself shuts down.

AthenaShi:
This can be done by just SIGTERM the proxy on shutdown.

For more information, (well not directly related):
https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/86

thrnio:
Sending a SIGTERM won't work because it causes Cloud SQL Proxy to exit 2, not exit 0, which Kubernetes will treat as a Job failure, not a success.

EamonKeane:
@thrnio having the same issue, did you find a workaround for this?

EamonKeane:
possible solution here? https://stackoverflow.com/questions/41679364/kubernetes-stop-cloudsql-proxy-sidecar-container-in-multi-container-pod-job

One possible solution would be a separate cloudsql-proxy deployment with a matching service. You would then only need your migration container inside the job that connects to your proxy service.

This comes with some downsides:

higher network latency, no pod local mysql communication
possible security issue if you provide the sql port to your whole kubernetes cluster
If you want to open cloudsql-proxy to the whole cluster you have to replace tcp:3306 with tcp:0.0.0.0:3306 in the -instance parameter on the cloudsql-proxy.

thrnio:
[...]

PR #212, opened 2018-09-25
Terminate gracefully after receiving a SIGTERM signal.

jbrook:
See #128 

Added a SIGTERM handler to the client with a configurable termination grace period (default 20s). During this period the client will refuse new connections. At the end of the period the process will exit. This allows in-flight database operations to complete when running as a sidecar container in Kubernetes and performing rolling updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,13 @@ download the associated JSON file, and set `-credential_file` to the path of the
 JSON file. You can also set the GOOGLE_APPLICATION_CREDENTIALS environment variable
 instead of passing this flag.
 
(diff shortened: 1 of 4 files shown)


jacobat:
This doesn't fully solve the issue AFAICT since other containers might try to establish connections even after being asked to terminate. An example chain of events: Web request hits application server -> pod is asked to terminate -> proxy receives sigterm -> application server tries to establish db connection resulting in error since the connection can't be established.

kurtisvg:
This feature is currently being developed in #206. If there is something about that implementation that doesn't cover your use-case, please let us know in either in #128 or in #206.

PR #218, opened 2018-09-28
Two fixes in GcloudConfig for usability

Carrotman42:
* Print out the error message from Gcloud when it can't print out its config
* Fix the error message printed out when reading config for reasons other than figuring out the gcloud project.


hfwang:
Thanks for this!

PR #222, opened 2018-10-29
refresh SSL certs early

lychung83:
Fixing https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/94. Review done internally.


lychung83:
Did manual testing for making iterative connections (and closing after short period of time) for 205 hours, but no high latency except for the first connection and there's also no connection failure.

PR #237, opened 2019-01-04
Skip checking GCE presence

molerus:
Adds a flag which allows to use gcloud authentication workflow even on instances running on GCE. It provides a workaround to use regular user account on GCE machines that don't have particular user configured as an Owner or Editor of a project.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -85,6 +85,7 @@ can be removed automatically by this program.`)
 	token     = flag.String("token", "", "When set, the proxy uses this Bearer token for authorization.")
 	tokenFile = flag.String("credential_file", "", `If provided, this json file will be used to retrieve Service Account credentials.
 You may set the GOOGLE_APPLICATION_CREDENTIALS environment variable for the same effect.`)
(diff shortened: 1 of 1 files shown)


Carrotman42:
I don't quite see how your change does what it says it does: you are changing how 'onGCE' is initialized, and that is only ever used by checkFlags. In checkFlags, when onGCE is false it just disables some more checks and verifies that the user hasn't set any flags which don't work outside of GCE.

Can you please explain your problem a little more? The GCE-related logic should be a fallback if no other specific flags are set (at least it used to be, I haven't followed along with recent changes...), so I think there's just another flag you can already use.

What "authentication workflow" is gcloud using that you want to use? If there's a service account you could point to that service account file with credential_file.

molerus:
In cloudsql-proxy documentation it says "cloud_sql_proxy takes a few arguments to configure authentication:" followed by a set of cases with no explicit precedence. The authentication workflow I was referring to is in point 4: gcloud user credentials (set from gcloud auth login).
It doesn't work however if the instance n which one wants to start the proxy is already running on GCE. In such case it tries to use credentials stored in VM's metadata and skips **gcloud auth** workflow. By introducing the flag in this PR I'm making it to skip the check.

hfwang:
@molerus is this still needed?

PR #244, opened 2019-01-23
Update download_proxy.sh

MathijsvVelde:
Added space to the `grep` pattern to exclude other lines containing "project" like `account = account@my-first-project-12345.iam.gserviceaccount.com`


MathijsvVelde:
I signed it!

kurtisvg:
Deduping to #245

PR #245, opened 2019-01-23
Update download_proxy.sh

MathijsvVelde:
Looking in wrong column
```bash
$ gcloud sql instances list
NAME    DATABASE_VERSION  LOCATION        TIER              PRIMARY_ADDRESS  PRIVATE_ADDRESS  STATUS
db-5-7  MYSQL_5_7         europe-west4-b  db-n1-standard-1  35.204.7.222     -                RUNNABLE
db-5-6  MYSQL_5_6         us-central      D0                -                -                RUNNABLE
```


kurtisvg:
Hey @MathijsvVelde thanks for taking the time to open this. 

We discussed this at length today and decided to just remove this script entirely (#247) since it's very useful over just picking the correct exe and seems likely to break again in the future.

PR #260, opened 2019-03-12
Clarify usage of `-instances_metadata` fixes #259

### LINKED ISSUE #259: Clarify documentation on expected value for `-instances_metadata`
When using `-instances_metadata` README.md states that:

> The given GCE metadata key will be polled...

To me this implies only the name of the metadata key is required. For example 

`-instances_metadata my-sql-instances`

In actual fact a path must be specified relative to `http://metadata.google.internal/computeMetadata/v1/`. In this case the expected input should be 

`-instances_metadata instance/attributes/my-sql-instances`

Documentation should be updated to reflect this.

matalo33:


matalo33:
I signed it!

broady:
that's pretty odd that you need to prefix with `instance/attributes` since the flag is called `instances_metadata`, but great catch. thank you.

Carrotman42:
The "instances" in --instances_metadata refers to a Cloud SQL database instance, and is similar to the --instances flag.

Issue #241, opened 2019-01-18
Feature Request: set socket name


wedi:
It would be cool if the name of the socket file could be changed via command line.

Reasoning: Third parties rely (way too) often on hard coded, distribution depending values like `/tmp/mysql.sock`, `/var/run/mysqld/mysqld.sock` etc. By being able to change the socket name no changes to third party software are neccessary and a silent db change in the background becomes possible. It's no problem to change the socket name in a full fledged sql installation either(`my.cnf`). Of course one can add a symlink but this is not practical when running the cloudsql proxy docker image.

Carrotman42:
I thought this was allowed already, but looking through the code I believe I am wrong. My suggestion would be to augment the existing ability to specify a TCP port to listen to, and instead handle 'unix' sockets as well. Something like:

    -instances=my-project:my-region:sql-inst=unix:/tmp/mysql.sock

For reference, this is the code that would need to be updated; it's not too complex, if you'd like to send in a PR: 

https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/2ec72f445aed645b40aeea03e73e6acff892d2e1/cmd/cloud_sql_proxy/proxy.go#L226-L245

kurtisvg:
Fixed in #289

PR #265, opened 2019-03-17
Allow to set the socket name per instance configuration

wedi:
This pull requests allows to set a custom socket name for each instance.
Example: -instances=my-project:us-central1:sql-inst=unix:custom_socket_name

This pull request also allows to override `-dir` by specifying an absolute path. This would come in handy for me but I can easily remove it from this pull request if it's not desired for some reason.

Resolves: #241

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -233,11 +233,19 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
 			// No "host" part of the address. Be safe and assume that they want a
 			// loopback address.
 			ret.Network = spl[0]
-			addr, ok := loopbackForNet[spl[0]]
-			if !ok {
-				return ret, fmt.Errorf("invalid %q: unrecognized network %v", instance, spl[0])
+			if spl[0] == "unix" {
+				if strings.HasPrefix(spl[1], "/") {
+					ret.Address = spl[1]
+				} else {
+					ret.Address = filepath.Join(dir, spl[1])
+				}
+			} else {
+				addr, ok := loopbackForNet[spl[0]]
+				if !ok {
+					return ret, fmt.Errorf("invalid %q: unrecognized network %v", instance, spl[0])
+				}
+				ret.Address = fmt.Sprintf("%s:%s", addr, spl[1])
 			}
-			ret.Address = fmt.Sprintf("%s:%s", addr, spl[1])
 		case 3:
 			// User provided a host and port; use that.
 			ret.Network = spl[0]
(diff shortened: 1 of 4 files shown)


wedi:
Signed the CLA.

wedi:
Hi. Is there anything I can do to push this ahead?

kurtisvg:
Hey @wedi,

Sorry for dropping the ball on this. I've been busy with Next 2019 for the past few weeks, but I definitely should have communicated that sooner.  Please give me a few days and I'll make sure this is reviewed and discussed sometime next week.

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -233,11 +233,19 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
...
 			// loopback address.
 			ret.Network = spl[0]
-			addr, ok := loopbackForNet[spl[0]]
-			if !ok {
-				return ret, fmt.Errorf("invalid %q: unrecognized network %v", instance, spl[0])
+			if spl[0] == "unix" {
+				if strings.HasPrefix(spl[1], "/") {
+					ret.Address = spl[1]
+				} else {
+					ret.Address = filepath.Join(dir, spl[1])
```
broady: do we want to append `".s.PGSQL.5432"` for postgres instances?

broady:
LGTM, just one question.

cmd/cloud_sql_proxy/proxy.go
wedi: I have no experience with google postgres instances so far. Is this `.s.PGSQL.5432` string appended to the socket name when postgres is used? This shouldn't happen as this pull request is about setting a completely custom socket name when one cannot (or doesn't want to) change the behaviour of some third party software.

kurtisvg: PostgreSQL treats the unix socket location as a directory, and specifies the socket name should be `.s.PGSQL.5432` - see "unix_socket_directories" [section](https://www.postgresql.org/docs/11/runtime-config-connection.html). The proxy currently automatically appends this suffix [here](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/6ac0b49e7197a3f91a3d438204a45e0032af2de9/cmd/cloud_sql_proxy/proxy.go#L276).

Most client side connectors treat it as a directory and automatically append the `.s.PGSQL.5432` automatically, which means if the user specifies `/my/socket/file`, it is actually using `/my/socket/file/.s.PGSQL.5432`. Since we already automatically append in one scenario, it seems best that we should continue to do so for this one to remain consistent. 

It might make more sense to move the append into the file creation to consolidate, which I think is [here](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/dc8853bb2cd44140b12bffdaa5603aa99647fcd5/cmd/cloud_sql_proxy/proxy.go#L136).

kurtisvg:
@wedi I've opened #289 to update this PR based on feedback - if you could comment on that PR that you are okay with me borrowing your commits, it'll make the CLAbot happy :)

wedi:
Done. Thanks for finishing this.

kurtisvg:
Closing in favor of #289

PR #266, opened 2019-03-20
Fix credential flow.

### LINKED ISSUE #263: -credential_file option does not seem to work
I've been unable to get the `-credential_file` option to work as documented. When I set `GOOGLE_APPLICATION_CREDENTIALS` to the path of a JSON file containing credentials for a service account, `cloud_sql_proxy` works as expected. However, when I unset `GOOGLE_APPLICATION_CREDENTIALS` and use the `-credential_file` option, I get the following error:

```
2019/03/15 11:46:32 current FDs rlimit set to 28494, wanted limit is 8500. Nothing to do here.
2019/03/15 11:46:32 using credential file for authentication; email=orion-966@western-notch-111123.iam.gserviceaccount.com
2019/03/15 11:46:32 google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
exit status 1
```

Adding some print debugging shows that the two approaches make it to the same point in `authenticatedClientFromPath`. Things seem to go haywire `goauth.JWTConfigFromJSON`, where I gave up on debugging. While `goauth.JWTConfigFromJSON` does not return an error, it seems to rely on `GOOGLE_APPLICATION_CREDENTIALS` in some way, resulting in the `could not find default credentials` error above.

kurtisvg:
Fixes #263.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -280,12 +280,12 @@ func authenticatedClientFromPath(ctx context.Context, f string) (*http.Client, e
 
 func authenticatedClient(ctx context.Context) (*http.Client, error) {
 	if *tokenFile != "" {
-		authenticatedClientFromPath(ctx, *tokenFile)
+		return authenticatedClientFromPath(ctx, *tokenFile)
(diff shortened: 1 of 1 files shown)


kurtisvg:
@hfwang PTAL - this needs to be in before next release

PR #268, opened 2019-03-27
test VM uses debian-9 and add flag for public/private SSH

hfwang:
Use debian-9 in tests, and add a flag that allows controlling whether the test VM has a public IP or not. If the instance does not have a public IP, then the private IP will be used when SSHing


hfwang:
Looks like debian-8 is starting to have complaints about pulling debs, the private IP is necessary to get tag-based firewall rules to apply.

PR #269, opened 2019-04-04
Proxy rejects requests for first generation instances.

hfwang:
First generation instances are not supported by the proxy, so this causes connections to them to fail slightly faster (otherwise we need to wait for the connection to timeout!)


hfwang:
The log and error have slightly different content... @Carrotman42 are you OK with that? Or would you rather I remove the logs altogether?

PR #273, opened 2019-05-02
Add InitWithClient and deprecate InitClient because we don't want to copy Client's field sync.RWMutex by value

mandyh2018:
Init proxy client using pointer because we don't want to copy its filed sync.RWMutex by value.


hfwang:
I think this may not be allowed actually since it changes the API surface...

mandyh2018:
hmm, I can also add a new API.

broady:
Can you tell us about the motivation behind this change? Is there a bug?

Carrotman42:
Copying a sync.RWMutex is not generally safe, although I think copying an unused (i.e. zero-value) RWMutex can't be noticed by the internal code used to detect this copying. In any case this was definitely an oversight when I wrote this API and the API should be either changed or deprecated/deleted-eventually.

kurtisvg:
broady@ and I discussed this today and the conclusion was that we should defiantly make this change, but play it safe and create a new function and deprecate the old one.

mandyh2018:
@kurtisvg Thanks! Is it part of your plan? Or I will do it later. :-)

kurtisvg:
@mandyh2018 If you would like to make the change you are welcome to, otherwise I will update this PR sometime before the next release.

mandyh2018:
broady@ are we okay with changing the existing API? I thought we want to add a new one instead.

mandyh2018:
kurtisvg@ Are you working on this? If not, I will add a new API this week. :-)

kurtisvg:
My plan is to just add a new function (likely `InitWithClient`) and mark the old one with a deprecated warning. I'm hoping to do it sometime this week, but please feel free to do it sooner if you have time :)

mandyh2018:
Let me work on this then. I have some time today and tomorrow. :-) So I just close this pull request and send a new one?  kurtisvg@

kurtisvg:
You can just update this PR with the changes

mandyh2018:
I updated this PR, please let me know if any comment. Thanks!

proxy/proxy/dial.go
```diff
@@ -80,12 +81,22 @@ func Init(auth *http.Client, connset *ConnSet, dialer Dialer) {
 
 // InitClient is similar to Init, but allows you to specify the Client
 // directly.
+
+// Deprecated: Client contains sync.RWMutex, which should not be copied by value.
```
broady: just "Deprecated: Use InitWithClient instead." i think.

mandyh2018: Done. Thx!

mandyh2018:
Thank you! I will wait until I become the authorized user?

mandyh2018:
I can see the warning:
Merging is blocked
The base branch restricts merging to authorized users. 
You’re not authorized to merge this pull request.

broady:
Thanks! Merged.

mandyh2018:
Thank you!

PR #283, opened 2019-06-06
Added the error handling of the incorrect "host" flag setting.

mandyh2018:
For the flag "host", the value should always end with '/'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,19 +82,6 @@ instead of passing this flag.
     ./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:3306 &
     mysql -u root -h 127.0.0.1
 
-## To use inside a Go program:
-If your program is written in [Go](https://golang.org) you can use the Cloud SQL Proxy as a library,
-avoiding the need to start the Proxy as a companion process.
-
-### MySQL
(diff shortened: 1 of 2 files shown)


cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -391,6 +391,12 @@ func main() {
 		return
 	}
 
+	if !strings.HasSuffix(*host, "/") {
```
kurtisvg: This should probably be at least below the logging set up (after L412), so that the log correctly reflects those options.

Carrotman42:
In another repo I work in, we simply utilize squash-and-merge to keep the master history free to lots of commits. I suggest going with that format instead of force-pushing: force-pushing can be dangerous if you're not careful.

mandyh2018:
Thanks! Will use squash-and-merge next time. :-)

kurtisvg:
We also utilize squash-and-merge when merging into master, but since PR are proposing changes to master, it is much easier to review when the commits only reflect the changes being proposed. Merge commits can make it harder to differentiate proposed changes from already accepted changes.

mandyh2018:
I filed a new PR #285. Thanks!

PR #285, opened 2019-06-07
Added the error handling of the incorrect host flag setting

mandyh2018:
For the flag "host", the value should always end with '/'.


mandyh2018:
Thanks! Could you please help to merge it since I am not authorized to do so? :-)

mandyh2018:
Thanks a lot!

PR #287, opened 2019-06-11
Add -remote_host_override for connecting to proxied instances

cian:
We have a CloudSQL DB with only a Private IP and a requirement to connect through the proxy only (for audit). When DBAs connect to this DB from their workstation (outside of the VPC), they must port-forward 3307 to their local machine over SSH as the private IP is not directly accessible. This flag allows them to tell the cloud_sql_proxy what `host:port` the CloudSQL server is on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ cloud_sql_proxy takes a few arguments to configure what instances to connect to 
   IP. Defaults to `PUBLIC,PRIVATE`
 * `-term_timeout=30s`: How long to wait for connections to close before shutting
   down the proxy. Defaults to 0.
+* `-remote_host_override=localhost:3307`: Override the API-supplied IP address for the CloudSQL server and instead connect to the supplied hostname:port. This can be useful when the CloudSQL server has a private IP only, and you are connecting from outside the VPC using SSH port forwarding. Defaults to API supplied IP address.
 
 Note: `-instances` and `-instances_metadata` may be used at the same time but
 are not compatible with the `-fuse` flag.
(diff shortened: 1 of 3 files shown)


broady:
I suspect it should go from:

```
./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:3306 &
```

to something like:

```
./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:0.0.0.0:3306 &
```

or maybe just

```
./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=tcp:3306 -host 0.0.0.0 &
```

As @kurtisvg implies, port should only be set via the -instances flag.

kurtisvg:
@broady This is setting the port that's being used to connect to the Cloud SQL instance, rather than the port that the proxy is used to listen (which is the one the -instances flag currently sets)

cian:
@kurtisvg I had the same concern myself actually, and two potential thoughts for the multiple instance problem:
1) Assuming they're on the same VPC I could add a flag to use `golang.org/x/net/proxy.SOCKS5`.
eg. `-remote_host_proxy=localhost:5000`

2) Support a map of some type either as an optional extension to the -instances flag:
eg. `-instances=my-project:us-central1:sql-inst=tcp:3306;src-tcp:localhost:3307,my-other-project:us-central1:sql-inst=tcp:3308;src-tcp:localhost:3309`
 or via another flag:
`-remote_host_override=my-project:us-central1:sql-inst=tcp:localhost:3307,my-other-project:us-central1:sql-inst=tcp:localhost:3309`

But was conscious that these would both involve a much larger change to the tool to support a need that we don't have, and I wasn't sure anyone else did.

If you think this multiple instance support from a single running proxy instance is essential, I'm happy to add this to the PR.

broady:
Would it make sense to support the ALL_PROXY environment variable so you can use a SOCKS proxy?

https://godoc.org/golang.org/x/net/proxy#FromEnvironment

Presumably you also want the calls to sqladmin API to go via your proxy?

Carrotman42:
> Presumably you also want the calls to sqladmin API to go via your proxy?

I'm not sure this would be a good idea, unless I'm missing something: afaik the sqladmin API is indifferent to the IP address you're coming from, so pushing the sqladmin API calls through the proxy would just add overhead. I don't see the positive side.

enocom:
Closing as stale.

PR #289, opened 2019-06-13
Set custom per instance socket address

kurtisvg:
Fork of #265, updated to include automatic addition of the suffix for Postgres instances.

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -231,42 +231,22 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
 	if proj == "" || name == "" {
 		return instanceConfig{}, fmt.Errorf("invalid instance connection string: must be in the form `project:region:instance-name`; invalid name was %q", args[0])
 	}
-
-	eM := "invalid instance options: must be in the form `unix:/path/to/socket`, `tcp:port`, `tcp:host:port`; invalid option was %q"
 	if len(args) == 1 {
 		// Default to listening via unix socket in specified directory
 		ret.Network = "unix"
 		ret.Address = filepath.Join(dir, instance)
 	} else {
 		// Parse the instance options if present.
-		opts := strings.Split(args[1], ":")
+		opts := strings.SplitN(args[1], ":", 2)
 		ret.Network = opts[0]
+		var err error
 		if ret.Network == "unix" {
-			// Listen via specified unix socket.
-			if len(opts) != 2 {
-				return instanceConfig{}, fmt.Errorf(eM, args[1])
-			}
-			if strings.HasPrefix(opts[1], "/") {
-				ret.Address = opts[1] // Root path.
-			} else {
-				ret.Address = filepath.Join(dir, opts[1])
-			}
+			ret.Address, err = parseUnixOpts(dir, opts)
 		} else {
-			// Listen via the specified dial address.
-			switch len(opts) {
-			case 2:
-				// No "host" part of the address. Be safe and assume that they want a loopback address.
-				addr, ok := loopbackForNet[opts[0]]
-				if !ok {
-					return ret, fmt.Errorf("invalid %q: unrecognized network %v", args[1], opts[0])
-				}
-				ret.Address = net.JoinHostPort(addr, opts[1])
-			case 3:
-				// User provided a host and port; use that.
-				ret.Address = net.JoinHostPort(opts[1], opts[2])
-			default:
-				return instanceConfig{}, fmt.Errorf(eM, args[1])
-			}
+			ret.Address, err = parseTcpOpts(opts)
+		}
+		if err != nil {
+			return instanceConfig{}, err
 		}
 	}
 
@@ -298,6 +278,33 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
 		return ret, fmt.Errorf("invalid %q: unsupported network: %v", instance, ret.Network)
 	}
 	return ret, nil
+}
+
+// parseUnixOpts parses the instance options when specifying unix socket options.
+func parseUnixOpts(dir string, opts []string) (string, error) {
+	// Listen via specified unix socket.
+	if len(opts) != 2 {
+		return "", fmt.Errorf("invalid instance options: must be in the form `unix:/path/to/socket`, `tcp:port`, `tcp:host:port`; invalid option was %q", strings.Join(opts, ":"))
+	}
+	if strings.HasPrefix(opts[1], "/") {
+		return opts[1], nil // Root path.
+	}
+	return filepath.Join(dir, opts[1]), nil
+}
+
+// parseTcpOpts parses the instance options when specifying tcp port options.
+func parseTcpOpts(opts []string) (string, error) {
+	if strings.Contains(opts[1], ":") {
+		// User provided a host and port; use that.
+		return opts[1], nil
+	} else {
+		// No "host" part of the address. Be safe and assume that they want a loopback address.
+		addr, ok := loopbackForNet[opts[0]]
+		if !ok {
+			return "", fmt.Errorf("invalid %q: unrecognized network %v", strings.Join(opts, ":"), opts[0])
+		}
+		return net.JoinHostPort(addr, opts[1]), nil
+	}
 }
 
 // parseInstanceConfigs calls parseInstanceConfig for each instance in the
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -250,7 +250,7 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
 				ret.Address = filepath.Join(dir, opts[1])
 			}
 		} else {
-			ret.Address, err = parseTcpOpts(opts[0], opts[1])
+			ret.Address, err = parseTCPOpts(opts[0], opts[1])
 		}
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/cloud_sql_proxy/proxy.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/fuse"
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy"
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/util"
-	sqladmin "google.golang.org/api/sqladmin/v1beta4"
```
broady: revert

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+		return instanceConfig{}, fmt.Errorf("invalid instance argument: must be either form - `<instance_connection_string>` or `<instance_connection_string>=<options>`; invalid arg was %q", instance)
+	}
+	// Parse the instance connection name - everything before the "=".
+	ret.Instance = args[0]
+	proj, _, name := util.SplitName(ret.Instance)
+	if proj == "" || name == "" {
+		return instanceConfig{}, fmt.Errorf("invalid instance connection string: must be in the form `project:region:instance-name`; invalid name was %q", args[0])
+	}
+	// Parse the instance options - everything before the "=".
+	eM := "invalid instance options: must be in the form `unix:/path/to/socket`, `tcp:port`, `tcp:host:port`; invalid option was %q"
```
broady: what is eM? error message? please inline where it's used (L254)

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+	if len(args) == 2 {
```
broady: a bit clearer to have the shorter code path first, i think.

if len(args) == 1 {
  // default unix
} else {
  // parse options
}

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+			}
+			ret.Network = "unix"
+			if strings.HasPrefix(opts[1], "/") {
+				ret.Address = opts[1] // Root path.
+			} else {
+				ret.Address = filepath.Join(dir, opts[1])
+			}
+		} else {
+			// Listen via the specified dial address.
+			ret.Network = opts[0]
```
broady: move up to L237 (no need for L243)

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+			switch len(opts) {
+			default:
```
broady: default at the bottom

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+				return instanceConfig{}, fmt.Errorf(eM, args[1])
+			case 2:
+				// No "host" part of the address. Be safe and assume that they want a loopback address.
+				addr, ok := loopbackForNet[opts[0]]
+				if !ok {
+					return ret, fmt.Errorf("invalid %q: unrecognized network %v", args[1], opts[0])
```
broady: s/unrecognized/unsupported/

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+				}
+				ret.Address = fmt.Sprintf("%s:%s", addr, opts[1])
+			case 3:
+				// User provided a host and port; use that.
+				ret.Address = fmt.Sprintf("%s:%s", opts[1], opts[2])
```
broady: use net.JoinHostPort

https://golang.org/pkg/net/#JoinHostPort

cmd/cloud_sql_proxy/proxy.go
broady: also, unfortunate this doesn't support ipv6 addresses very well.

cmd/cloud_sql_proxy/proxy.go
kurtisvg: It's used on L241 and L254 - I would prefer to keep the same error for both lines

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
-		if proj == "" || name == "" {
-			return instanceConfig{}, fmt.Errorf("invalid instance name: must be in the form `project:region:instance-name`; invalid name was %q", instance)
-		}
-		// We allow people to omit the region due to historical reasons. It'll
-		// fail later in the code if this isn't allowed, so just assume it's
-		// allowed until we actually need the region in this API call.
-		in, err := sql.Instances.Get(proj, name).Do()
-		if err != nil {
+	// Use the SQL Admin API to verify compatibility with the instance.
+	sql, err := sqladmin.New(cl)
```
broady: very surprising for HTTP requests to happen inside a "parseX" function. split them out?

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+	if err != nil {
+		return instanceConfig{}, err
+	}
+	sql.BasePath = *host
+	inst, err := sql.Instances.Get(proj, name).Do()
```
broady: probably want some retries here

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+	if err != nil {
+		return instanceConfig{}, err
+	}
+	if inst.BackendType == "FIRST_GEN" {
+		logging.Errorf("WARNING: proxy client does not support first generation Cloud SQL instances.")
```
broady: doesn't look like a warning. but i notice there's the word "WARNING" everywhere even though things are not warnings but actual fatal errors.

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,77 @@ var validNets = func() map[string]bool {
...
+		return instanceConfig{}, fmt.Errorf("%q is a first generation instance", instance)
+	}
+	// Postgres instances use a special suffix on the unix socket.
+	// See https://www.postgresql.org/docs/11/runtime-config-connection.html
+	if ret.Network == "unix" && strings.HasPrefix(strings.ToLower(inst.DatabaseVersion), "postgres") {
+		// Verify the folder exists.
```
broady: s/folder/directory/

cmd/cloud_sql_proxy/proxy.go
broady: ok, please give it a better name or a concrete type, like:

```
type fooError {
  invalidOption string
}

func (e fooError) Error() string {
  return fmt.Sprintf(...)
}
```

cmd/cloud_sql_proxy/proxy.go
kurtisvg: leaving: existing behavior

cmd/cloud_sql_proxy/proxy.go
kurtisvg: leaving: existing behavior

cmd/cloud_sql_proxy/proxy.go
kurtisvg: leaving: existing behavior

cmd/cloud_sql_proxy/proxy.go
kurtisvg: leaving: existing behavior

wedi:
Thanks for stepping in. Ofc I'm fine with that. 👍

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -285,6 +280,33 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
...
+		return "", fmt.Errorf("invalid instance options: must be in the form `unix:/path/to/socket`, `tcp:port`, `tcp:host:port`; invalid option was %q", strings.Join(opts, ":"))
+	}
+	if strings.HasPrefix(opts[1], "/") {
+		return opts[1], nil // Root path.
+	}
+	return filepath.Join(dir, opts[1]), nil
+}
+
+// parseTcpOpts parses the instance options when specifying tcp port options.
+func parseTcpOpts(opts []string) (string, error) {
```
broady: `(net, addrOpt string)`

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -221,62 +221,57 @@ var validNets = func() map[string]bool {
...
+		return instanceConfig{}, fmt.Errorf("invalid instance connection string: must be in the form `project:region:instance-name`; invalid name was %q", args[0])
+	}
+	if len(args) == 1 {
+		// Default to listening via unix socket in specified directory
+		ret.Network = "unix"
+		ret.Address = filepath.Join(dir, instance)
 	} else {
-		sql, err := sqladmin.New(cl)
+		// Parse the instance options if present.
+		opts := strings.SplitN(args[1], ":", 2)
```
broady: bounds check on opts (len == 2)

PR #310, opened 2019-09-20
Bubble err up on fail to parse instance configs instead of ignoring

### LINKED ISSUE #288: cloudsql-proxy gets into bad state if error occurs while connecting
I have a systemd service that simply runs cloud_sql_proxy. If the network isn't available when the service is started, I see the following messages:

```
2019/06/12 10:10:27 Rlimits for file descriptors set to {&{8500 1048576}}
2019/06/12 10:10:40 errors parsing config:
	Get [... project url ommitted ...]: dial tcp: lookup www.googleapis.com on 127.0.1.1:53: server misbehaving
2019/06/12 10:10:40 Ready for new connections
```

I have the cloud_sql_proxy configured to accept local connections via a unix socket file, like so:
```
cloud_sql_proxy -dir="/tmp/cloudsql" -instances="${OUR_DB_INSTANCE}"
```

After the series of events above, the cloud_sql_proxy is in an unrecoverable state. It doesn't retry to connect, and the unix socket file doesn't exist, so my client application can't connect to it.

Expected behaviour:
The cloud_sql_proxy should either exit with an error code (so systemd can restart it), or it should retry when there is an error connecting.

My work-around for now is to create a "watchdog" service that restarts the cloud_sql_proxy service if the unix socket directory is empty after more than 10 seconds, but this is obviously a hack.

EDIT:
The socket file does actually exist, but it refuses connections (so my original hack didn't work) when cloud_sql_proxy is in this state. I get the following python exception when I try to connect to it:
```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not connect to server: Connection refused
```

terev:
Fixes #288 

This change causes the program to exit on error during initial startup. This will prevent the container from entering a bad state on startup after a failed authentication, network error etc.

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -334,7 +334,7 @@ func CreateInstanceConfigs(dir string, useFuse bool, instances []string, instanc
 
 	cfgs, err := parseInstanceConfigs(dir, instances, cl)
 	if err != nil {
-		logging.Errorf("%v", err)
+		return nil, err
 	}
 
 	if dir == "" {


terev:
Are commits in master built and pushed to gcr? Or is it just tagged commits?

kurtisvg:
Just tagged commits - but there is a Dockerfile in the repo now if you want to build and run locally

brugz:
I have some follow up questions:

1) How was this tested? I see that there were no changes to the associated test code. Perhaps we should write a test for this where authentication or other fails.

2) This looks like it will bubble up the err in the event that setup fails. I don't think cloud-sql-proxy will exit if some kind of transient behavior puts the binary into a bad state. Is there anything we can do about that?

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -334,7 +334,7 @@ func CreateInstanceConfigs(dir string, useFuse bool, instances []string, instanc
 
 	cfgs, err := parseInstanceConfigs(dir, instances, cl)
 	if err != nil {
-		logging.Errorf("%v", err)
+		return nil, err
```
brugz: Can we test this error handling at all?

terev:
1. I tested this in a kubernetes pod where istio is installed. Previously the proxy would output an error and continue to say it's accepting connections but with this change the container exited and restarted until the istio sidecar was ready and forwarding traffic to Google apis. Sure I can have a look at adding a test for auth errors and maybe network errors.

kurtisvg:
> 2. This looks like it will bubble up the err in the event that setup fails. I don't think cloud-sql-proxy will exit if some kind of transient behavior puts the binary into a bad state. Is there anything we can do about that?

As mentioned in the issue, this bubbles it up to this  [`log.fatal`](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/a5559d6926be506ca859927b9ee94e2a2e6515f4/cmd/cloud_sql_proxy/cloud_sql_proxy.go#L474), which causes the process to exit. Previously it would be ignored, which is what causes the bad state in the first place.

rvora:
I can confirm that our random auth errors when deployed with istio-proxy were mitigated with this change. cloud-sql-proxy may restart once or twice but then it works instead of client service crashing continuously due to inability of cloud-sql-proxy to get connections.

terev:
At some point it would be nice if the proxy were to optionally retry network errors in order to improve boot time of the container. Any eta of when this change might land?

brugz:
I checked in with some folks at Google; soft-eta is some time in the next 2 weeks, barring any major complications.

Until then, you can build and deploy with local edits using the dockerfile :) 

Also, I agree that occasional retries make sense. This was what I was alluding to here:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/310#issuecomment-535348290

kurtisvg:
@broady - FYI, I did verify this is only used `main` as we discussed offline.

Stono:
Hey folks, we're using the latest cloudsql proxy and sometimes see _nothing_ on startup, it just hangs, where we would typically see:

```
2019/10/22 11:00:11 current FDs rlimit set to 1048576, wanted limit is 8500. Nothing to do here.
2019/10/22 11:00:11 using credential file for authentication; email=email here
019/10/22 11:00:11 Listening on 127.0.0.1:3306 for instance here
2019/10/22 11:00:11 Ready for new connections
```

Carrotman42:
Please file a new issue.

When you are encountering this issue, please send SIGQUIT to the binary to
get it to exit and print stack traces so that we can see what is happening.
Also, please report your OS, platform, and other information that can help
us classify the problem.

PR #311, opened 2019-09-23
Return errors, and avoid bad updates

bajb:
If an error is hit while parsing instance configurations, this error should bubble up.

Also, while watching instances, if the instance configs error, we should assume the instance list is bad, and not process the result.

### DIFF
--- a/cmd/cloud_sql_proxy/proxy.go
+++ b/cmd/cloud_sql_proxy/proxy.go
@@ -336,7 +336,6 @@ func CreateInstanceConfigs(dir string, useFuse bool, instances []string, instanc
 
 	cfgs, err := parseInstanceConfigs(dir, instances, cl)
 	if err != nil {
-		logging.Errorf("%v", err)
 		// Error when unable to correctly parse the instance configuration
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


bajb:
@googlebot I signed it!

kurtisvg:
It looks too me like skipping is the correct option. 

This looks right, but I'm not quite sure why we are re-parsing the configs after the initial start up. This might be related to the flag for project wide instances, but at first glance seems to be occurring regardless of that flag setting.

Carrotman42:
> This looks right, but I'm not quite sure why we are re-parsing the configs after the initial start up. 

This is to support `-instances_metadata` [0], which wants to call `WatchInstances` (which launches the `watchInstancesLoop` routine which was modified in this PR).

[0] https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/a5559d6/cmd/cloud_sql_proxy/cloud_sql_proxy.go#L70

bajb:
We are using the instances_metadata for our setup.  Based on our logs: https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/288#issuecomment-532652232  the initial create looks like it worked fine, but the loop has then failed to parse the configs.

kurtisvg:
@bajb - would you mind rebasing? I would do it for you, but I'd prefer not to update your master branch without asking 😄

bajb:
Sure, would you prefer to keep or remove the logging?

bajb:
I have kept it in for now, as it would be helpful for us to see when these errors happen.  I can remove if needed

kurtisvg:
I think prefer remove - it will get logged one step up, so it becomes redundant.

kurtisvg:
@bajb Thanks!

PR #338, opened 2019-12-27
Update dependencies and correct base url.

### LINKED ISSUE #331: Cloud Proxy 404s on SQL Admin API call - client base path has updated
https://github.com/googleapis/google-api-go-client/commit/8a410c21381766a810817fd6200fce8838ecb277#diff-fc3696d32750cf4e3f741d7d50fda5f4L79-R79
the diff is too big, but 5 days ago this happened:
```diff
- const basePath = "https://www.googleapis.com/sql/v1beta4/"
+ const basePath = "https://sqladmin.googleapis.com/"
```
And cloudsql proxy dialer ends up here:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/46c5533ff5bad7ea6a5e4966a670249d299f5880/proxy/proxy/dial.go#L75

Which creates the cert source with forced API path which is now wrong since google API version v0.14.0.

Note that you don't actually need to provide the path, empty string should work fine and it will use the default from the api:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/46c5533ff5bad7ea6a5e4966a670249d299f5880/proxy/certs/certs.go#L83-L89

As a consequence of that if using default dialer proxy with latest API, you'll get errors like so:
> The requested URL <code>/**sql/v1beta4/sql/v1beta4**/projects/{project_id}/instances/{instance}/createEphemeral?alt=json&amp;prettyPrint=false</code> was not found on this server.

kurtisvg:
Update dependencies and correct base url for latest Cloud SQL Admin API library.

Fixes #331.


kurtisvg:
@broady - PTAL

PR #340, opened 2020-01-03
Added skip_failed_instance_config as a flag to to skip config validation

hilts-vaughan:
This will let you skip failed instance configurations, while emitting a message if you have them and still run the proxy. #340 

I ran some tests with invalid strings and by specifying instances that were deleted, seemed good to me. They error as I would expect. Since I am not a power user though, I would not mind another pass through from someone that is. :)


README.md
```diff
@@ -89,7 +91,7 @@ instead of passing this flag.
     # For programs which require a the Unix Domain Socket at a specific location, set an absolute path (overrides -dir):
     ./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst=unix:/my/custom/sql-socket &
     mysql -u root -S /my/custom/sql-socket
-    
+
```
hilts-vaughan: Is this my editor? Do you want me to revert this?

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -328,16 +328,21 @@ func parseInstanceConfigs(dir string, instances []string, cl *http.Client) ([]in
...
-// instanceConfig.
+// instanceConfig. It is possible for the instanceConfig to be empty if no valid
+// configurations were specified, however `err` will be set.
 func CreateInstanceConfigs(dir string, useFuse bool, instances []string, instancesSrc string, cl *http.Client) ([]instanceConfig, error) {
+	// Used to signify that there was no valid parsed config
+	var emptyConfigs []instanceConfig
+
 	if useFuse && !fuse.Supported() {
-		return nil, errors.New("FUSE not supported on this system")
+		return emptyConfigs, errors.New("FUSE not supported on this system")
```
hilts-vaughan: I thought an empty slice would be nicer than `nil`, though I am not sure what the "Go" way is. To be honest, I did this before I found `len` of a nil array was 0.

Is it more golang-like to revert this to using `nil`? I'm new to Golang.

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -328,16 +328,21 @@ func parseInstanceConfigs(dir string, instances []string, cl *http.Client) ([]in
...
 	}
 
 	cfgs, err := parseInstanceConfigs(dir, instances, cl)
 	if err != nil {
 		// Error when unable to correctly parse the instance configuration
-		return nil, err
+		// However, return the configuration of what was parsed correctly
+		return cfgs, err
```
hilts-vaughan: This is the core of the change -- we're returning what we could parse right here and then using that above.

README.md
kurtisvg: Nah it's fine.

cmd/cloud_sql_proxy/proxy.go
kurtisvg: I don't think this is a correct way to fix this. 

If we return here and `cfgs` has a valid value, then the rest of this function is ignored and the other flags that might have been set are ignored and not processed correctly( like `-dir` or `-fuse`).

hilts-vaughan: Thanks Kurtis. I completely forgot about all those other potential things.

Would it make more sense to aggregate the configurations and errors?

kurtisvg: I'm not sure.

Thinking about this more in-depth with the current code path, a `-skip-startup-validation` flag might not be the behavior we want to satisfy the problem. Originally when @broady and I had discussed this issue we had discussed just separating the parsing of the config with the validation that it was valid. However I think to satisfy the App Engine team we actually just want to parse the instances individually and continue running as long as one instance was parsed successfully.

So perhaps a better solution would be to let the flag be something like `-skip-failed-instance-parses`, and then check the flag inside `parseInstanceConfigs` level, after each individual `parseInstanceConfig`, and only return values for instances that are parsed successfully.

cmd/cloud_sql_proxy/proxy.go
hilts-vaughan: Thanks.

cmd/cloud_sql_proxy/proxy.go
hilts-vaughan: That would be the MVP I guess for our flex team unblocked. I think I can do that and move some things around. Let me try it out and see how it looks.

hilts-vaughan: When you say parse properly, do you mean something that does not return an error from `parseInstanceConfig`? I don't think checking the connection string would be enough -- since some customers might have "dead connection strings" that point to deleted instances.

kurtisvg: Yes, something that does not return an error from parseInstanceConfig. Maybe `-skip-failed-instance-config` would be better?

hilts-vaughan: Yeah, that sounds good. I just did the refactoring; I agree with you looking at this more now.

There's no need to burden the CLI utility with needing to know how to handle errors and conditionals. Makes for a messy interface.

I have something working; just refining tests and doing some manual testing.

hilts-vaughan:
Still working on it, just getting a draft up.

hilts-vaughan:
I can rebase and clean up the history after.  Does this look more in line with what you were thinking?

hilts-vaughan:
Sorry for the spam, it's been a while since I used the Github workflow model.

I think this should be good now.

README.md
```diff
@@ -34,6 +34,8 @@ cloud_sql_proxy takes a few arguments to configure what instances to connect to
   IP. Defaults to `PUBLIC,PRIVATE`
 * `-term_timeout=30s`: How long to wait for connections to close before shutting
   down the proxy. Defaults to 0.
+* `--skip_failed_instance_config`: Setting this flag will allow you to prevent the proxy from terminating when
```
kurtisvg: While I prefer unix style flags, currently it's consistent with the rest of the binary to use one. 
```suggestion
* `-skip_failed_instance_config`: Setting this flag will allow you to prevent the proxy from terminating when
```

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -305,15 +307,21 @@ func parseTCPOpts(ntwk, addrOpt string) (string, error) {
...
+func parseInstanceConfigs(dir string, instances []string, cl *http.Client, skipFailedInstanceConfigs bool) ([]instanceConfig, error) {
 	errs := new(bytes.Buffer)
 	var cfg []instanceConfig
 	for _, v := range instances {
 		if v == "" {
 			continue
 		}
 		if c, err := parseInstanceConfig(dir, v, cl); err != nil {
-			fmt.Fprintf(errs, "\n\t%v", err)
+			if skipFailedInstanceConfigs == true {
```
kurtisvg:
```suggestion
			if skipFailedInstanceConfigs {
```

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -305,15 +307,21 @@ func parseTCPOpts(ntwk, addrOpt string) (string, error) {
...
+				logging.Infof("There was a problem when parsing a instance configuration but ignoring due to the configuration. Error: %v", err)
+				// we intentionally don't add to `errs` since we don't want to bubble these
```
kurtisvg: nit: This comment can probably be left out. 

At the very least, it should be properly capitalized/punctuated - see [Comment Sentences](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences).

README.md
hilts-vaughan: You're right; done.

cmd/cloud_sql_proxy/proxy.go
hilts-vaughan: Done.

kurtisvg:
I poked @broady to take a look earlier this week and they mentioned two concerns and wanted time to look at in more in depth: 

1. Wasn't sure the flag name was clear(mentioned wanting the word "validation" in it)
2. Raised a concern with the best practice for accessing the value set by the flag

I can poke him again and get more clarification on these concerns. 

That said, we need to hold off on merging even once settled this as it's not even clear it's still required by the Flex team (as they've expressed the ability to fix the specific use case this flag was intended for without it).

PR #360, opened 2020-03-07
proxy: Added a connectivity test on startup to test for bad routing

### LINKED ISSUE #348: The proxy does not perform connectivity tests on startup to the actual instances
Some customers are confused by "Ready for new connections" and then try and connect and find it fails. This is often a routing problem but some of them incorrectly assume that the message above means everything was validated and fine.

Should we perform some kind of connectivity test when creating the instances or is this too much?

hilts-vaughan:
Fixes #348 

It's just a basic test for now. Maybe we should put this behind a flag. This will help customers fail fast instead of it being "ready for connections" but being unable to dial into the DB at all, which is obviously pretty important.


kurtisvg:
Definitely want this behavior flagged. Currently leaning towards disabled by default.

hilts-vaughan:
Sure, I'll get some flags behind it. 

Should we include the examples to have it?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -321,6 +321,20 @@ func stringList(s string) []string {
...
+	// The connectivity test right now is just a basic dial to make sure it is reachable
+	// and nothing more. This will help rule out basic network connectivity problems, though
+	// such as firewalls and the like. 
+	conn, err := client.Dial(instanceName)
+	
+	if err != nil {
+		return fmt.Errorf("Performed connectivity test to %v and got an error: %v", instanceName, err)
+	}
+
+	conn.Close()
```
kurtisvg: This should be deferred (and called after L328) so it always happens.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -321,6 +321,20 @@ func stringList(s string) []string {
...
 }
 
+func checkInstanceConnectivity(instanceName string, client *proxy.Client) (error) {
+	// The connectivity test right now is just a basic dial to make sure it is reachable
+	// and nothing more. This will help rule out basic network connectivity problems, though
+	// such as firewalls and the like. 
+	conn, err := client.Dial(instanceName)
+	
+	if err != nil {
+		return fmt.Errorf("Performed connectivity test to %v and got an error: %v", instanceName, err)
```
kurtisvg: Should say something like "Failed to connect successfully to using IP %s using instance %s: ". Ideally it should indicate if it's using public/private as well, since this would be helpful for debugging.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -536,6 +549,15 @@ func main() {
 		RefreshCfgThrottle: refreshCfgThrottle,
 	}
 
+	for _, instanceConfig := range cfgs {
```
kurtisvg: We should definitely flag this behavior by default (I think defaulting to disabled, at least for now).

cmd/cloud_sql_proxy/cloud_sql_proxy.go
kurtisvg: Additionally, it would be helpful to find out if there is a simple way to verify we have VPC access to rule out some troubleshooting (perhaps there is a metadata server we can query?)

kurtisvg:
> Sure, I'll get some flags behind it.
> 
> Should we include the examples to have it?

One thing I've been meaning to add is a section with troubleshooting suggestions. Perhaps this would be a good place to highlight it?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: Will this work if `conn` was null in the case of an error?

hilts-vaughan:
Sure, that sounds good. Had a couple more customer cases come through about this; wouldn't be bad to have something to help people troubleshoot this themselves.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
kurtisvg: Go doesn't have `null`, it has `nil`. The main difference is that `nil` is a "zero value" and safe to use for things like this.

hilts-vaughan: Got it; I need to read me some more Go docs.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: 1. Ack, it looks like getting the IP is a more complex since right now the proxy client abstracts all that way. I will look into what we can do there. 

2. Looks like it possible to get the private vs. public information from the Admin API. Is the right thing to add that information to the `instanceConfig` struct?

hilts-vaughan: re: getting the address; the Go errors already include this actually. Not sure if it's OK to depend on that implementation detail or not but since it's just diagnostic information maybe that's OK for the sake of simplicity.

hilts-vaughan:
https://cloud.google.com/sql/docs/mysql/sql-proxy#troubleshooting

I can add the flag to this page once we have decided on the details; I think that's the best bet?

kurtisvg:
@hilts-vaughan can you rebase this? I'm planning on a release soon and I'd like to include this.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -536,6 +554,17 @@ func main() {
 		RefreshCfgThrottle: refreshCfgThrottle,
 	}
 
+	if *performConnectivityTestsOnStartup == true {
```
broady, line 557: `if *performConnectivityTestsOnStartup {`

(no need for `== true`)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -536,6 +554,17 @@ func main() {
...
+		for _, instanceConfig := range cfgs {
+			err := checkInstanceConnectivity(instanceConfig.Instance, instanceConfig.IsPublic, proxyClient);
+			if err != nil {
+				log.Fatal(err)
+			}
```
broady, line 562: Thoughts:

- what about timeouts?
- perform the tests concurrently? (fail on first failure)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -536,6 +554,17 @@ func main() {
...
+		}	
+	}
+	
+	logging.Infof("Ready for new connections")
```
broady, line 566: Maybe accept connections while the connectivity test is happening?

And if there's a reason not to do that, explain that in the flag doc.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -536,6 +554,17 @@ func main() {
 		RefreshCfgThrottle: refreshCfgThrottle,
 	}
 
+	if *performConnectivityTestsOnStartup == true {
+		for _, instanceConfig := range cfgs {
+			err := checkInstanceConnectivity(instanceConfig.Instance, instanceConfig.IsPublic, proxyClient);
+			if err != nil {
+				log.Fatal(err)
```
broady, line 561: Use the `logging` package?

followed by
`os.Exit(...)`

hilts-vaughan:
Thanks for the review. I'll look at getting some of those changes integrated today. I'll rebase it too while I'm at it. 

Would you prefer one commit kvg?

kurtisvg:
> Thanks for the review. I'll look at getting some of those changes integrated today. I'll rebase it too while I'm at it.
> 
> Would you prefer one commit kvg?

What ever makes sense is fine. I usually do larger changes in separate commits but I also often group misc feedback into a single "Addresses feedback." commit.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: I think the biggest reason I had was that the messaging was confusing since "Ready for new connections" was throwing a lot of customers off, thinking things were good to go.

Without knowing the results of the test, it could be misleading. We could as a compromise allow the connections but not log anything out until we know for sure?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: 1. If it times out, it should fail I would imagine. That's what we want, right? 
2. You could; there's no reason you could not. I'm a bit new to Go; what's the best way to do this? Would I be doing something like this or is there something simpler? https://stackoverflow.com/a/24239322/1554844

It feels like something that is built into most other languages and is made very easy.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: Actually, that's what we're doing here looking back I think. We're accepting connections but we don't tell the user until we know for sure. I could revise the messaging though (or maybe even add a new log before and after to make it more clear.)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: We use `fatal` in the code all over. I count almost no errors and `os.Exit`, especially not used that way.

Could you let me know if you think that's the best way to proceed still?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: I implemented 2. Let me know if there is a better way, it wasn't all that bad in Go actually.

hilts-vaughan:
I had to remove the `defer` since it was trying to operate on the `nil` and was dereferencing it. It works fine under the new model.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -324,6 +325,27 @@ func stringList(s string) []string {
 	return spl
 }
 
+func checkInstanceConnectivity(instanceName string, public bool, client *proxy.Client) error {
+	// The connectivity test right now is just a basic dial to make sure it is reachable
+	// and nothing more. This will help rule out basic network connectivity problems, though
+	// such as firewalls and the like.
```
kurtisvg, line 331: nit: 
```suggestion
	// The connectivity test right now is just a basic dial to make sure it is reachable
	// and nothing more. This will help rule out basic network connectivity problems,
	// such as firewalls and the like.
```

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -324,6 +325,27 @@ func stringList(s string) []string {
...
+	if conn != nil {
+		conn.Close()
+	}
+
+	if err != nil {
+		if public {
+			return fmt.Errorf("Performed connectivity test to public instance %v and got an error: %v. Do you have a route to it?", instanceName, err)
+		} else {
+			return fmt.Errorf("Performed connectivity test to private instance %v and got an error: %v. Do you have a route to it?", instanceName, err)
+		}
```
kurtisvg, line 343:
```suggestion
		ipType := "public"
		if !public {
			ipType = "private"
		} 
		return fmt.Errorf("Unable to connect to instance %v via %s IP. Verify you have a valid connection path to the instance's IP. (err: %v). ", instanceName, ipType, err)
		}
```

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -324,6 +325,27 @@ func stringList(s string) []string {
 	return spl
 }
 
+func checkInstanceConnectivity(instanceName string, public bool, client *proxy.Client) error {
+	// The connectivity test right now is just a basic dial to make sure it is reachable
+	// and nothing more. This will help rule out basic network connectivity problems, though
+	// such as firewalls and the like.
+	conn, err := client.Dial(instanceName)
+
```
kurtisvg, line 333: nit: nix newline

cmd/cloud_sql_proxy/cloud_sql_proxy.go
kurtisvg: This seems right to me - it makes sense to wait to say "Ready for new connections" until we've definitely reached a stable state.

cmd/cloud_sql_proxy/proxy.go
```diff
@@ -273,6 +274,15 @@ func parseInstanceConfig(dir, instance string, cl *http.Client) (instanceConfig,
 	if err != nil {
 		return instanceConfig{}, err
 	}
+
+	// Check to see if one of the addresses is public or not
+	// and mark that on the config
+	for _, mapping := range inst.IpAddresses {
+		if mapping.Type == "PRIMARY" {
+			ret.IsPublic = true
```
kurtisvg, line 282: What if they've specified private as a preferred IP type? Will this still return public?

README.md
```diff
@@ -34,6 +34,8 @@ cloud_sql_proxy takes a few arguments to configure what instances to connect to
   IP. Defaults to `PUBLIC,PRIVATE`
 * `-term_timeout=30s`: How long to wait for connections to close before shutting
   down the proxy. Defaults to 0.
+* `-perform_connectivity_tests`: Performs connectivity tests on startup to the
+  databases and exits the proxy if one of them fails.
```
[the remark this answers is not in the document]
hilts-vaughan: Sure, I'll defer to you both on the flag name. I I think either is good.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
hilts-vaughan: Done.

cmd/cloud_sql_proxy/proxy.go
hilts-vaughan: Not sure I follow. For private instances, this should be false since that's the default value in `Go` from what I can tell.

kurtisvg: A Cloud SQL instance can have both a public and private ip. By default, the proxy uses a PUBLIC ip if it exists, otherwise a PRIVATE one. Users can override this behavior by specifying a different preference using [`-ip_address_types`]. If I'm reading this correctly, the current logic doesn't take this into account, meaning it will mark an instance as public even if the private ip is used to dial.

hilts-vaughan: I get what you mean. I'm going to add another flag and try and reconcile it with a nicer error message. 

Not sure what else; we could try and provide both? But the user may not care about checking both perse.

hilts-vaughan: Done. Two queries:

1. Is this OK or should we do something else with this?
2. Is there an easy way for me to write some unit tests for this?

kurtisvg: I don't think this approach is probably the right way to approach this either. I think probably some better approaches would be one of the following:

1. Reimplement similar logic to [`findIpAddr`](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/3acc260458d1198c32060d20eadfa36cce723a23/proxy/certs/certs.go#L208) here in cmd, except returns the IP type used rather than the addr  (`findIpType`?). 

2. Modify `findIpAddr` to return addr and type, and then move the "test connectivity" function into the client so that it has access to it.

Option 2 is probably better since we don't have to reimplement logic and don't run the risk that one changes but not the other, but might be somewhat invasive. I'm not sure without taking a closer look.

hilts-vaughan: There is no obvious reason that `findIpAddr` has to be in `certs` either from what I can tell since it is free most of dependencies inside of that module. You could probably split it out into something else.

That being said, at what cost? Does the user really need to know public / private? They're going to get the IP in the actual exception anyway so they could go check. Do you think there is enough value add here? 

The hairy part is probably the "cert source" structures and splitting those out.

Thoughts?

kurtisvg: At some point, we definitely want to do this. One of the bigger issues with the proxy is that folks don't understand it's not a VPN, and doesn't grant a connection to an instance via Private IP if it didn't have it before. As some point I'd like to be able to provide troubleshooting steps specific to the IP type to make the error actionable (Using private IP and can't connect - Is the firewall egress rules allow for port 3307? Are you on a VPC?) and maybe even add some diagnostic info (like checking for a VPC metadata server to confirm connectivity).

That said, having this info included as part of this PR is "nice to have" but not required - we can always expand upon it later.

kurtisvg:
Closing this PR since it's gone stale.

PR #376, opened 2020-04-15
fix: Prevent the binary from getting checked in when using go build

hilts-vaughan:
## Change Description

I've almost checked the binary a few times in; is there a reason we need to have this in the repo? 

## Checklist

  Small change; can open an issue if there's enough of a reason to

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,7 @@
 # Jetbrains IDE
 .idea/
 
+# Binary output
+cmd/cloud_sql_proxy/cloud_sql_proxy
+
 tests/cloud_sql_proxy
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,5 @@
 # Jetbrains IDE
 .idea/
 
-# Binary output
-cmd/cloud_sql_proxy/cloud_sql_proxy
-
-tests/cloud_sql_proxy
+# Compiled binary
+*/cloud_sql_proxy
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,4 +2,7 @@
 # Jetbrains IDE
 .idea/
 
-tests/cloud_sql_proxy
+# Binary output
+cmd/cloud_sql_proxy/cloud_sql_proxy
+
+tests/cloud_sql_proxy
```
kurtisvg, line 8: nit: Probably best to leave a newline at the end

kurtisvg:
@hilts-vaughan I generalized a bit. Does that look okay to you?

hilts-vaughan:
I generally would prefer to be more explicit but I guess given you can build from any directory, that's probably fine. :)

Your project! It meets my use case.

kurtisvg:
> I generally would prefer to be more explicit but I guess given you can build from any directory, that's probably fine. :)
> 
> Your project! It meets my use case.

So I would have just done `cmd/cloud_sql_proxy/cloud_sql_proxy`. Could you expand why you wanted to ignore `tests` too? Do you usually copy the binary in there?

hilts-vaughan:
I'm confused. It's ignored in `tests` in `master` right now. 

You have proposed to change to `*/cloud_sql_proxy`

I just added a line.

kurtisvg:
Your right, I looked at the diff wrong

PR #378, opened 2020-04-15
test: Refactor tests to run without GCE instance. 

kurtisvg:


tests/connection_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"database/sql"
+	"fmt"
+	"testing"
+)
+
+// proxyConnTest is a test helper to verify the proxy works with a basic connectivity test.
+func proxyConnTest(t *testing.T, connName, driver, dsn string, port int, dir string) {
+	ctx := context.Background()
+
+	var args []string
```
shubha-rajan, line 29: Would we eventually need to add a `-credential_file` arg here to pass in service account credentials?

kurtisvg: At some point we probably want to add tests for different auth flow, but for now I'm just worried about maintaining parity with the existing tests.

Issue #388, opened 2020-05-01
Feature Request: Exit 0 on graceful shutdown from SIGTERM


karlkfi:
## Feature Description
cloud_sql_proxy currently exits with code 2 after receiving a SIGTERM. I would like it to exit 0 to indicate successful shutdown. This would allow for using SIGTERM to shutdown the proxy in a Kubernetes job after the primary container exits 0 without the job being marked as a failure.

## Alternatives Considered
Another option would be to allow configuring the successful shutdown exit code. This may be more reverse compatible, if required. But the current behavior could also be considered a bug.

Carrotman42:
The current implementation only exits with a nonzero status if there is an error during shutdown; unfortunately, currently that error is not printed out. See https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/389 for a fix for that specific problem.

By my read of the code (https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/3acc260458d1198c32060d20eadfa36cce723a23/proxy/proxy/client.go#L358-L372), the only reason an error would occur is if there are active connections to the Proxy after waiting an amount of time specified by `--term_time`. Note that flag's default value is 0, so if there are any connections at the moment of SIGTERM the Proxy would exit with a nonzero status.

If you expect connections to the Proxy to be torn down "around" the time of the SIGTERM, I suggest setting `--term_time` to some value so that there's time for the Proxy to notice the connections have been closed.

karlkfi:
hmm. The only thing using the proxy was the app (dbmate) that had exited and the liveness/readiness probes (netcat). I had term_time set to 10s, but i can increase it. I'll try to see if I can figure out what's holding open the connections... But thanks for the PR. It'd be nice to know what the failure reason is.

karlkfi:
So it turns out my netcat liveness/readiness probes were leaking open connections.
Changing them to  use `-d` (ignore stdin and FIN immediately) and `-w 5` (timeout before before being killed) solved the issue. cloudsql-proxy now exits 0 on sigterm.

PR #389, opened 2020-05-01
Print out any errors during SIGTERM-caused shutdown

Carrotman42:
Before this change any error encountered during a forced shutdown via proxyClient.Shutdown will not be printed out, yet when there is an error the Proxy will still exit with a nonzero status. In addition to exiting with a nonzero status, this CL causes the errors to actually be printed out.

Note that log.Fatal(f) exits the program with exit code 1, which is a change from the prior implementation which exited with code 2. Let me know if exit code 2 was particularly important; I will happily log a message and explicitly exit(2).

I noticed this upon reading issue https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/388.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -547,11 +547,10 @@ func main() {
 		logging.Infof("Received TERM signal. Waiting up to %s before terminating.", *termTimeout)
 
 		err := proxyClient.Shutdown(*termTimeout)
-
 		if err == nil {
 			os.Exit(0)
 		}
-		os.Exit(2)
+		log.Fatalf("Error during SIGTERM shutdown: %v", err)
 	}()
 
 	proxyClient.Run(connSrc)
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -550,7 +550,8 @@ func main() {
 		if err == nil {
 			os.Exit(0)
 		}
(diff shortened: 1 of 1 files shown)


kurtisvg:
Thanks for taking the time to PR these fixes. 

In the original PR, it doesn't look like 2 was picked for any specific reason. However, since it _was_ picked intentionally and mentioned in the PR, I'm hesitant to change it unless we have a good reason.

Carrotman42:
+1, that is a very good point.

karlkfi:
IME, 2 usually means user error, while 1 is general catchall.
This comes from a loose interpretation of [shell exit codes with special meaning](https://www.tldp.org/LDP/abs/html/exitcodes.html).

It probably originates from the [Bash manual](https://linux.die.net/man/1/bash), if not earlier.

> The return value is 0 if the string matches the pattern, and 1 otherwise. If the regular expression is syntactically incorrect, the conditional expression's return value is 2.

> Shell builtin commands return a status of 0 (true) if successful, and non-zero (false) if an error occurs while they execute. All builtins return an exit status of 2 to indicate incorrect usage.

An open connection could loosely be considered user error, tho I probably would have specified a different exit code for the case where the cause is the client operator's and not the server operator's fault.

Carrotman42:
I agree that sometimes exit codes are documented to mean things. But IMHO if a specific exit code is not explicitly documented, it should not be relied on by users. Without more docs, exit codes should just be treated as Boolean, i.e. zero for success and nonzero for error.

Now that doesn't mean people won't rely on them without documentation :) which I why I agreed it'd be better to not change things given that it was easy to not change.

PR #391, opened 2020-05-04
Optimize term-timeout wait.

kurtisvg:
Add a small optimization to better cap the max amount of time the proxy spends waiting for incoming connections to close. 

/cc @Carrotman42

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -359,9 +359,16 @@ func NewConnSrc(instance string, l net.Listener) <-chan Conn {
 // close. Returns an error if there are still active connections after waiting
 // for the whole length of the timeout.
 func (c *Client) Shutdown(termTimeout time.Duration) error {
-	termTime := time.Now().Add(termTimeout)
-	for termTime.After(time.Now()) && atomic.LoadUint64(&c.ConnectionsCounter) > 0 {
-		time.Sleep(time.Second)
+	term, tick := time.After(termTimeout), time.Tick(100 * time.Millisecond)
+	for {
+		select {
+			case <- tick:
+				if atomic.LoadUint64(&c.ConnectionsCounter) > 0 {
+					continue
+				}
+			case <- term:
+		}
+		break
 	}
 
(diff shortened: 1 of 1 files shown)


proxy/proxy/client.go
```diff
@@ -359,9 +359,16 @@ func NewConnSrc(instance string, l net.Listener) <-chan Conn {
 // close. Returns an error if there are still active connections after waiting
 // for the whole length of the timeout.
 func (c *Client) Shutdown(termTimeout time.Duration) error {
-	termTime := time.Now().Add(termTimeout)
-	for termTime.After(time.Now()) && atomic.LoadUint64(&c.ConnectionsCounter) > 0 {
-		time.Sleep(time.Second)
+	term, tick := time.After(termTimeout), time.Tick(100 * time.Millisecond)
+	for {
+		select {
+			case <- tick:
```
Carrotman42, line 365: Please run gofmt.

proxy/proxy/client.go
```diff
@@ -359,9 +359,16 @@ func NewConnSrc(instance string, l net.Listener) <-chan Conn {
 // close. Returns an error if there are still active connections after waiting
 // for the whole length of the timeout.
 func (c *Client) Shutdown(termTimeout time.Duration) error {
-	termTime := time.Now().Add(termTimeout)
-	for termTime.After(time.Now()) && atomic.LoadUint64(&c.ConnectionsCounter) > 0 {
-		time.Sleep(time.Second)
+	term, tick := time.After(termTimeout), time.Tick(100 * time.Millisecond)
```
Carrotman42, line 362: nit: `defer tick.Stop()`. Even though this is (currently) on only the shutdown path, it seems prudent to not leak resources anyway.

kurtisvg: `time.Tick` returns a channel and can't be stopped, but I can use `NewTicker` instead.

Carrotman42: Whoops duh, thanks for figuring out what I meant.

PR #394, opened 2020-05-07
Add kubernetes examples.

### LINKED ISSUE #216: Kubernetes.md seems to be out of date
The instructions in Kubernetes.md seem to refer to a configuration that's no longer supported. Does the proxy still support listening on e.g. 0.0.0.0?

kurtisvg:
Add some examples of using the proxy from kuberenetes, along with region tags for referencing from the cloud docs.

/cc @tlichtenberg, @GabeWeiss - would appreciate feedback

Fixes #216


examples/kubernetes/README.md
```diff
@@ -0,0 +1,233 @@
...
+      <YOUR-GSA-NAME>@<YOUR-GCP-PROJECT>.iam.gserviceaccount.com
+    ```
+1. Add an annotation to `<YOUR-KSA-NAME>` to complete the binding:
+    ```sh
+    kubectl annotate serviceaccount \
+       <YOUR-KSA-NAME> \
+       iam.gke.io/gcp-service-account=<YOUR-GSA-NAME>@<YOUR-GCP-PROJECT>.iam.gserviceaccount.com
+    ```
+1. Finally, make sure so specify the service account for the k8s object:
+    > [proxy_with_workload_identity.yaml](proxy_with_workload_identity.yaml#L2-L7)
```
dmahugh, line 103: to

examples/kubernetes/README.md
```diff
@@ -0,0 +1,233 @@
...
+service account to represent your application. It is recommended that you create
+a service account unique to each application, instead of using the same service
+account everywhere. This model is more secure since it allows your to limit
+permissions on a per-application basis.
+
+The service account for your application needs to meet the following criteria:
+
+1. Belong to a project with the [Cloud SQL Admin API][admin-api] enabled.
+1. [Has been granted][grant-sa] the 
+   [`Cloud SQL Client` IAM role (or equivalent)][csql-roles]
```
dmahugh, line 57: Minor nit: consistent use of periods at the end of list items. Some have them and some don't here.

examples/kubernetes/README.md
kurtisvg: Done.

examples/kubernetes/README.md
kurtisvg: Done.

examples/kubernetes/README.md
```diff
@@ -0,0 +1,235 @@
+# Using the Cloud SQL proxy on Kubernetes
+
+The Cloud SQL proxy is the recommended way to connect to Cloud SQL, even when
+using private IP. This is because the proxy provides strong encryption and 
+authentication using IAM, which can help keep your database secure.
```
GabeWeiss, line 5: -can help
+helps

examples/kubernetes/README.md
```diff
@@ -0,0 +1,235 @@
...
+
+The Cloud SQL proxy is the recommended way to connect to Cloud SQL, even when
+using private IP. This is because the proxy provides strong encryption and 
+authentication using IAM, which can help keep your database secure. 
+
+## Configure your application with Secrets
+
+In Kubernetes, [Secrets][ksa-secret] are a secure way to pass configuration 
+details to your application. You can create a Secret with details such as your
+database name, user, and password which can be injected into your application as
```
GabeWeiss, line 11: Clarification... is a Secret considered a bundle of key/val pairs? Or is each key/val pair a "Secret"? I haven't looked at the Secrets documentation to know how it's broken down. Am I creating a single "Secret" which contains db name, db user and db password? Or is each one of those called a "Secret" that is then bundled together somehow?

examples/kubernetes/README.md
```diff
@@ -0,0 +1,235 @@
...
+env vars.
+
+1. Create a secret with information needed to access your database:
+    ```shell
+    kubectl create secret generic <YOUR-DB-SECRET> \
+        --from-literal=db_user=<YOUR-DATABASE-USER> \
+        --from-literal=db_password=<YOUR-DATABASE-PASSWORD> \
+        --from-literal=db_name=<YOUR-DATABASE-NAME>
+    ```
```
GabeWeiss, line 20: From this block above, looks like a Secret is considered a bundle of key/val pairs together...that's a little muddy, but not sure about specifically how I'd clarify other than maybe just calling it out explicitly above like "A secret is a bundle of key/value pair configurations that you can inject into your application as env vars"

examples/kubernetes/README.md
```diff
@@ -0,0 +1,235 @@
...
+## Providing the service account to the proxy
+
+Next, you need to configure Kubernetes to provide the service account to the 
+Cloud SQL Proxy. There are two recommended ways to do this. 
+
+### Workload Identity
+
+If you are using [Google Kubernetes Engine][gke],  the preferred method is to
+use GKE's [Workload Identity][workload-id] feature. This method allows you to
+bind a [Kubernetes Service Account (KSA)][ksa] to a Google Service Account
```
GabeWeiss, line 78: add "(GSA)" after Google Service Account same as you did for the KSA for consistency. Even though it happens right after, it's still a cognitive blip without the parenthetical reference

kurtisvg: It's there, just on the next line (cause markdown):

> If you are using [Google Kubernetes Engine][gke],  the preferred method is to
> use GKE's [Workload Identity][workload-id] feature. This method allows you to
> bind a [Kubernetes Service Account (KSA)][ksa] to a Google Service Account
> (GSA). The GSA will then be accessible to applications using the matching KSA.

examples/kubernetes/README.md
kurtisvg: Added a line clarifying a single Secret is composed of multiple key/values.

PR #415, opened 2020-07-10
feat(docker): Add "-alpine" and "-buster" based images. 

### LINKED ISSUE #330: Release official alpine-based container image for cloud-sql-proxy
Hi guys, 

will be very cool, if you could add alpine version of cloudsql-proxy image into gcr.io/cloudsql-docker/gce-proxy registry. Because for now, if I need cloud_sql_proxy binary for alpine, I should build it by myself.

### LINKED ISSUE #371: Release official debian-based container image for cloud-sql-proxy
We should offer a version of the cloud-sql-proxy container image that's based on debian (buster). See discussion in #317 why this might be useful to users.

kurtisvg:
## Change Description

Add Dockerfiles for `alpine` and `buster` based images.

Note: I'll need to manually push the images once this is merged.

## Checklist

## Relevant issues:

- Fixes #330
- Fixes #371


thvd:
@kurtisvg is there any disadvantage by using the alpine version? Of course it's going to be smaller in size, so that's of course better. So is there still a use-case for using the old one?

kurtisvg:
@thvd The default (distroless based) containers are more secure. I would only recommend using the the other versions if you absolutely need a shell for some reason. 

I don't consider size to be a significant factor here. While smaller is certainly better, there is only a 5mb (10mb vs 15mb)  difference, which isn't really noticeable. It's almost certainly an order of magnitude smaller than whatever your main application container is, and in most circumstances the image will be cached after the first pull.

PR #419, opened 2020-07-16
chore(docs): Remove "v" from container image tags

### LINKED ISSUE #418: Wrong Docker Image Tags
## Description
The docs say the following Docker image tags are available:

> - v$VERSION - default image (recommended)
> - v$VERSION-alpine - uses alpine:3 as a base image
> - v$VERSION-buster - uses debain:buster as a base image

If you check us.gcr.io/cloudsql-docker/gce-proxy, you're actually tagging images like `$VERSION`, `v$VERSION-alpine`, and `v$VERSION-buster`. Also, `v$VERSION-alpine` and `v$VERSION-buster` only exist for `v1.17`.

## Solution
Update the docs to match the actual docker image tags that are and will be available.

shubha-rajan:
## Change Description

Updated documentation for proxy container images in README.md

## Relevant issues:

- Fixes #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,9 +105,9 @@ __Note:__
 
 Each image is tagged with the version of the proxy it was released with. The 
 following tags are currently supported:
-  * `v$VERSION` - default image (recommended)
-  * `v$VERSION-alpine` - uses [`alpine:3`](https://hub.docker.com/_/alpine) as a base image
-  * `v$VERSION-buster` - uses [`debian:buster`](https://hub.docker.com/_/debian) as a base image
+  * `$VERSION` - default image (recommended)
+  * `$VERSION-alpine` - uses [`alpine:3`](https://hub.docker.com/_/alpine) as a base image (only supported from v1.17 up)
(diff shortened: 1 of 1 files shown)


kurtisvg:
So I'd actually prefer if we stuck to using `v$VERSION` so it's consistent with the release tag we have to use on GitHub. If we take this course of action, there are a few choices to make:

1. Do we backport `v$VERSION` tag to previous versions? (probably)
2. Do we continue to push both versions of the tag in the future? If so, we'll need to update the default Cloud Build configuration.

kurtisvg:
> So I'd actually prefer if we stuck to using `v$VERSION` so it's consistent with the release tag we use on GitHub. If we go this route, there are a few choices to make:
> 
> 1. Do we backport `v$VERSION` tag to previous versions? (probably)
> 2. Do we continue to push both versions of the tag in the future? If so, we'll need to update the default Cloud Build configuration.

shubha-rajan:
I think backporting makes sense. As for pushing both versions in the future, would it break users if we only pushed the `v$VERSION` and do we want to officially support both tags?

kurtisvg:
Do you mind fixing the BUILD files in this PR as well?

PR #448, opened 2020-09-01
chore: add release-please action

shubha-rajan:
## Change Description

Adding a workflow to use the [release-please Github action](https://github.com/GoogleCloudPlatform/release-please-action) to automate release PR and CHANGELOG generation, as well as update the version in a text file.


version.txt
```diff
@@ -0,0 +1 @@
+1.18.0-dev
```
kurtisvg, line 1: Let's try it without the version.txt first and see what happens

shubha-rajan: Sounds good.

PR #471, opened 2020-09-18
chore: Add a workflow to update the versionString on release PRs

shubha-rajan:
This GitHub Actions workflow checks out the source branch of any PR generated by `release-please[bot]` and pushes another commit to update the version string in the proxy. It *should* push the commit as release-please bot and get past CLA issues that way

### DIFF
--- a/.github/workflows/update-versions.yml
+++ b/.github/workflows/update-versions.yml
@@ -1,0 +1,37 @@
+
+name: update-version
+on:
+  pull_request:
+          types: [opened]
+          branches: [master]
+
+jobs:
+  update-version-in-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with: 
+          ref: ${{ github.head_ref }}
+
+      - if: github.actor == "release-please[bot]"
+        name: find-and-replace-version
+        run: |
+          VERSION=$(cat version.txt)
+          sed -i -r 's/[0-9]+.[0-9]+.[0-9]+(-dev)?/'"$VERSION"'/' cmd/cloud_sql_proxy/cloud_sql_proxy.go
+          
+      - if: github.actor == "release-please[bot]"
+        name: setup-git-config
+        run: |
+          git config user.name "release-please[bot]"
+          git config user.email "55107282+release-please[bot]@users.noreply.github.com"
+          
+      - if: github.actor == "release-please[bot]"
+        name: push-commit
+        run: |
+          git add cmd/cloud_sql_proxy/cloud_sql_proxy.go
+          git commit -m "Update version"
+          git push origin ${{ github.head_ref }}
+      
+
+        
--- a/.github/workflows/update-versions.yml
+++ b/.github/workflows/update-versions.yml
@@ -2,8 +2,9 @@
 name: update-version
 on:
   pull_request:
-          types: [opened]
           branches: [master]
+          paths: 
+            - 'version.txt'
 
 jobs:
   update-version-in-pr:
--- a/.github/workflows/update-versions.yml
+++ b/.github/workflows/update-versions.yml
@@ -20,16 +20,8 @@ jobs:
         run: |
           VERSION=$(cat version.txt)
           sed -i -r 's/(var versionString = ")[0-9]+.[0-9]+.[0-9]+(-dev)?/\1'"$VERSION"'/' cmd/cloud_sql_proxy/cloud_sql_proxy.go
-          
-      - if: github.actor == "release-please[bot]"
-        name: setup-git-config
-        run: |
           git config user.name "release-please[bot]"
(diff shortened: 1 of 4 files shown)


.github/workflows/update-versions.yml
```diff
@@ -0,0 +1,37 @@
+
+name: update-version
+on:
+  pull_request:
+          types: [opened]
+          branches: [master]
```
kurtisvg, line 6: Hmm so this means it'll run just on open? What happens when release please has to rerun? 

I think we should restrict this based on `[paths](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)` as well - we only want this to run if version.txt has changed.

.github/workflows/update-versions.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with: 
+          ref: ${{ github.head_ref }}
+
+      - if: github.actor == "release-please[bot]"
+        name: find-and-replace-version
+        run: |
+          VERSION=$(cat version.txt)
+          sed -i -r 's/[0-9]+.[0-9]+.[0-9]+(-dev)?/'"$VERSION"'/' cmd/cloud_sql_proxy/cloud_sql_proxy.go
```
kurtisvg, line 21: We probably want a slightly more strict regex for the entire line (`var versionString`) instead to make sure this doesn't change any where else we might want to have version numbers in the future.

.github/workflows/update-versions.yml
```diff
@@ -0,0 +1,37 @@
...
+          
+      - if: github.actor == "release-please[bot]"
```
kurtisvg, line 23: Is there any advantage to these being different steps?

kurtisvg (changes requested):

.github/workflows/update-versions.yml
shubha-rajan: it's mostly for logging purposes. if something goes wrong, we know exactly which step is the problem. We can also do it in one step if you prefer that.

.github/workflows/update-versions.yml
shubha-rajan: I agree on the path restriction. I forgot that release please closes and reopens the PR every time it edits it.

.github/workflows/update-versions.yml
shubha-rajan: done

.github/workflows/update-versions.yml
kurtisvg: I have a slight preference towards a single step for readability (and mostly because of the multiple if checks), but I haven't seen how the logging is handled. If you feel the separate steps is significant for debugging, please feel free to leave.

.github/workflows/update-versions.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  update-version-in-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with: 
+          ref: ${{ github.head_ref }}
+
+      - if: github.actor == "release-please[bot]"
```
kurtisvg, line 18: Are we sure we only want to bump this version if the actor is release-please? What are the consequences is someone else edits the file?

kurtisvg (changes requested):

.github/workflows/update-versions.yml
shubha-rajan: Doing it in one step should be fine. I can update this.

.github/workflows/update-versions.yml
shubha-rajan: I guess my thought was that nobody should be editing the version file unless they have a really good reason to (in which case they should know to edit the version in both places, and would probably do a global find/replace anyway). I'd personally prefer bots to only push commits on bot PRs.

I'm thinking we could even add a workflow that generates a PR to append the `-dev` to the versions in both places after a new release tag is pushed, so that way we'd never have to manually update versions.

kurtisvg: That workflow would use GitHub actions, which would use a different user than this does. Does that affect this?

shubha-rajan: nope. because we would be able to update in both places at the same time with the other workflow, so we wouldn't need this to push a commit on to the PR. My thought was that this was more of a configuration that would make the release PRs update in both places since release-please only updates `version.txt` by default.

kurtisvg: SGTM

kurtisvg (approved):
LGTM - is there a good way to test this before we merge?

shubha-rajan:
I can't really think of a good way to test it. Especially since the action needs a PR from a specific github actor to do anything.

kurtisvg:
Ok - I'm okay with merging for now, let's just make sure we keep an eye on the next time release-please opens a new PR.

PR #504, opened 2020-10-14
feat: use regionalized instance ids to prevent global conflicts with sqladmin v1

shubha-rajan:
Update the SQLAdmin API calls to use an instance name with a region prefix with the format <region>~<instance-name> to prevent issues with conflicting instance names when sqladmin v1 is released and instance names become regionally unique instead of globally unique

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -345,7 +345,7 @@ func listInstances(ctx context.Context, cl *http.Client, projects []string) ([]s
 				for _, in := range r.Items {
 					// The Proxy is only support on Second Gen
 					if in.BackendType == "SECOND_GEN" {
-						ch <- fmt.Sprintf("%s:%s:%s", in.Project, in.Region, in.Name)
+						ch <- in.ConnectionName
 					}
 				}
 				return nil
--- a/proxy/certs/certs.go
+++ b/proxy/certs/certs.go
@@ -169,8 +169,9 @@ func (s *RemoteCertSource) Local(instance string) (ret tls.Certificate, err erro
 		return ret, err
 	}
 
-	p, _, n := util.SplitName(instance)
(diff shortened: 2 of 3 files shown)


proxy/certs/certs.go
```diff
@@ -169,8 +169,9 @@ func (s *RemoteCertSource) Local(instance string) (ret tls.Certificate, err erro
 		return ret, err
 	}
 
-	p, _, n := util.SplitName(instance)
-	req := s.serv.SslCerts.CreateEphemeral(p, n,
+	p, region, n := util.SplitName(instance)
```
kurtisvg, line 172: nit: probably `r` to be consistent (the variable isn't going far)

kurtisvg (changes requested):
I don't know if the response name will be changed if theres a conflict - if so, this line may need to be updated: 
https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/15fe7fe379608380c9cef3b4d04a2254d8dfb2bb/cmd/cloud_sql_proxy/cloud_sql_proxy.go#L348

We'll need to reach out to the API team to verify (and probably make sure to test this when there is a known conflict to to be sure).

proxy/certs/certs.go
shubha-rajan: fixed.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -345,7 +345,7 @@ func listInstances(ctx context.Context, cl *http.Client, projects []string) ([]s
 				for _, in := range r.Items {
 					// The Proxy is only support on Second Gen
 					if in.BackendType == "SECOND_GEN" {
-						ch <- fmt.Sprintf("%s:%s:%s", in.Project, in.Region, in.Name)
+						ch <- fmt.Sprintf("%s", in.ConnectionName)
```
kurtisvg, line 348: No need to use `fmt` here - we can just send the string as is?

kurtisvg (approved):

PR #550, opened 2020-12-01
fix: Ensure neccesary fields are 64-bit aligned

### LINKED ISSUE #543: bug: v1.19.0 x86-32 binaries segmentation violate on first connection
## Bug Description

I got the proxy with

`wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.386 -O cloud_sql_proxy && chmod +x cloud_sql_proxy`

for months now and after your release yesterday I got the following error when I am using it inside my Python Scripts.

I run it inside gitlab:

```
Transform Prod SQL:
  image: python:3.7.2-alpine
  stage: Transform SQL
  variables:
    MYSQL_PWD: $SQL_PASS_PROD
  script:
    - wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.386 -O cloud_sql_proxy && chmod +x cloud_sql_proxy
    - ./cloud_sql_proxy -instances=xxxx=tcp:0.0.0.0:3306 -credential_file=client.json &
```

Error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x804a0ac]
goroutine 102 [running]:
runtime/internal/atomic.Xadd64(0xa4780b4, 0x1, 0x0, 0x0, 0x0)
	/usr/local/go/src/runtime/internal/atomic/asm_386.s:105 +0xc
github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy.(*Client).handleConn(0xa478060, 0xffe56b9a, 0x34, 0x868f080, 0xa8ba090)
	/workspace/proxy/proxy/client.go:129 +0x50
created by github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy.(*Client).Run
	/workspace/proxy/proxy/client.go:120 +0x99
ERROR 2013 (HY000): Lost connection to MySQL server at 'handshake: reading initial communication packet', system error: 11

```

kurtisvg:
## Change Description

Moves the `Client.ConnectionsCounter`  field to the start of the struct since these values are guaranteed to be 64-bit aligned. Also adds a regression test to ensure that it doesn't get moved out of alignment in the future. 

## Checklist

## Relevant issues:

- Fixes #543

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -60,6 +60,13 @@ type CertSource interface {
 // Client is a type to handle connecting to a Server. All fields are required
 // unless otherwise specified.
 type Client struct {
+	// ConnectionsCounter is used to enforce the optional maxConnections limit
+	ConnectionsCounter uint64
+
+	// MaxConnections is the maximum number of connections to establish
+	// before refusing new connections. 0 means no limit.
+	MaxConnections uint64
+
 	// Port designates which remote port should be used when connecting to
 	// instances. This value is defined by the server-side code, but for now it
 	// should always be 3307.
@@ -95,13 +102,6 @@ type Client struct {
 	// RefreshCertBuffer is the amount of time before the configuration expires to
 	// attempt to refresh it. If not set, it defaults to 5 minutes.
 	RefreshCfgBuffer time.Duration
-
-	// MaxConnections is the maximum number of connections to establish
-	// before refusing new connections. 0 means no limit.
-	MaxConnections uint64
-
-	// ConnectionsCounter is used to enforce the optional maxConnections limit
-	ConnectionsCounter uint64
 }
 
 type cacheEntry struct {
--- a/proxy/proxy/client_test.go
+++ b/proxy/proxy/client_test.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+	"unsafe"
 )
 
 const instance = "instance-name"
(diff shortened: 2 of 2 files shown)


kurtisvg:
@Carrotman42 PTAL if you have time

shubha-rajan (approved):

Carrotman42:
I have a small criticism that usually one should put the more interesting (and more required) fields at the top of some struct, and have optional/uninteresting fields more toward the bottom. So putting this field up at the top of the struct seems wrong to me, especially since it seems like this field should never be set by users directly anyway. (see later when I suggest that we unexport it.)

If we are going to have a unit test anyway to verify that it's aligned appropriately, to me it seems better to leave the field be and just add padding until the test passes. I admit that's not ideal from a maintenance perspective, but if you're only ever tacking fields on to the end of the struct it won't require additional changes. Plus it's very easy to fix, just tedious.

Another option that I just thought of: the restriction applies only to 64 bit integers, so we can just change the type of this field to uint32. There's no way this proxy (or the Cloud SQL backend, or any database for that matter) is going to support more than 2^32 connections at the same time in any near future, that sounds just too high to be realistic especially given that each connection requires at least 4-6K of memory to maintain on this client Proxy side.

(and along with a type change, I think we should unexport that field by renaming it to connectionsCounter: it seems like a mistake to have exported it because we don't want outside callers to modify this field at all. If someone needs to read the current value, we can expose a getter for it.)

wydt?

kurtisvg:
@Carrotman42 So I actually did consider switching it to uint32 instead for the same reasons you mentioned, but ended up dismissing that route because I wanted to avoid a breaking change. FWIW, I agree with you that it probably shouldn't have been exported, but at this point it's been like that for 3 years so unless I'm missing something it's pretty much stuck as is.  

RE: Padding vs top of struct, padding the struct feels like a pretty hacky/unintuitive approach. If the only concern is signaling importance, it seems better to leave it at the top of the struct.

Carrotman42:
I still disagree with the field placement but don't feel strongly enough :)

Lgtm, thanks for handling fixing this!

PR #609, opened 2021-02-12
fix: improve logging for file descriptor limits

### LINKED ISSUE #413: Improve error message when hard rlimit for file descriptors is too low
When I run cloud_sql_proxy from my Ubuntu 18.04 VM, it works, but I get this confusing error message every time:

    failed to setup file descriptor limits: failed to set rlimit {&{8500 4096}} for max file descriptors: invalid argument

The root cause of this is that my hard limit for file descriptors available to the process are (for whatever reason) 4096.  In this case, I believe the proxy should try to give the user a more informational error message, rather than trying something that is all but certain to fail, and printing out the inevitable "invalid argument" error.

This is the problem area:
https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/5f761d7ef539bfe4fb65c6856d439496cddbfcc7/proxy/limits/limits.go#L62-L64

I think it's fine to *try* setting the limit in both ways here, but rather than simply outputting whatever error is returned, in the case where we are pretty darn sure it's going to fail, we should set `rLim.Cur = rLim.Max` and output a more helpful message, like "Warning: Hard limit on file descriptors (4096) is lower than the requested fd_rlimit.  This proxy will only be able to handle ~2048 connections. To hide this warning, please use the fd_rlimit flag to request a limit within the available range."

enocom:
Fixes #413

### DIFF
--- a/proxy/limits/limits.go
+++ b/proxy/limits/limits.go
@@ -66,16 +66,23 @@ func SetupFDLimits(wantFDs uint64) error {
 		rlim2.Max = wantFDs
 		rlim2.Cur = wantFDs
 		if err := syscallSetrlimit(syscall.RLIMIT_NOFILE, rlim2); err == nil {
-			logging.Verbosef("Rlimits for file descriptors set to {%v}", rlim2)
+			logging.Verbosef("Rlimits for file descriptors set to {Current = %v, Max = %v}", rlim2.Cur, rlim2.Max)
 			return nil
 		}
 	}
 
 	rlim.Cur = wantFDs
 	if err := syscallSetrlimit(syscall.RLIMIT_NOFILE, rlim); err != nil {
-		return fmt.Errorf("failed to set rlimit {%v} for max file descriptors: %v", rlim, err)
+		return fmt.Errorf(
+			`failed to set rlimit {Current = %v, Max = %v} for max file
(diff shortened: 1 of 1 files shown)


proxy/limits/limits.go
```diff
@@ -66,16 +66,23 @@ func SetupFDLimits(wantFDs uint64) error {
 		rlim2.Max = wantFDs
 		rlim2.Cur = wantFDs
 		if err := syscallSetrlimit(syscall.RLIMIT_NOFILE, rlim2); err == nil {
-			logging.Verbosef("Rlimits for file descriptors set to {%v}", rlim2)
+			logging.Verbosef("Rlimits for file descriptors set to {Current = %v, Max = %v}", rlim2.Cur, rlim2.Max)
```
Carrotman42, line 69: Optional flyby: consider writing `%+v` instead of `{ ... }`, it has a similar output (https://play.golang.org/p/YLazOI6dGTu)

kurtisvg (approved):

shubha-rajan (approved):

proxy/limits/limits.go
enocom: That's a good idea. I just wish the leading ampersand wasn't included in this case, as it's a user facing message instead of a debug message.

PR #617, opened 2021-02-17
chore: print sha256sum only in release script

enocom:
When running the `sha256sum --binary` command with input from standard
in, the output is:

$SHA_HERE *-

As noted in the man page for sha256sum, the asterix indicates "binary
mode." The trailing dash means standard in.

This commit removes the extraneous information such that only the SHA
will be printed in the final output.

See https://unix.stackexchange.com/questions/487028/print-sha-sums-without-at-the-end
for more.

### DIFF
--- a/.kokoro/release_artifacts.sh
+++ b/.kokoro/release_artifacts.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script distributes the artifacts for the Cloud SQL proxy to their different channels. 
+# This script distributes the artifacts for the Cloud SQL proxy to their different channels.
 
 set -e # exit immediatly if any step fails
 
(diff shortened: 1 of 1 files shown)


kurtisvg:
The [manpage for sha256sum](https://man7.org/linux/man-pages/man1/sha256sum.1.html) seems to indicate that while binary mode doesn't matter for Linux users, it does for Windows users. Should we leave the sha as an indicator to windows users that it's in binary mode?

enocom:
I've verified on my Windows machine that verifying the EXE doesn't require any specification of binary or text mode. So in a word, no. I think we can drop the trailing asterisk.

kurtisvg (approved):

kurtisvg:
SGTM, thanks for checking.

Issue #433, opened 2020-08-20
Add pgx Cloud SQL Proxy in-process example


StevenACoffman:
The authors of the [lib/pq](https://github.com/lib/pq) library are encouraging new projects to use [pgx](https://github.com/jackc/pgx) instead:
 
> This package is effectively in maintenance mode and is not actively developed. Small patches and features are only rarely reviewed and merged. We recommend using [pgx](https://github.com/jackc/pgx) which is actively maintained.

I am connecting to cloudsql from GKE, but I would prefer not to run a sidecar. However, I am using [pgx](https://github.com/jackc/pgx), rather than pg. I have modified the [hook_test.go](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/proxy/dialers/postgres/hook_test.go) file to use pgx as follows:
```go
package main

import (
	"context"
	"fmt"
	"log"
	"net"
	"os"
	"regexp"
	"time"

	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy"
	"github.com/jackc/pgx/v4"
)

var (
	host = os.Getenv("DBHOST") // DBHOST for GCP cloudsql is in format project:region:instance
	name = os.Getenv("DBNAME")
	user = os.Getenv("DBUSER")
	pass = os.Getenv("DBPASS")
)

// parses an addr like "[project:region:instance]:port"
var instanceRegexp = regexp.MustCompile(`^\[(.+)\]:[0-9]+$`)

// Example shows how to use cloudsqlpostgres dialer
func main() {

	// In-process Proxy
	// Note that sslmode=disable is required it does not mean that the connection
	// is unencrypted. All connections via the proxy are completely encrypted.
	dsn := fmt.Sprintf("host=%s user=%s password=%s dbname=%s sslmode=disable", host, user, pass, name)

	config, configErr := pgx.ParseConfig(dsn)
	if configErr != nil {
		log.Fatal(configErr)
	}

	config.DialFunc = func(ctx context.Context, network string, addr string) (net.Conn, error) {
		matches := instanceRegexp.FindStringSubmatch(addr)
		if len(matches) != 2 {
			return nil, fmt.Errorf("failed to parse addr: %q. It should conform to the regular expression %q", addr, instanceRegexp)
		}
		instance := matches[1]
		return proxy.Dial(instance)
	}

	config.LookupFunc = func(ctx context.Context, host string) ([]string, error) {
		return []string{host}, nil
[...]

PR #621, opened 2021-02-18
chore: update README

### LINKED ISSUE #362: Cutting down permissions to a minimum
I've been using the cloudsql-proxy for a while, and wonder if perhaps I've given it too much permissions? Google Cloud Console recommended that I cut down on my permissions to from all of the permissions of a "Cloud SQL Admin" to `cloudsql.instances.connect` and `cloudsql.instances.get` as given to the "Cloud SQL Client" role. Could this be enough? I only know that during the last 89 days, it has been enough, and my proxy has restarted in this time frame as well.

image

The permissions below are "Cloud SQL Admin" permissions that we would loose by downgrading to the "Cloud SQL Client" role, which was recommended by GCP platform based on me not using it anyhow:

```
cloudsql.backupRuns.create
cloudsql.backupRuns.delete
cloudsql.backupRuns.get
cloudsql.backupRuns.list
cloudsql.databases.create
cloudsql.databases.delete
cloudsql.databases.get
cloudsql.databases.list
cloudsql.databases.update
cloudsql.instances.addServerCa
cloudsql.instances.clone
cloudsql.instances.create
cloudsql.instances.delete
cloudsql.instances.demoteMaster
cloudsql.instances.export
cloudsql.instances.failover
cloudsql.instances.import
cloudsql.instances.list
cloudsql.instances.listServerCas
cloudsql.instances.promoteReplica
cloudsql.instances.resetSslConfig
cloudsql.instances.restart
cloudsql.instances.restoreBackup
cloudsql.instances.rotateServerCa
cloudsql.instances.startReplica
cloudsql.instances.stopReplica
cloudsql.instances.truncateLog
cloudsql.instances.update
cloudsql.sslCerts.create
cloudsql.sslCerts.createEphemeral
cloudsql.sslCerts.delete
cloudsql.sslCerts.get
cloudsql.sslCerts.list
cloudsql.users.create
cloudsql.users.delete
cloudsql.users.list
cloudsql.users.update
resourcemanager.projects.get
resourcemanager.projects.list
serviceusage.quotas.get
serviceusage.services.get
serviceusage.services.list
```

## Reference
About the predefined roles "Cloud SQL Admin" and "Cloud SQL Client": https://cloud.google.com/sql/docs/mysql/project-access-control

### LINKED ISSUE #401: Bind address option not documented
## Description

In order to allow  multiple instances or to make the port accessible from outside of the Docker container you have to bind cloud sql proxy to tcp:0.0.0.0:<PORT_NUMBER>, but this setting is not documented in the help of the script or in the documentation page.

---

❯ ./cloud_sql_proxy --version
Cloud SQL Proxy: version 1.16; sha e802c2cb94aec939bfe7497637c5685d3dddd6af built Wed Oct  9 22:22:13 UTC 2019
❯ ./cloud_sql_proxy --help

The Cloud SQL Proxy allows simple, secure connectivity to Google Cloud SQL. It
is a long-running process that opens local sockets (either TCP or Unix sockets)
according to the parameters passed to it. A local application connects to a
Cloud SQL instance by using the corresponding socket.

Connection:
  -instances
    To connect to a specific list of instances, set the instances parameter
    to a comma-separated list of instance connection strings. For example:

           -instances=my-project:my-region:my-instance

    For connectivity over TCP, you must specify a tcp port as part of the
    instance string. For example, the following example opens a loopback TCP
    socket on port 3306, which will be proxied to connect to the instance
    'my-instance' in project 'my-project':

            -instances=my-project:my-region:my-instance=tcp:3306

     When connecting over TCP, the -instances parameter is required.

## Solution

In this Google documentation page, there is an example[1] which is exactly the use case that I'm describing,  this example should be taken as a reference and included in the documentation.

[1]: https://cloud.google.com/sql/docs/mysql/connect-docker#tcp-sockets

By default I'm assuming that the bind is done on 127.0.0.1 so the help section should be probably changed in something similar to this :

---
   ❯ ./cloud_sql_proxy --help

The Cloud SQL Proxy allows simple, secure connectivity to Google Cloud SQL. It
is a long-running process that opens local sockets (either TCP or Unix sockets)
according to the parameters passed to it. A local application connects to a
Cloud SQL instance by using the corresponding socket.

Connection:
  -instances
    To connect to a specific list of instances, set the instances parameter
    to a comma-separated list of instance connection strings. For example:

           -instances=my-project:my-region:my-instance

    For connectivity over TCP, you must specify a tcp port as part of the
    instance string. For example, the following example opens a loopback TCP
    socket on port 3306, which will be proxied to connect to the instance
    'my-instance' in project 'my-project'. For scenarios where you have to make
    the socket open from outside you can change the <custom_bind_address> to 0.0.0.0
       

            -instances=my-project:my-region:my-instance=tcp:<custom_bind_address>:3306

     When connecting over TCP, the -instances parameter is required.
---

### LINKED ISSUE #498: Better documentation for -max_connections flag
## Feature Description

To be able to limit the number of connections or requests on the database via the proxy.

In the case of a container for which we do not master the code but also the configuration, it would be good to be able to limit, as well as the memory, the CPU ... the fact of calling too much Google Cloud SQL.

For example if you have 20 pods using the same CloudSQL instance with 20 databases, one nasty pod that consumes too much I/O from the DB could penalize other application pods.

In all highly shared environments, it would be a big advantage to protect yourself from this kind of side effect.

Obviously if we have control over the code and the applications, it is the developers' duty to configure the connection pools, but in the opposite case, we have to create x Google Cloud SQL instances to protect ourselves but often customers do not want and cannot.

## Alternatives Considered

Either use existing features of MySQL and Postgre drivers or implement this on the proxy side to have something uniform

### LINKED ISSUE #599: Document how to run this project's tests
## Description

Running `go test ./...` from the project root fails on a fresh clone. I don't see any documentation on how to run tests in the README or CONTRIBUTING.

## Solution

There should be a short section describing how to run tests, such that contributors can verify their changes localy.

enocom:
This commit cleans up the project README and consolidates all additional
documentation into the docs folder.

Fixes #362.
Fixes #599.
Fixes #498.
Fixes #401.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,8 @@ cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
 cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432 -ip_address_types=PRIVATE
 ```
 
-For more details, see [Private IP Requirements][private-ip].
+In order to connect using Private IP, you must have access through your
+project's VPC. For more details, see [Private IP Requirements][private-ip].
 
 ## Credentials
 
--- a/README.md
+++ b/README.md
@@ -1,145 +1,333 @@
+# Cloud SQL Proxy
 
-## Cloud SQL Proxy
-The Cloud SQL Proxy allows a user with the appropriate permissions to connect
-to a Second Generation Cloud SQL database without having to deal with IP whitelisting or SSL
-certificates manually. It works by opening unix/tcp sockets on the local machine
-and proxying connections to the associated Cloud SQL instances when the sockets
-are used.
+![CI][ci-badge]
+[![Go Reference][pkg-badge]][pkg-docs]
 
-To build from source, ensure you have [go installed](https://golang.org/doc/install)
-and have set [GOPATH](https://github.com/golang/go/wiki/GOPATH). Then, simply do a go get:
+The [Cloud SQL Proxy][proxy-page] is a binary that provides IAM-based
+authorization and encryption when connecting to a Cloud SQL instance.
 
-    GO111MODULE=on go get github.com/GoogleCloudPlatform/cloudsql-proxy/cmd/cloud_sql_proxy
+See the [Connecting Overview][connection-overview] page for more information on
+connecting to a Cloud SQL instance, or the [About the proxy][about-proxy] page
+for details on how the Cloud SQL proxy works.
 
-The cloud_sql_proxy will be placed in `$GOPATH/bin` after `go get` completes.
+Note: The Proxy *cannot* provide a network path to a Cloud SQL instance if one
+is not already present (e.g., the proxy cannot access a VPC if it does not
+already have access to it).
 
-cloud_sql_proxy takes a few arguments to configure what instances to connect to and connection behavior:
+## Installation
 
-* `-fuse`: requires access to `/dev/fuse` as well as the `fusermount` binary. An
-  optional `-fuse_tmp` flag can specify where to place temporary files. The
-  directory indicated by `-dir` is mounted.
-* `-instances="project1:region:instance1,project3:region:instance1"`: A comma-separated list
-  of instances to open inside `-dir`. Also supports exposing a tcp port and renaming the default Unix Domain Sockets; see examples below.
-  Same list can be provided via INSTANCES environment variable, in case when both are provided - proxy will use command line flag.
-* `-instances_metadata=metadata_key`: Usable on [GCE](https://cloud.google.com/compute/docs/quickstart) only. The given [GCE metadata](https://cloud.google.com/compute/docs/metadata) key will be
-  polled for a list of instances to open in `-dir`. The metadata key is relative from `computeMetadata/v1/`. The format for the value is the same as the 'instances' flag. A hanging-poll strategy is used, meaning that changes to
-  the metadata value will be reflected in the `-dir` even while the proxy is
-  running. When an instance is removed from the list the corresponding socket
-  will be removed from `-dir` as well (unless it was also specified in
-  `-instances`), but any existing connections to this instance will NOT be
-  terminated.
-* `-ip_address_types=PUBLIC,PRIVATE`: A comma-delimited list of preferred IP
-  types for connecting to an instance. For example, setting this to PRIVATE will
-  force the proxy to connect to instances using an instance's associated private
-  IP. Defaults to `PUBLIC,PRIVATE`
-* `-term_timeout=30s`: How long to wait for connections to close before shutting
-  down the proxy. Defaults to 0.
-* `-skip_failed_instance_config`: Setting this flag will allow you to prevent the proxy from terminating when
-	some instance configurations could not be parsed and/or are unavailable.
-* `-log_debug_stdout=true`: This is to log non-error output to stdOut instead of stdErr. For example, if you don't want connection related messages to log as errors, set this flag to true. Defaults to false.
-* `-enable_iam_login`: This enables the proxy to use Cloud SQL IAM database authentication. This will cause the proxy to use IAM account credentials for database user authentication. 
+For 64-bit Linux, run:
+
+```
+wget https://storage.googleapis.com/cloudsql-proxy/v1.19.2/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy
+chmod +x cloud_sql_proxy
+```
+
+Releases for additional OS's and architectures and be found on the [releases
+page][releases].
+
+For alternative distributions, see below under [third party](#third-party).
+
+### Container Images
+
+There are containerized versions of the proxy available from the following
+Google Cloud Container Registry repositories:
+
+* `gcr.io/cloudsql-docker/gce-proxy`
+* `us.gcr.io/cloudsql-docker/gce-proxy`
+* `eu.gcr.io/cloudsql-docker/gce-proxy`
+* `asia.gcr.io/cloudsql-docker/gce-proxy`
+
+Each image is tagged with the associated proxy version. The following tags are
+currently supported:
+
+* `$VERSION` - default image (recommended)
+* `$VERSION-alpine` - uses [`alpine:3`](https://hub.docker.com/_/alpine)
+  as a base image (only supported from v1.17 up)
+* `$VERSION-buster` - uses [`debian:buster`](https://hub.docker.com/_/debian)
+  as a base image (only supported from v1.17 up)
+
+We recommend using the latest version of the proxy and updating the version
+regularly. However, we also recommend pinning to a specific tag and avoid the
+latest tag. Note: the tagged version is only that of the proxy. Changes in base
+images may break specific setups, even on non-major version increments. As such,
+it's a best practice to test changes before deployment, and use automated
+rollbacks to revert potential failures.
+
+### Install from Source
+
+To install from source, ensure you have the latest version of [Go
+installed](https://golang.org/doc/install).
+
+Then, simply run:
+
+```
+go get github.com/GoogleCloudPlatform/cloudsql-proxy/cmd/cloud_sql_proxy
+```
+
+The `cloud_sql_proxy` will be placed in `$GOPATH/bin` after `go get` completes.
+
+## Usage
+
+All the following invocations assume valid credentials are present in the
+environment. The following examples all reference an `INSTANCE_CONNECTION_NAME`,
+which takes the form: `myproject:myregion:myinstance`. To find the
+`INSTANCE_CONNECTION_NAME`, run `gcloud sql instances describe <INSTANCE_NAME>`
+where `INSTANCE_NAME` is the name of the database instance.
+
+### TCP socket example
+
+``` bash
+# Starts the proxy listening on 127.0.0.1:5432
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432
+```
+
+``` bash
+# Starts the proxy listening on on port 5432 on *all* interfaces
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:0.0.0.0:5432
+```
+
+### Unix socket example
+
+``` bash
+# The proxy will mount a Unix domain socket at /cloudsql/<INSTANCE_CONNECTION_NAME>
+# Note: The directory specified by `-dir` must exist.
+cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
+```
+
+### Private IP example
+
+```
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432 -ip_address_types=PRIVATE
+```
+
+For more details, see [Private IP Requirements][private-ip].
+
+## Credentials
+
+The Cloud SQL proxy uses a Cloud IAM account to authorize connections against a
+Cloud SQL instance. The proxy sources the credentials for these accounts in the
+following order:
+
+1. The `-credential_file` flag
+2. The `-token` flag
+3. The service account key at the path stored in the
+   `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
+4. The gcloud user credentials (set from `gcloud auth login`)
+5. The [Application Default Credentials](https://cloud.google.com/docs/authentication/production)
+
+Note: Any account connecting to a Cloud SQL database will need one of the
+following IAM roles:
+
+- Cloud SQL Client (preferred)
+- Cloud SQL Editor
+- Cloud SQL Admin
+
+Or one may manually assign the following IAM permissions:
+
+- `cloudsql.instances.connect`
+- `cloudsql.instances.get`
+
+See [Roles and Permissions in Cloud SQL][roles-and-permissions] for details.
+
+When the proxy authenticates under the Compute Engine VM's default service
+account, the VM must have at least the `sqlservice.admin` API scope (i.e.,
+"https://www.googleapis.com/auth/sqlservice.admin") and the associated project
+must have the SQL Admin API enabled. The default service account must also have
+at least writer or editor privileges to any projects of target SQL instances.
+
+
+## CLI Flags
+
+The Cloud SQL Proxy takes a few arguments to configure what instances to connect
+to and connection behavior. For a full list of flags supported by the proxy,
+use `cloud_sql_proxy -help`.
+
+### Authentication Flags
+
+#### `-credential_file`
+
+Specifies the path to a JSON [service account][service-account] key the proxy
+uses to authorize or authenticate connections.
+
+#### `-token`
+
+When set, the proxy uses this Bearer token for authorization.
+
+#### `-enable_iam_login`
+
+Enables the proxy to use Cloud SQL IAM database authentication. This will cause
+the proxy to use IAM account credentials for database user authentication. For
+details, see [Overview of Cloud SQL IAM database authentication][iam-auth]
+
+### Connection Flags
+
+#### `-instances="project1:region:instance1,project3:region:instance1"`
+
+A comma-separated list of instances to open inside `-dir`. Also supports
+exposing a TCP port and renaming the default Unix Domain Sockets; see examples
+below.  Same list can be provided via INSTANCES environment variable, in case
+when both are provided - proxy will use command line flag.
+
+**Example**
+
+Using TCP sockets:
+
+```
+./cloud_sql_proxy -instances=my-project:us-central1:sql-inst=tcp:3306 &
+mysql -u root -h 127.0.0.1
+```
+
+Using Unix sockets:
+
+```
+./cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:sql-inst &
+mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+```
+
+To specify a custom Unix socket name:
+
+```
+./cloud_sql_proxy -dir=/cloudsql \
+    -instances=my-project:us-central1:sql-inst=unix:custom_socket_name &
+mysql -u root -S /cloudsql/custom_socket_name
+```
+
+To specify a custom location for a Unix socket (overrides `-dir`):
+
+```
+./cloud_sql_proxy -dir=/cloudsql \
+    -instances=my-project:us-central1:sql-inst=unix:/my/custom/sql-socket &
+mysql -u root -S /my/custom/sql-socket
+```
+
+#### `-fuse`
+
+Requires access to `/dev/fuse` as well as the `fusermount` binary. An optional
+`-fuse_tmp` flag can specify where to place temporary files. The directory
+indicated by `-dir` is mounted.
+
+**Example**
+
+Using `-fuse`, you do not need to specify instance names ahead of time:
+
+```
+./cloud_sql_proxy -dir=/cloudsql -fuse &
+mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+```
+
+#### `-instances_metadata=metadata_key`
+
+Usable on [GCE](https://cloud.google.com/compute/docs/quickstart) only. The
+given [GCE metadata](https://cloud.google.com/compute/docs/metadata) key will be
+polled for a list of instances to open in `-dir`. The metadata key is relative
+from `computeMetadata/v1/`. The format for the value is the same as the
+'instances' flag. A hanging-poll strategy is used, meaning that changes to the
+metadata value will be reflected in the `-dir` even while the proxy is running.
+When an instance is removed from the list the corresponding socket will be
+removed from `-dir` as well (unless it was also specified in `-instances`), but
+any existing connections to this instance will NOT be terminated.
+
+**Example**
+
+```
+./cloud_sql_proxy -dir=/cloudsql \
+    -instances_metadata instance/attributes/<custom-metadata-key> &
+mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+```
 
 Note: `-instances` and `-instances_metadata` may be used at the same time but
 are not compatible with the `-fuse` flag.
 
-cloud_sql_proxy authentication can be configured in a few different ways. Those listed higher on the list will override options lower on the list:
+#### `-max_connections`
 
-1. `credential_file` flag
-2. `token` flag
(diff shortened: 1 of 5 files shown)


shubha-rajan (approved):

kurtisvg:
Can we make sure these issues are addressed as well: #498 #401

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+4. gcloud _user_ credentials (set from `gcloud auth login`)
+5. Default Application Credentials via goauth:
+
+   1. `GOOGLE_APPLICATION_CREDENTIALS` (again)
+   2. gcloud _application default_ credentials (set from ` gcloud auth application-default login`)
+   3. appengine.AccessToken (for App Engine Go < =1.9)
+   4. GCE/GAE metadata credentials
+
+Note: any account connecting to a Cloud SQL database will need the
+`roles/cloudsql.client` role. See [Roles and Permissions in Cloud
```
kurtisvg, line 83: Please specifically list the IAM permissions needed or`Cloud SQL client` role or higher. The App Engine page does a pretty good job:
image

README.md
```diff
@@ -1,144 +1,86 @@
...
 
-### .Net Proxy Wrapper (Nuget Package)
-  
-  Install via Nuget, follow these [instructions](https://github.com/expert1-pty-ltd/cloudsql-proxy#install-via-nuget).
+``` bash
+# Must create a /cloudsql directory first
+cloud_sql_proxy -dir=/cloudsql -instances=my-project:us-central1:my-instance
+```
+
+#### TCP socket example
```
kurtisvg, line 49: Please move TCP to the top - going forward, it should be messaged as the default path.

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+
+* `-term_timeout=30s`
+
+How long to wait for connections to close before shutting down the proxy.
+Defaults to 0.
+
+* `-skip_failed_instance_config`
+
+Setting this flag will allow you to prevent the proxy from terminating when some
+instance configurations could not be parsed and/or are unavailable.
```
kurtisvg, line 50:
```suggestion
Setting this flag will allow you to prevent the proxy from terminating if any errors
occur during instance configuration. Please note that this may mean some instances
may fail to set up correctly while others work if the proxy restarts..
```

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+project must have the SQL Admin API enabled.  The default service account must
+also have at least WRITER/EDITOR priviledges to any projects of target SQL
+instances.
+
+## Example Invocations
+
+The following examples demonstrate various features of the proxy and included
+the associate command to connect to a MySQL database from the CLI.
+
+Using Unix Sockets:
```
kurtisvg, line 98: nit
```suggestion
Using Unix sockets:
```

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Using `-fuse`, you do not need to specify instance names ahead of time:
+
+```
+./cloud_sql_proxy -dir=/cloudsql -fuse &
+mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+```
+
+Using TCP sockets:
```
kurtisvg, line 120: Please provide TCP options first.

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+SQL](https://cloud.google.com/sql/docs/mysql/roles-and-permissions) for details.
+
+When the proxy authenticates under the default service account of the Compute
+Engine VM it is running on the VM must have at least the sqlservice.admin API
+scope ("https://www.googleapis.com/auth/sqlservice.admin") and the associated
+project must have the SQL Admin API enabled.  The default service account must
+also have at least WRITER/EDITOR priviledges to any projects of target SQL
+instances.
+
+## Example Invocations
```
kurtisvg, line 93: At a more meta level, I prefer when the examples are next to the documentation for the flag itself. If someone just wants a list of flags they can use `-help`.

docs/library-usage.md
```diff
@@ -0,0 +1,25 @@
+# Cloud SQL proxy dialer for Go
```
kurtisvg, line 1: IMO, this should stay in the README under the proxy module.

docs/library-usage.md
```diff
@@ -0,0 +1,25 @@
...
+
+You can also use the Cloud SQL proxy directly from a Go program.
+
+These packages are primarily used as implementation for the Cloud SQL proxy
+command, and may be changed in backwards incompatible ways in the future.
+
+## To use inside a Go program:
```
kurtisvg, line 8:
```suggestion
## Usage
```

docs/library-usage.md
```diff
@@ -0,0 +1,25 @@
...
+
+If your program is written in [Go](https://golang.org) you can use the Cloud SQL
+Proxy as a library, avoiding the need to start the Proxy as a companion process.
+
+### MySQL
```
kurtisvg, line 13: Can you add a section here for using the Dial function directly? See #433

docs/library-usage.md
```diff
@@ -0,0 +1,25 @@
...
+[go-sql-driver](https://github.com/go-sql-driver/mysql)
+you can use helper functions found in the [`proxy/dialers/mysql`package](https://godoc.org/github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/dialers/mysql).
+
+See [example usage](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/tests/dialers_test.go).
+
+### Postgres
+
+If you're using the the Postgres [lib/pq](https://github.com/lib/pq), you can use the `cloudsqlpostgres` driver from [here](https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/master/proxy/dialers/postgres).
+
+See [example usage](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/proxy/dialers/postgres/hook_test.go).
```
kurtisvg, line 25: nit: consider using relative paths instead

docs/library-usage.md
```diff
@@ -0,0 +1,25 @@
...
+
+You can also use the Cloud SQL proxy directly from a Go program.
+
+These packages are primarily used as implementation for the Cloud SQL proxy
+command, and may be changed in backwards incompatible ways in the future.
+
+## To use inside a Go program:
+
+If your program is written in [Go](https://golang.org) you can use the Cloud SQL
+Proxy as a library, avoiding the need to start the Proxy as a companion process.
```
kurtisvg, line 11: maybe call out the Java and python connectors as well.

docs/install-from-source.md
```diff
@@ -0,0 +1,8 @@
+# Install from Source
```
kurtisvg, line 1: IMO, this section it too short to have it's own page. I'd rather have it in the main README.

docs/kubernetes-sidecar.md
```diff
@@ -0,0 +1,5 @@
+# Running as a Kubernetes Sidecar
```
kurtisvg, line 1: Same with this section.

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+mysql -u root -S /cloudsql/custom_socket_name
+```
+
+To specify a custom location for a Unix socket (overrides `-dir`):
+
+```
+./cloud_sql_proxy -dir=/cloudsql \
+    -instances=my-project:us-central1:sql-inst=unix:/my/custom/sql-socket &
+mysql -u root -S /my/custom/sql-socket
+```
```
kurtisvg, line 141: I would also add an example + context around Public vs Private IP, as well as an example that shows how to specify the bind interface.

kurtisvg (changes requested):

README.md
```diff
@@ -1,144 +1,86 @@
...
-rollbacks to revert potential failures. 
-
-## To use from Kubernetes:
-
-### Deploying Cloud SQL Proxy as a sidecar container
-Follow this [page](https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/main/examples/kubernetes#run-the-cloud-sql-proxy-as-a-sidecar). See also
-[Connecting from Google Kubernetes Engine](https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine).
+### Installation
 
+The proxy may be installed as a [gcloud component][gcloud-install], or as a
```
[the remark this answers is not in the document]
enocom: Do you mean downloading the binary from dl.google.com as shown here? https://cloud.google.com/sql/docs/mysql/quickstart-proxy-test#install-proxy

README.md
```diff
@@ -1,144 +1,86 @@
...
+following Google Cloud Container Registry repositories:
 
-__WARNING__: _These distributions are not officially supported by Google._
+* `gcr.io/cloudsql-docker/gce-proxy`
+* `us.gcr.io/cloudsql-docker/gce-proxy`
+* `eu.gcr.io/cloudsql-docker/gce-proxy`
+* `asia.gcr.io/cloudsql-docker/gce-proxy`
 
-### Installing via Homebrew
+Each image is tagged with the associated proxy version. The following tags are
```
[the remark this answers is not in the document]
enocom: Done.

:+1:

docs/cli-usage.md
```diff
@@ -0,0 +1,141 @@
...
+## Authentication
+
+Authentication can be configured in a few different ways. Those listed higher on
+the list will override options lower on the list:
+
+1. `credential_file` flag
+2. `token` flag
+3. Service account key at path stored in `GOOGLE_APPLICATION_CREDENTIALS`
+4. gcloud _user_ credentials (set from `gcloud auth login`)
+5. Default Application Credentials via goauth:
```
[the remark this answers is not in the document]
enocom: It seems this list can be simplified with a link to the [Application Default Credentials](https://cloud.google.com/docs/authentication/production) (which is what the [Goauth library](https://pkg.go.dev/golang.org/x/oauth2/google) links to itself).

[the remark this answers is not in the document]
enocom: I just noticed that -- this was an existing typo. It's fixed now.

docs/cli-usage.md
enocom: I'll move all examples below their flags. I agree it's more useful.

docs/library-usage.md
enocom: Happy to do either -- just want to be consistent with the socket factory.

docs/install-from-source.md
enocom: Done.

docs/kubernetes-sidecar.md
enocom: Done.

README.md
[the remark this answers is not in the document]
kurtisvg: I would prefer we move away from the dl.google.com. The releases pages uses links to a storage bucket which is versioned, that we should transition to the main source of "blessed" binaries.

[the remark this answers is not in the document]
enocom: Got it. Since the one-liner would have to be provided for each os and architecture, how about just a link to the releases page?

enocom:
I've addressed almost all the comments and resolved #498 and #401.

Here's what's left:

- [ ] [add an example + context around Public vs Private IP, as well as an example that shows how to specify the bind interface.](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/621#discussion_r578812630)
- [ ] [Add an example that mentions how to specify IP types (making sure to call out how access via Private IP works, and linking to the Cloud docs about it)](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/621#discussion_r578804732)
- [ ] [Add a section here for using the Dial function (See #433)](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/621#discussion_r578811412)

docs/library-usage.md
enocom: Given that we're working on structural improvements to the proxy, should we wait on this since its API will likely change?

README.md
```diff
@@ -1,145 +1,308 @@
...
+for details on how the Cloud SQL proxy works.
+
+Note: The Proxy *cannot* provide a network path to a Cloud SQL instance if one
+is not already present (e.g., the proxy cannot access a VPC if it does not
+already have access).
+
+## Installation
+
+The proxy may be installed from the [releases page][releases].
+
```
kurtisvg, line 20: I think we should consider just picking what we think the most popular distribution (64x Linux?), and give users a command to download and set as an executable. I would then change this to something "Releases for additional OS's and architectures and be found on the releases pages).

README.md
```diff
@@ -1,145 +1,308 @@
...
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432
+```
+
+### Unix socket example
+
+``` bash
+# Must create a /cloudsql directory first
+cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
+```
+
```
kurtisvg, line 87: nit: Add a line here that says something like "This will cause the proxy to mount a Unix domain socket a `/cloudsql/<INSTANCE_CONNECTION_NAME>`

README.md
```diff
@@ -1,145 +1,308 @@
...
 are not compatible with the `-fuse` flag.
 
-cloud_sql_proxy authentication can be configured in a few different ways. Those listed higher on the list will override options lower on the list:
+* `-max_connections`
+
+If provided, the maximum number of connections to establish before refusing new
+connections. Defaults to 0 (no limit). Note: requests made through the
+SQL Admin API are rate limited to 2x the number of connections per minute.
+
+### Credentials and Authentication Flags
```
kurtisvg, line 178: I think this first part of this section (why credentials are important and in what order the proxy gets them - up to about the `-credential_file` flag) should be elevated out of examples. Understanding where the credentials come from is an important step to running the proxy.

README.md
```diff
@@ -1,145 +1,308 @@
...
+
+Authentication can be configured in a few different ways. Those listed higher on
+the list will override options lower on the list:
```
kurtisvg, line 181: I think we need more emphasis around why the credentials are important here:
"The Cloud SQL proxy uses a Cloud IAM account to authorize connections against a Cloud SQL instance. The proxy sources the credentials for these accounts in the following order:"

README.md
```diff
@@ -1,145 +1,308 @@
...
-When the proxy authenticates under the default service account of the
-Compute Engine VM it is running on the VM must have at least the
-sqlservice.admin API scope ("https://www.googleapis.com/auth/sqlservice.admin")
-and the associated project must have the SQL Admin API
-enabled.  The default service account must also have at least WRITER/EDITOR
-priviledges to any projects of target SQL instances.
-
-Specifying the `-credential_file` flag allows use of the proxy outside of
-Google's cloud. Simply [create a new service
+3. [Application Default Credentials](https://cloud.google.com/docs/authentication/production)
```
kurtisvg, line 185: It's very important we don't loose #3 and #4 here - `gcloud auth login` is actually an entirely different than the `auth application-default` version, so both of these steps need to stay.

README.md
```diff
@@ -1,145 +1,308 @@
...
+IAM roles:
+
+- Cloud SQL Client (preferred)
+- Cloud SQL Editor
+- Cloud SQL Admin
+
+Or you can manually assign the following IAM permissions:
+
+- cloudsql.instances.connect
+- cloudsql.instances.get
```
kurtisvg, line 197: nit:
```suggestion
- `cloudsql.instances.connect`
- `cloudsql.instances.get`
```

README.md
```diff
@@ -1,145 +1,308 @@
...
+
+See [Roles and Permissions in Cloud
+SQL](https://cloud.google.com/sql/docs/mysql/roles-and-permissions) for details.
+
+When the proxy authenticates under the default service account of the Compute
```
kurtisvg, line 202: Start this section with "On Google Compute Engine (GCE), when using the default service account the VM *must* have at least the sqlservice.admin API scope`. 

We want to make sure to start with "Google Compute Engine (GCE)" so it's easier to read at a glance and determine if the user needs to keep reading.

README.md
```diff
@@ -1,145 +1,308 @@
...
+
+* `-credential_file`
+
+Allows use of the proxy outside of Google's cloud. Simply [create a new service
 account](https://cloud.google.com/sql/docs/mysql/sql-proxy#create-service-account),
 download the associated JSON file, and set `-credential_file` to the path of the
-JSON file. You can also set the GOOGLE_APPLICATION_CREDENTIALS environment variable
-instead of passing this flag.
+JSON file. You can also set the `GOOGLE_APPLICATION_CREDENTIALS` environment
+variable instead of passing this flag.
```
kurtisvg, line 215: nit: This section feels very verbose. Can we try to make more concise? Maybe:

"Specifies the path to a JSON [service account](https://cloud.google.com/iam/docs/service-accounts) key the proxy uses to authorize or authenticate connections."

README.md
```diff
@@ -1,145 +1,308 @@
...
 
-    # To retrieve instances from a custom metadata value (only when running on GCE)
-    ./cloud_sql_proxy -dir=/cloudsql -instances_metadata instance/attributes/<custom-metadata-key> &
-    mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+* `-enable_iam_login`
 
-    # For -fuse you do not need to specify instance names ahead of time:
-    ./cloud_sql_proxy -dir=/cloudsql -fuse &
-    mysql -u root -S /cloudsql/my-project:us-central1:sql-inst
+Enables the proxy to use Cloud SQL IAM database authentication. This will cause
```
kurtisvg, line 223: Let's add a link to this page: https://cloud.google.com/sql/docs/postgres/authentication

README.md
```diff
@@ -1,145 +1,308 @@
...
+where `INSTANCE_NAME` is the name of the database instance.
+
+### TCP socket example
+
+``` bash
+# Starts the proxy listening on 127.0.0.1:5432
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432
+```
+
+### Unix socket example
```
kurtisvg, line 81: I think we should defiantly add a section following this one that's something like "Example using private IP"

README.md
```diff
@@ -1,145 +1,308 @@
...
+
+``` bash
+# Must create a /cloudsql directory first
+cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
+```
+
+## CLI Flags
+
+The Cloud SQL Proxy takes a few arguments to configure what instances to connect
```
kurtisvg, line 90: Maybe add "For a full list of flags supported by the proxy, use the `-help` flag."

README.md
```diff
@@ -1,145 +1,308 @@
...
 
 Note: `-instances` and `-instances_metadata` may be used at the same time but
 are not compatible with the `-fuse` flag.
 
-cloud_sql_proxy authentication can be configured in a few different ways. Those listed higher on the list will override options lower on the list:
+* `-max_connections`
+
+If provided, the maximum number of connections to establish before refusing new
+connections. Defaults to 0 (no limit). Note: requests made through the
+SQL Admin API are rate limited to 2x the number of connections per minute.
```
kurtisvg, line 176: > Note: requests made through the SQL Admin API are rate limited to 2x the number of connections per minute.

We should probably just remove this line - this is incorrect:
1. The total number of _refresh attempts_ is limited to 2 time per minute _per instance_
2. There is additional retry logic _per request_. Currently each request can be tried up to 5x times (this may change soon)

README.md
```diff
@@ -1,145 +1,308 @@
...
+```
 
 Note: `-instances` and `-instances_metadata` may be used at the same time but
 are not compatible with the `-fuse` flag.
 
-cloud_sql_proxy authentication can be configured in a few different ways. Those listed higher on the list will override options lower on the list:
+* `-max_connections`
+
+If provided, the maximum number of connections to establish before refusing new
+connections. Defaults to 0 (no limit). Note: requests made through the
```
kurtisvg, line 175:
```suggestion
If provided, the maximum number of connections to establish before refusing new
connections. Defaults to 0 (no limit).
```

We need to disambiguate between max connections to the instance and max connections to the proxy (and also if max connections is _per Cloud SQL instance_ or _per invocation of the proxy_ (I _think_ its the second but I can't remember off hand).

kurtisvg (changes requested):

README.md
enocom: Done.

README.md
enocom: How about moving the section to be the first under CLI Flags?

README.md
enocom: Done.

README.md
enocom: Restored.

README.md
enocom: Done.

README.md
enocom: Done.

README.md
enocom: Agreed.

README.md
kurtisvg: I think it needs to be a top level section - or maybe even work it into usage somehow? It feels  a bit long though, but it's important to how the proxy works. Maybe put it after usage but have a one liner that references it below?

README.md
enocom: Done. PTAL.

README.md
enocom: Done.

README.md
enocom: For now I've removed this note entirely.

README.md
enocom: For now I've removed the note. Future passes can add this detail back in, especially if (or when) it changes.

README.md
```diff
@@ -1,145 +1,333 @@
...
+cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
+```
+
+### Private IP example
+
+```
+cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:5432 -ip_address_types=PRIVATE
+```
+
+For more details, see [Private IP Requirements][private-ip].
```
kurtisvg, line 104: nit:
```suggestion
In order to connect using Private IP, you must have access through your projects VPC. For more details, see [Private IP Requirements][private-ip].
```

kurtisvg (approved):

README.md
enocom: Fixed.

PR #631, opened 2021-02-24
feat: add ARM releases

shubha-rajan:
## Change Description

Added steps to compile binaries for Linux/ARM32 and Linux/ARM64, as well as Darwin/ARM64. The Linux versions have been tested with QEMU, but not on an actual ARM machine yet. 
Not sure how to test the Darwin one (for M1 Macs), although it's probably low priority, since Apple says that [the new Macs have Rosetta to run programs compiled for AMD64](https://support.apple.com/en-us/HT211861). 

## Checklist

### DIFF
--- a/.build/gcs_upload.yaml
+++ b/.build/gcs_upload.yaml
@@ -36,6 +36,24 @@ steps:
     args:
       - "-c"
       - 'go build -ldflags "-X main.versionString=${_VERSION} -X main.metadataString=$$GOOS.$$GOARCH" -o cloud_sql_proxy.$$GOOS.$$GOARCH ./cmd/cloud_sql_proxy'
+  - id: linux.arm64
+    name: "golang:1.16"
+    env:
+      - "GOOS=linux"
+      - "GOARCH=arm64"
+    entrypoint: "bash"
+    args:
+      - "-c"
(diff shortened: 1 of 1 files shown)


.build/gcs_upload.yaml
```diff
@@ -19,7 +19,7 @@ options:
 
 steps:
   - id: linux.amd64
-    name: "golang:1.14"
```
enocom, line 22: Was there a particular reason we'd pinned to Go 1.14 here?

enocom (approved):
LGTM

.build/gcs_upload.yaml
shubha-rajan: It was specifically because Go dropped support for cross-compiling for 32-bit Darwin in 1.15. So we decided to compile all of them with the same version, but at our last team sync, we decided to compile all other binaries with 1.16 except for the `darwin.386` one

kurtisvg:
Can we separate this into 2 prs? One for moving the version we compile with to 1.16 and one for adding the ARM releases? Seems like it should be 2 different items in the release notes

kurtisvg:
We should also add Go 1.16 tests before we use that to compile, just in case :)

shubha-rajan:
> We should also add Go 1.16 tests before we use that to compile, just in case :)

Adding those this afternoon! We'll merge #634 after tests pass.

PR #638, opened 2021-02-26
chore: add GitHub action to check for version string mismatch

shubha-rajan:
## Change Description

Adds a GitHub Actions check that looks for a mismatch between the version string in `cmd/cloud_sql_proxy/cloud_sql_proxy.go` and `version.txt`

## Checklist


shubha-rajan:
The new check is currently failing because the versions don't match. Once we merge #637 it should pass

.github/workflows/tests.yaml
```diff
@@ -16,19 +16,30 @@ name: tests
...
       run: go get golang.org/x/tools/cmd/goimports
     - name: Checkout code
       uses: actions/checkout@v2
     - run: goimports -w .
     - run: go mod tidy
     - name: Verify no changes from goimports and go mod tidy. If you're reading this and the check has failed, run `goimports -w . && go mod tidy`.
-      run: git diff --exit-code
\ No newline at end of file
+      run: git diff --exit-code
+  version-mismatch:
```
kurtisvg, line 35: Should this just be part of link maybe? Then we don't have to do another checkout step and it's still just 1 status

kurtisvg (changes requested):

.github/workflows/tests.yaml
shubha-rajan: sure

kurtisvg (approved):

PR #644, opened 2021-02-28
Fix goroutine leak

### LINKED ISSUE #427: Multiple calls to refreshCfg leak goroutines
## Bug Description

There can be multiple Goroutines waiting on [`refreshCertAfter`](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v1.17.0/proxy/proxy/client.go#L230) for the same database instance. If [Dial](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v1.17.0/proxy/proxy/client.go#L282) fails [tryConnect](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v1.17.0/proxy/proxy/client.go#L284) of `cachedCfg` it calls `refreshCfg` which creates another Goroutine waiting `refreshCertAfter`.

It seems that each refreshing Goroutine will keep refreshing itself every hour and they only die off if they get limited by refreshCfg throttling.

We had misconfigured routing to the SQL instance's IP address. Requests to SQL instance worked sometimes and sometimes failed on [handshake](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v1.17.0/proxy/proxy/client.go#L322).

Effectively this led to massive spam of `ephemeral certificate for instance x will expire soon, refreshing now.` in our logs after the service had run for an hour. There was no issue with ephemeral certificates but only with connection to the database.

In our case it was a configuration issue but a long-running `cloudsql-proxy` can ocassionally fail database connections and leak ephemeral certificate refreshes Goroutines. refreshCfg throttling gives some protection against this.

## How to reproduce

  1. Sometimes route Cloud SQL instance IP to wrong IP (or randomly fail [tryConnect ](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v1.17.0/proxy/proxy/client.go#L297))
  2. Make database queries
  3. Wait for an hour to see multiple `ephemeral certificate for instance x will expire soon, refreshing now` log entries

## Environment

1. OS type and version: `golang:1.14-stretch` image
2. Cloud SQL Proxy version (`./cloud_sql_proxy -version`): 1.17.0

enocom:
## Change Description

This PR removes a leak in the proxy client caused by not stopping [this goroutine](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/bbaf1acf2afb90c458562929da132451ad1c8f2c/proxy/proxy/client.go#L252) after subsequent connection failures.

## Checklist

## Relevant issues:

- Fixes #427

### DIFF
--- a/proxy/proxy/client_test.go
+++ b/proxy/proxy/client_test.go
@@ -92,6 +92,9 @@ func TestContextDialer(t *testing.T) {
 	if _, err := c.DialContext(context.Background(), instance); err != errFakeDial {
 		t.Errorf("unexpected error: %v", err)
 	}
+
+	// leakcheck.RegisterIgnoreGoroutine("go.opencensus.io@v0.22.5/stats/view/worker.go")
+	// leakcheck.Check(t)
 }
 
 func TestClientCache(t *testing.T) {
--- a/internal/leakcheck/leakcheck.go
+++ b/internal/leakcheck/leakcheck.go
@@ -1,0 +1,119 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package leakcheck contains functions to check leaked goroutines.
+//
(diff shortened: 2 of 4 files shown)


proxy/proxy/client_test.go
```diff
@@ -92,6 +92,9 @@ func TestContextDialer(t *testing.T) {
 	if _, err := c.DialContext(context.Background(), instance); err != errFakeDial {
 		t.Errorf("unexpected error: %v", err)
 	}
+
+	// leakcheck.RegisterIgnoreGoroutine("go.opencensus.io@v0.22.5/stats/view/worker.go")
+	// leakcheck.Check(t)
```
enocom, line 97: Leaving this here for reviewers to see how I wrote a failing test with the leak, and then fixed it. Before merging this PR, these two lines will be deleted.

enocom:
Closing this as @kurtisvg has some ideas about a better approach.

PR #646, opened 2021-02-28
chore: ensure IAM user is present before test run

enocom:
Also, add the associated environment variable to .envrc.example.


shubha-rajan (approved):
LGTM

kurtisvg (changes requested):
It would be nice to have this only block the tests that need this variable.

shubha-rajan (approved):

kurtisvg (approved):

PR #665, opened 2021-03-08
chore: create .gcloudignore

enocom:
Adds a gcloudignore file to prevent accidental omissions when adding to gitignore.


kurtisvg (changes requested):
Are we sure we want to add this?

We should also add .envrc files as well

enocom:
With better defined directives in the gitignore, we probably don't need this. Just wanted it available for reference.

kurtisvg:
It's maybe worth updating the .gitignore file that any additions to it also are reflected in Cloud Build

Issue #662, opened 2021-03-05
Enable IAM Authn fails to refresh certificate when using ADC retrieved from a metadata server


enocom:
## Bug Description

When running the proxy on GKE with the `-enable_iam_login` flag, the proxy will initially succeed at startup, but about 1 hour later when it tries to refresh the ephemeral certificate, it fails with a message like the following:

```
2021/03/05 22:48:21 failed to refresh the ephemeral certificate for kvg-testing:us-central1:test2, returning previous cert instead: new ephemeral certificate expires too soon: current time: 2021-03-05 22:48:21.070824294 +0000 UTC m=+302.212008501, certificate expires: 2021-03-05 23:43:20.959727769 +0000 UTC m=+3602.100911936
```

This is likely related to how we're doing OAuth2 token refresh, which differs based on the credentials in use.

Related to #643.

kurtisvg:
Specifically, we think this is related to where the ADC are generated from. We think the issue is that [the ADC return a `gaeTokenSource`](https://github.com/golang/oauth2/blob/9bb904979d93d76cf7000f77c4a1d7bc06c7a708/google/appengine_gen1.go#L31) that doesn't use a ReuseTokenSource, but [instead does some internal caching](https://github.com/golang/oauth2/blob/9bb904979d93d76cf7000f77c4a1d7bc06c7a708/google/appengine_gen1.go#L56-L62), which is preventing us from getting a new token early.

kurtisvg:
I was able to verify it's not a problem with the `gaeTokenSource` - the token is definitely a [`ComputeTokenSource`](https://github.com/golang/oauth2/blob/9bb904979d93d76cf7000f77c4a1d7bc06c7a708/google/google.go#L183). 

Not sure why or where the token is getting cached for whatever reason. Might be happening on the server side.

enocom:
This is not a problem with golang.org/x/oauth2. 

The `ComputeTokenSource` is [calling the metadata service](https://github.com/golang/oauth2/blob/9bb904979d93d76cf7000f77c4a1d7bc06c7a708/google/google.go#L200-L206). In effect, it's making the following call:

```
$ curl -i -H "Metadata-Flavor: Google" $GCE_METADATA_HOST/computeMetadata/v1/instance/service-accounts/default/token

HTTP/1.0 200 OK
Server: BaseHTTP/0.3 Python/2.7.13
Date: Wed, 10 Mar 2021 15:43:15 GMT
[...]

PR #667, opened 2021-03-08
fix: shorten refresh window when cert duration is less than expected

kurtisvg:
## Change Description

Shortens the time before doing a new refresh when the duration is less than expected, rather than returning an error. 

## Checklist

## Relevant issues:

- Related to #662 (but probably doesn't fix)


enocom:
I see this has been pulled in to #674. I'm going to close this. Please re-open if I've misunderstood.

PR #674, opened 2021-03-10
fix: fix cache to prevent multiple concurrent refreshes

### LINKED ISSUE #427: Multiple calls to refreshCfg leak goroutines

kurtisvg:
## Change Description

Expands on the behavior introduced in #666 by reconfiguring the `cachedCfg` function to prevent multiple simultaneous refreshes from occurring at once. 

## Checklist

## Relevant issues:

- Fixes #427

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -256,9 +256,8 @@ func isExpired(cfg *tls.Config) bool {
 	return time.Now().After(cfg.Certificates[0].Leaf.NotAfter)
 }
 
-// startRefresh kicks off a refreshCfg asynchronously, returning a channel that is closed
-// when it's complete and updates the cacheEntry when complete. This function should only
-// be called from the scope of "cachedCfg", which controls the logic around throttling refreshes.
+// startRefresh kicks off a refreshCfg asynchronously, that updates the cacheEntry and closes the returned channel once the refresh is completed. This function
+// should only be called from the scope of "cachedCfg", which controls the logic around throttling refreshes.
 func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) chan struct{} {
 	done := make(chan struct{})
 	go func() {
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -126,8 +126,9 @@ type cacheEntry struct {
 	addr    string
 	version string
 	cfg     *tls.Config
-	// done is a channel that will be closed once any pending refresh for the instance is complete.
-	done chan bool
+	// done represents the status of any pending refresh operation related to this instance.
+	// If unset the op hasn't started, if open the op is still pending, and if closed the op has finished.
+	done chan struct{}
 }
 
 // Run causes the client to start waiting for new connections to connSrc and
@@ -258,8 +259,8 @@ func isExpired(cfg *tls.Config) bool {
 // startRefresh kicks off a refreshCfg asynchronously, returning a channel that is closed
 // when it's complete and updates the cacheEntry when complete. This function should only
 // be called from the scope of "cachedCfg", which controls the logic around throttling refreshes.
-func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) chan bool {
-	done := make(chan bool)
+func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) chan struct{} {
+	done := make(chan struct{})
 	go func() {
 		defer close(done)
 		addr, cfg, ver, err := c.refreshCfg(instance)
@@ -299,14 +300,15 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
 
 // isValid returns true if the cacheEntry is still useable
 func isValid(c cacheEntry) bool {
-	// if the cfg is unset or if there was an error returned during the last attempt
-	return !(c.cfg == nil || c.err != nil)
+	// the entry is only valid there wasn't an error retrieving it and it has a cfg
+	return c.err == nil && c.cfg != nil
 }
 
 func needsRefresh(e cacheEntry, refreshCfgBuffer time.Duration) bool {
 	if e.done == nil { // no refresh started
 		return true
-	} else if !isValid(e) || time.Since(e.lastRefreshed) >= refreshCfgBuffer {
+	}
+	if !isValid(e) || time.Since(e.lastRefreshed) >= refreshCfgBuffer {
 		// if the entry is invalid or close enough to expiring check
 		// use the entry's done channel to determine if a refresh has started yet
 		select {
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -183,6 +183,7 @@ func (c *Client) handleConn(conn Conn) {
 func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, version string, err error) {
 	c.refreshCfgL.Lock()
 	defer c.refreshCfgL.Unlock()
+	logging.Verbosef("refreshing ephemeral certificate for instance %s", instance)
 
 	mycert, err := c.Certs.Local(instance)
 	if err != nil {
@@ -286,7 +287,7 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
 		timeToRefresh := certExpiration.Sub(now) - refreshCfgBuffer
 		if timeToRefresh <= 0 {
 			// If a new certificate expires before our buffer has expired, we should wait a bit and schedule a new refresh to much closer to the expiration's date
-			// This situation probably only occurs when the oauth2 token isn't refreshed before the cert is, so by scheduling closer to the expriation we can hope the ouath2 token is newer.
+			// This situation probably only occurs when the oauth2 token isn't refreshed before the cert is, so by scheduling closer to the expiration we can hope the oauth2 token is newer.
 			timeToRefresh = certExpiration.Sub(now) - (5 * time.Second)
 			logging.Errorf("new ephemeral certificate expires sooner than expected (adjusting refresh time to compensate): current time: %v, certificate expires: %v", now, certExpiration)
 		}
@@ -306,7 +307,7 @@ func needsRefresh(e cacheEntry, refreshCfgBuffer time.Duration) bool {
 	if e.done == nil { // no refresh started
 		return true
 	}
-	if !isValid(e) || time.Since(e.lastRefreshed) >= refreshCfgBuffer {
+	if !isValid(e) || e.cfg.Certificates[0].Leaf.NotAfter.Sub(time.Now()) <= refreshCfgBuffer {
 		// if the entry is invalid or close enough to expiring check
 		// use the entry's done channel to determine if a refresh has started yet
 		select {
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -218,7 +218,7 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, vers
 func (c *Client) refreshCertAfter(instance string, timeToRefresh time.Duration) {
 	<-time.After(timeToRefresh)
 	logging.Verbosef("ephemeral certificate for instance %s will expire soon, refreshing now.", instance)
-	if _, _, _, err := c.cachedCfg(instance); err != nil {
+	if _, _, _, err := c.cachedCfg(context.Background(), instance); err != nil {
 		logging.Errorf("failed to refresh the ephemeral certificate for %s before expiring: %v", instance, err)
 	}
 }
@@ -320,7 +320,7 @@ func needsRefresh(e cacheEntry, refreshCfgBuffer time.Duration) bool {
 	return false
 }
 
-func (c *Client) cachedCfg(instance string) (string, *tls.Config, string, error) {
+func (c *Client) cachedCfg(ctx context.Context, instance string) (string, *tls.Config, string, error) {
 	c.cacheL.RLock()
 
 	throttle := c.RefreshCfgThrottle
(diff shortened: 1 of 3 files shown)


enocom (commented):
Adding some initial comments. I have yet to load test this PR.

proxy/proxy/client.go
```diff
@@ -292,33 +240,126 @@ func isExpired(cfg *tls.Config) bool {
...
+			timeToRefresh = certExpiration.Sub(now) - (5 * time.Second)
+			logging.Errorf("new ephemeral certificate expires sooner than expected (adjusting refresh time to compensate): current time: %v, certificate expires: %v", now, certExpiration)
+		}
+		go c.refreshCertAfter(instance, timeToRefresh)
+	}()
+	return done
+}
+
+// isValid returns true if the cacheEntry is still useable
+func isValid(c cacheEntry) bool {
```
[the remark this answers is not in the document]
kurtisvg: I generally prefer sticking with positive checks - go/tott-530

proxy/proxy/client.go
```diff
@@ -292,33 +240,126 @@ func isExpired(cfg *tls.Config) bool {
...
+		}
+	}
+	return false
+}
+
+func (c *Client) cachedCfg(instance string) (string, *tls.Config, string, error) {
 	c.cacheL.RLock()
-	ret, ok := c.cfgCache[instance]
+
+	throttle := c.RefreshCfgThrottle
```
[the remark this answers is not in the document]
kurtisvg: tbh, I'm not sure. Both fields are under the `refreshCfgL` mutex, which makes me think they might be projected by it. However I'm now realizing they both are exported, which isn't great since the mutex isn't. 

They were previously inside a critical section, so I attempted to maintain the same here.

proxy/proxy/client.go
```diff
@@ -292,33 +240,126 @@ func isExpired(cfg *tls.Config) bool {
...
+			lastRefreshed: time.Now(),
+			err:           err,
+			addr:          addr,
+			version:       ver,
+			cfg:           cfg,
+			done:          done,
+		}
+		c.cacheL.Unlock()
+
+		certExpiration := cfg.Certificates[0].Leaf.NotAfter
```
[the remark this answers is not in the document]
kurtisvg: I'm not sure I understand what your suggestion is. Are you suggesting moving it into `refreshCfg` or just scoping to `startRefresh`? 

You don't want to move it into `refreshCfg`, because you want to update the cacheEntry and schedule the refresh again, regardless of how the last operation went.

You don't want it to move scope up into `startRefresh` because you don't have the results of the refresh operation yet.

proxy/proxy/client.go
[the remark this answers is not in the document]
enocom: Interesting. TIL.

proxy/proxy/client.go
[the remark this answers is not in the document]
enocom: Based on internal usage in this file, I see no other writes to these fields. I suspect these fields are intended to be supplied when the `Client` is created and not modified later. But since they're exported, I suppose it's possible they could be changed later.

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
+		now := time.Now()
+		timeToRefresh := certExpiration.Sub(now) - refreshCfgBuffer
+		if timeToRefresh <= 0 {
+			// If a new certificate expires before our buffer has expired,
+			// we should wait a bit and schedule a new refresh to much closer to the expiration's date
+			// TODO: This situation probably only occurs for Issue #622 - when the oauth2 token isn't refreshed before the cert is
```
enocom, line 276: Could we revert #680 with the code that follows?

proxy/proxy/client.go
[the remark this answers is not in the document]
enocom: I understand this code better now. Let me pose a different question below. Resolving.

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
+			timeToRefresh = certExpiration.Sub(now) - (5 * time.Second)
+			logging.Errorf("new ephemeral certificate expires sooner than expected (adjusting refresh time to compensate): current time: %v, certificate expires: %v", now, certExpiration)
+		}
+		logging.Infof("Scheduling refresh of ephemeral certifcate in %s", timeToRefresh)
+		go c.refreshCertAfter(instance, timeToRefresh)
```
enocom, line 281: Does this have to happen in a separate goroutine? I think we might be able to inline the function here and run it as part of the same goroutine. We'd need to close `done` on line 268 to do that.

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
 	return time.Now().After(cfg.Certificates[0].Leaf.NotAfter)
 }
 
-func (c *Client) cachedCfg(instance string) (string, *tls.Config, string) {
+// startRefresh kicks off a refreshCfg asynchronously, returning a channel that is closed
```
enocom, line 244: nit: returning a channel and updating the cacheEntry when it's complete.

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
+		}
+		c.cacheL.Unlock()
+	}
 
-	// Don't waste time returning an expired/invalid cert.
-	if !ok || ret.err != nil || isExpired(ret.cfg) {
-		return "", nil, ""
+	if !isValid(e) {
+		// if the previous result was invalid, wait for the next result to complete
+		// TODO: Ask for context?
```
enocom, line 347: What do you mean with this TODO?

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
-		}
-	}
-
-	addr, cfg, _, err := c.refreshCfg(instance)
+	addr, cfg, _, err := c.cachedCfg(instance)
 	if err != nil {
 		return nil, err
 	}
+
+	// TODO: attempt an early refresh if an connect fails?
```
enocom, line 366: I think we can hold off on this for now until we better understand how this code performs generally.

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
+	go func() {
+		defer close(done)
+		addr, cfg, ver, err := c.refreshCfg(instance)
+
+		c.cacheL.Lock()
+		old := c.cfgCache[instance]
+		// if we failed to refresh cfg do not throw out potentially valid one
+		if err != nil && !isExpired(old.cfg) {
+			logging.Errorf("failed to refresh the ephemeral certificate for %s, returning previous cert instead: %v", instance, err)
+			addr, cfg, ver, err = old.addr, old.cfg, old.version, old.err
```
enocom, line 258: In this case, do we know `old.err` to be nil? If so, it would be better to assign `err` to `nil` directly.

proxy/proxy/client.go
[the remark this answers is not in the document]
enocom: Actually, reading the link above, I see it says, "If your conditional looks like “if not … else …” then reverse it to put the positive case first.".

That makes me think we flip the method to `isInvalid`, or to stick with a positive phrasing, `isStale`, or `isErred`, etc. That way we write `if isStale` instead of requiring readers to parse `if !isValid`, which requires a little more mental effort to understand.

enocom (commented):
I've read through the original carefully and now this PR. It's looking pretty good. I've added some additional comments.

I think we're on a good path here and after resolving the comments, we should load test this before cutting a new release, just to shake out any additional issues.

proxy/proxy/client.go
enocom: If later calls to `tryConnect` fail, what stops this goroutine?

proxy/proxy/client.go
```diff
@@ -292,33 +241,129 @@ func isExpired(cfg *tls.Config) bool {
...
+		// the state may have changed between critical sections, so double check
+		e = c.cfgCache[instance]
+		if needsRefresh(e, refreshCfgBuffer) {
+			if time.Since(e.lastRefreshed) >= throttle {
+				// start a new refresh and update the cachedEntry to reflect that
+				e.done = c.startRefresh(instance, refreshCfgBuffer)
+				e.lastRefreshed = time.Now()
+				c.cfgCache[instance] = e
+			} else {
+				// TODO: Should this be a returned error?
```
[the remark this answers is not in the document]
kurtisvg: I think we'd only want to send a throttled error if the existing entry is no longer useable, otherwise we may throttle and return valid results but also a non-nil err.  

I think for right now I'm going to leave this as is, since it's preserving the existing behavior. I'll leave the TODO and we can determine in the future if we wish to pursue.

proxy/proxy/client.go
kurtisvg: What we discussed yesterday is the following two choices:

without #680:
```
55m - Refresh successfully but cert expires early - schedules a new one at 59:55s
59m:55s - Second refresh attempted. If it fails, cert will expire before next attempt. 
```
If successful, this doubles the sqladmin quota usage. 

With #680:
```
59m - Refresh. If failed, schedules a new one at 59:30s
```
Even if the first fails, we get 2 attempts (and a 3rd started) before the cert expires. Quota usage is as expected.

proxy/proxy/client.go
kurtisvg: > Does this have to happen in a separate goroutine? I think we might be able to inline the function here and run it as part of the same goroutine.

This is easier to leave in a separate goroutine because the other options are:
1. iterate recursively -> (likely) eventual stack overflow
2. iterate in a loop -> would require a significant rewrite so that it can be started or terminated. I'd prefer to keep the existing logic outside of the caching mostly intact for now. 

> If later calls to tryConnect fail, what stops this goroutine?

The current logic is (simplified):
1. A happens to `cachedCfg`
2. If we need a new refresh, call `startRefresh`
3. `startRefresh` starts goroutine, performs a refresh, and schedules a call to `cachedCfg` 55 minutes in the future
4. `cachedCfg` is called again

The goroutine will terminate when `cachedCfg` is called. If it's called at the right time, it kicks off a new refresh. If it's called at the wrong time the next call to `cachedCfg` will determine if it needs to be started.  .

proxy/proxy/client.go
kurtisvg: Not sure what this means. Can you clarify?

proxy/proxy/client.go
kurtisvg: We don't know if `old.err` is `nil`. It doesn't really matter - if our current result failed, the old result is either good or bad. It's at least better than the current result.

proxy/proxy/client.go
enocom: :+1:

proxy/proxy/client.go
enocom: Just updating the doc string to read a little better.

enocom (approved):
My comments are mostly stylistic and aren't blocking.

proxy/proxy/client.go
kurtisvg: Update cachedCfg to use a context arg (which I've just done).

proxy/proxy/client.go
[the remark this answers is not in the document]
kurtisvg: We're not using an `if not … else …`, we're just using an "if else", so that doesn't apply here. If we were, it would be better to put the `isValid` case first and then `!isValid` case second. 

The problem is that the "In" from "Invalid" is a negation, and if you ever need to inverse a negation you end up having a double negative to accomplish that (`!IsInvalid`). 

IMO, the summary wraps it up quite nicely:
>  in general checking the presence or absence of a positive is simpler for readers to understand than checking the presence or absence of a negative.

proxy/proxy/client.go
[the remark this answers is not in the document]
kurtisvg: Will play it safe and leave it for now - I don't think it's a significant performance difference.

enocom (approved):
LGTM.

Let's talk about l294 (i.e., `go c.refreshCartAfter`) offline.

PR #678, opened 2021-03-10
chore: update k8s readme to include workload identity for node pool

### LINKED ISSUE #675: Kubernetes README doesn't mention workload identity for node pool
## Description

The [Kubernetes example](https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/main/examples/kubernetes#workload-identity) advises users to enable workload identity on their cluster, but is silent on enabling workload identity on the cluster's node pool (a separate step). 

## Solution

The README linked above should add a note about enabling workload identity on both the cluster and the node pool.

shubha-rajan:
fixes #675


enocom (approved):
Thanks for grabbing this!

PR #680, opened 2021-03-11
fix: lower refresh buffer and config throttle when IAM authn is enabled

### LINKED ISSUE #662: Enable IAM Authn fails to refresh certificate when using ADC retrieved from a metadata server

enocom:
With IAM login enabled, the proxy would fail to refresh its certificate
when using Application Default Credentials retrieved from the metadata
server. The metadata server returns an access token that is refreshed
only when the expiration is within ~60 seconds. As a result, when the
proxy attempted a refresh of the certificate, it would fail because the
token did not have enough time left on it relative to the default buffer
(i.e., 5 minutes).

This commit configures the buffer to 55 seconds when IAM login is
enabled. When a token has ~60 seconds before expiration, the refresh
logic will schedule a refresh 5 seconds in the future. When 5 seconds
have elapsed, the refresh logic will get a new token whose expiration is
in an hour and restart the clock on the next refresh.

Fixes #662.

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -32,7 +32,16 @@ import (
 const (
 	DefaultRefreshCfgThrottle = time.Minute
 	keepAlivePeriod           = time.Minute
-	defaultRefreshCfgBuffer   = 5 * time.Minute
+	// DefaultRefreshCfgBuffer is the minimum amount of time for which a
+	// certificate must be valid to ensure the next refresh attempt has adequate
+	// time to complete.
+	DefaultRefreshCfgBuffer = 5 * time.Minute
+	// IAMLoginRefreshCfgBuffer is the minimum amount of a time for which a
+	// certificate holding an Access Token must be valid. Because some token
+	// sources (e.g., ouath2.ComputeTokenSource) are refreshed with only ~60
+	// seconds before expiration, this value must be smaller than the
+	// DefaultRefreshCfgBuffer.
+	IAMLoginRefreshCfgBuffer = 55 * time.Second
 )
 
 var (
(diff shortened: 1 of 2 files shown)


proxy/proxy/client.go
```diff
@@ -172,7 +182,7 @@ func (c *Client) refreshCfg(instance string) (addr string, cfg *tls.Config, vers
 
 	refreshCfgBuffer := c.RefreshCfgBuffer
 	if refreshCfgBuffer == 0 {
-		refreshCfgBuffer = defaultRefreshCfgBuffer
+		refreshCfgBuffer = DefaultRefreshCfgBuffer
```
kurtisvg, line 185: Since the throttle is at 1 minute, this essentially means that if it doesn't successfully refresh on this attempt, the next refresh will be throttled until the certs expiration. Should we consider lowering the throttle as well? 

For the Java connector, we rate limit at 2 attempts per minute. Maybe we should just match that? WDYT?

enocom: Matching the Java connector seems like a good idea. I'll update this PR to do that.

enocom: Fixed in the latest commit.

kurtisvg (approved):
One lingering thought before we discuss this: This has potentially significant degradation on the chance of a successful refresh during a partial control plane outage, and the changes only apply to one credential path (ADC -> Workload Identity).

We should make sure we're okay with the trade off here before we merge.

enocom:
To summarize a conversation elsewhere, we recognize that a shorter buffer for IAM auth (i.e., 55 seconds) means in the event of control plane problems, there is a higher chance that refresh attempts will fail. I've filed an internal bug to explore options for getting a token with more time on it before expiration. In the mean time though, this PR resolves #662 and at least allows clients to use the `enable_iam_login` with no issue when the control plane isn't in a degraded state.

Issue #290, opened 2019-06-21
Feature Request: Better support for proxy as daemon


dhduvall:
This is related to #4, but I don't think it's possible to run `cloud-sql-proxy` as a service other services can depend on because  it doesn't do its own daemonization. Using the `daemon` utility would still allow the service to be online (as far as `systemd` is concerned, so it'll start dependent services) before `cloud-sql-proxy` is actually listening on its socket, because the external utility can't know when that point occurs.

I know that `systemd` isn't universal, but `github.com/coreos/go-systemd/daemon.SdNotify()` doesn't do anything if `$NOTIFY_SOCKET` isn't set, and the package produces only a 16k archive (x86_64 Linux), so it's a small weight to carry around for a one-time operation even on systems where it's not relevant.

pperez:
Ah, exactly what i was looking for ... seems like i'm going to add a little sleep in my script :/

enocom:
Duplicate of #234.

dhduvall:
I'd say that #234 is not strictly a duplicate of this one, since this issue is specifically about making sure that `systemd` doesn't mark the proxy service as online until the proxy is ready to take requests, and #234 introduces the requirement that the proxy be self-daemonizing.

Without diving too deeply into the possible implementation, I admit that it may be possible to daemonize in such a way that the grandparent process doesn't exit until the grandchild tells it that its ready to handle requests, but the communication between them needed to do so isn't nearly as simple as the naive implementation (such as using the `daemon()` library call), where the grandparent exits as soon as the `fork()` is successful (and thus telling _its_ parent that the daemon is ready).

enocom:
Thanks for the note @dhduvall. 

For now, I'm going to add your note about making sure the proxy works well with `systemd` in #234 and consider that issue over there to be a tracking issue generally for this work.

PR #719, opened 2021-04-05
feat: Add support for systemd notify

### LINKED ISSUE #234: Better support for running as a daemon process
I'm trying to run cloud_sql_proxy in a Google Colab notebook. We can authenticate and the proxy is actually running.
```
!./cloud_sql_proxy --instances=my-project:my-region:my-instance
```

However the notebook will not let me execute other cells, as the cloud_sql_proxy command does not terminate. I would hope that if we send the task to background, the notebook will continue processing.

Note that adding a `&` to run in the background doesn't seem to work in the notebook. The process continues to run foreground.

Please see https://stackoverflow.com/a/49142801 for a complete example.

carlpett:
## Change Description

This PR adds support for running with `Type=notify` under systemd. The benefit of this is that service startup is not marked as successful until the signal has been sent, meaning other services with appropriate `Requires` and `After` directives won't try to start until the proxy is ready to handle connections

## Checklist

- [x] Make sure to open an issue as a  [bug/issue](https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/new/choose)   before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea. (Kinda? There already was one...)

## Relevant issues:

- Related to #234


cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -614,12 +615,23 @@ func main() {
 
 	logging.Infof("Ready for new connections")
 
+	go func() {
```
carlpett, line 618: I debated this a bit, but ended up running the notifies in separate goroutines to guard against any potential weirdness. Would gladly have a discussion on if this is a good idea or not.
The call dials a unix socket (but only if `NOTIFY_SOCKET` is set), and I'm wary that under the right set of unfortunate circumstances that could hang. But I don't know exactly what those circumstances would be (maybe if systemd itself is hanging)

carlpett:
I'm not sure why the CLA doesn't work. I'm fairly sure I've signed it previously, but now it appears to try to match it up against my corporate identity when I open the link? (Not US-based, and my employer does not have rights over my spare time contributions)

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -41,6 +41,7 @@ import (
 	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/util"
 
 	"cloud.google.com/go/compute/metadata"
+	"github.com/coreos/go-systemd/v22/daemon"
```
kurtisvg, line 44: How big of a dependency is this? Do you know what the change in binary size is?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
kurtisvg: Do you have any idea what the impact on systems without systemd might be? 

Is this actually ready at this point in time? Or does it need to wait for `proxyCleint.Run()` to be called below.

kurtisvg: Looking at the docs, looks like this only kicks in if `$NOTIFY_SOCKET` is set (which you referred to above). So this shouldn't have any effect on systems without systemd present? Have you done any testing?

kurtisvg (changes requested):
We may want to consider putting it behind a flag as well - user's without systemd might be

cmd/cloud_sql_proxy/cloud_sql_proxy.go
carlpett: Pretty small:
```shell
$ ls -l cloud_sql_proxy*
-rwxr-xr-x. 1 cape cape 16481993  5 apr 20.56 cloud_sql_proxy*
-rwxr-xr-x. 1 cape cape 16494725  5 apr 20.26 cloud_sql_proxy-with-systemd*
```

cmd/cloud_sql_proxy/cloud_sql_proxy.go
carlpett: Yes, exactly, the code bails early if the environment variable isn't set. So non-systemd users will not get any change in behaviour unless they set that themselves.

>Is this actually ready at this point in time?

Probably no, you are very correct that it should be holding until after `proxyClient.Run`, at least. In my testing the connection set up fast enough that I didn't see that the log statement above my addition was a bit premature. 

But yes, I've done testing, and apart from that embarrassing part, the happy cases work well. I'm not sure what the less happy ones would be, though. From digging a bit, it looks like even `proxyClient.Run` might not be entirely correct, since it just kicks off the connections starting?

kurtisvg:
@carlpett also RE: CLA bot, it uses the commit email. Is that address correct for the CLA?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
kurtisvg: I think we're actually okay, because we start receiving connections with the `connSrc` is initialized, even if they aren't handled until `Run()` is called. However it's probably logically better to move this down closer to `.Run()` and add a comment, to help minimize the chance of a potentially long running bit sneaking in-between the two sections.

carlpett:
Yes, should be. But signing in with my private account if doesn't seem to find the old CLA, so :shrug: Re-signed it now.
@googlebot I signed it!

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -629,5 +635,13 @@ func main() {
...
 	}()
 
+	// If running under systemd with Type=notify, we'll send a message to the
+	// service manager that we are ready to handle connections now, and any other
+	// units that are waiting for us can start.
+	go func() {
+		if _, err := daemon.SdNotify(false, daemon.SdNotifyReady); err != nil {
+			logging.Errorf("Failed to notify systemd of readiness: %v", err)
+		}
+	}()
```
carlpett, line 645: @kurtisvg Moved it down here. Anything else that should be mentioned in the comment to avoid the risks you mentioned around code evolution?

kurtisvg (approved):
LGTM. We are planning a release tomorrow, but we're gonna wait until it's complete before merging this in. 

Also @enocom is gonna run a couple of smoke tests on:
- [ ] linux w/ systemd
- [ ] windows (w/ systemd)

kurtisvg:
Also, this would have fixed #290, but I'm not sure it's addressing #234 as a whole. We should follow up on that before closing it.

enocom:
Sorry for the slow response. I'll be double-checking this today.

enocom:
I ran the following test to confirm that this PR works with systemd where `Type=notify` on Linux.

First I used the following unit file:

```
eno@scratch:~/cloudsql-proxy$ cat /etc/systemd/system/cloud-sql-proxy.service 
[Unit]
Description=GCP CloudSQL Proxy
After=network.target

[Service]
Type=notify
User=eno
Restart=always
ExecStart=/home/eno/cloudsql-proxy/cloud_sql_proxy -instances=PROJECT_ID_OMITTED:us-central1:postgres-instance=tcp:5432

[Install]
WantedBy=multi-user.target
````

And then I build the proxy from this PR and confirmed the service started:

```
eno@scratch:~$ sudo systemctl status cloud-sql-proxy
● cloud-sql-proxy.service - GCP CloudSQL Proxy
   Loaded: loaded (/etc/systemd/system/cloud-sql-proxy.service; disabled; vendor preset: enabled)
   Active: active (running) since Thu 2021-04-15 17:33:28 UTC; 5min ago
 Main PID: 5273 (cloud_sql_proxy)
    Tasks: 7 (limit: 4665)
   Memory: 13.2M
   CGroup: /system.slice/cloud-sql-proxy.service
           └─5273 /home/eno/cloudsql-proxy/cloud_sql_proxy -instances=PROJECT_ID_OMITTED:us-central1:postgres-instance=tcp:5432

Apr 15 17:33:17 scratch systemd[1]: Starting GCP CloudSQL Proxy...
Apr 15 17:33:17 scratch cloud_sql_proxy[5273]: 2021/04/15 17:33:17 Rlimits for file descriptors set to {Current = 8500, Max = 524288}
Apr 15 17:33:18 scratch cloud_sql_proxy[5273]: 2021/04/15 17:33:18 Listening on 127.0.0.1:5432 for PROJECT_ID_OMITTED:us:us-central1:postgres-instance
Apr 15 17:33:18 scratch cloud_sql_proxy[5273]: 2021/04/15 17:33:18 Ready for new connections
Apr 15 17:33:28 scratch systemd[1]: Started GCP CloudSQL Proxy.
~
```

Next, I rebuilt the binary, omitted the call to `daemon.Sdnotify`, and confirmed the service failed to start.

enocom:
I tried this change out on a Window machine as well.

Here's what I did:
1. Built the proxy from this PR
2. Ran it
3. Made sure I could connect to my instance as usual

All works as expected.

enocom:
As for #234, I think this PR won't entirely fix the issue and so we should keep it open after merging this.

enocom:
Thanks @carlpett for this PR. Looks great.

PR #726, opened 2021-04-12
fix: Allow combined use of structured logs and -log_debug_stdout

### LINKED ISSUE #725: Structured logging always logs to stderr
## Bug Description

Now as structured logging has been released, I've tried to use it in combination with the `-log_debug_stdout` flag, so that all log items are structured, and only errors are written to `stderr`. As it turned out, all logs are written to `stderr` if `-structured_logs` is turned on.

## Example command
This command should leave an empty terminal, since `stdout` logs are redirected to `/dev/null`, however `"level":"info"` logs appear in the terminal from `stderr`. 
```
$ ./cloud_sql_proxy.linux.amd64 -verbose=false -credential_file=./secret.json -log_debug_stdout=true -structured_logs=true >/dev/null

{"level":"info","ts":1618224828.160324,"caller":"cloud_sql_proxy/cloud_sql_proxy.go:615","msg":"Ready for new connections"}
```
Also if I pipe `stderr` to `/dev/null` using `2>/dev/null` I get an empty terminal.

## Proposal

Structured logging uses [zap](https://pkg.go.dev/go.uber.org/zap):
```
logger, err := zap.NewProduction()
```
The `NewProduction` uses `stderr` for all logs, as it is declared in its [source](https://github.com/uber-go/zap/blob/404189cf44aea95b0cd9bddcb0242dd4cf88c510/config.go#L120).

I think setting the `OutputPaths` to `[]string{"stdout"}` when `-log_debug_stdout` is set could solve the issue.

## Environment

1. OS type and version: Ubuntu 5.8.0-48-generic x86_64 GNU/Linux
2. Cloud SQL Proxy version (`./cloud_sql_proxy -version`): Cloud SQL Auth proxy: 1.21.0+linux.amd64

kundralaci:
## Change Description

Fix the combined use of structured logs (`-structured_logs`) and standard output debug logs (`-log_debug_stdout`).

## Relevant issues:

- Fixes #725

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -57,8 +57,12 @@ func DisableLogging() {
 
 // EnableStructuredLogs replaces all logging functions with structured logging
 // variants.
-func EnableStructuredLogs() (func(), error) {
-	logger, err := zap.NewProduction()
+func EnableStructuredLogs(logDebugStdout bool) (func(), error) {
+	config := zap.NewProductionConfig()
+	if logDebugStdout {
+		config.OutputPaths = []string{"stdout"}
+	}
+	logger, err := config.Build()
 	if err != nil {
 		return func() {}, err
 	}
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -58,7 +58,7 @@ func DisableLogging() {
 
 // EnableStructuredLogs replaces all logging functions with structured logging
 // variants.
-func EnableStructuredLogs() (func(), error) {
+func EnableStructuredLogs(logDebugStdout bool) (func(), error) {
 	// Define level-handling logic.
 	highPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
 		return lvl >= zapcore.ErrorLevel
@@ -67,9 +67,15 @@ func EnableStructuredLogs() (func(), error) {
 		return lvl < zapcore.ErrorLevel
 	})
 
-	// Lock console
-	consoleDebugging := zapcore.Lock(os.Stdout)
+	// Lock consoles
 	consoleErrors := zapcore.Lock(os.Stderr)
+
+	var consoleDebugging zapcore.WriteSyncer
+	if logDebugStdout {
+		consoleDebugging = zapcore.Lock(os.Stdout)
+	} else {
+		consoleDebugging = consoleErrors
+	}
 
 	// Create JSON encoder using opinionated production preset
 	consoleEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -22,6 +22,7 @@ import (
 	"os"
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 // Verbosef is called to write verbose logs, such as when a new connection is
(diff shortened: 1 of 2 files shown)


enocom (changes requested):
See my comment [on the issue](https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/725#issuecomment-817971029). Happy to work through this with you, and will otherwise pick this work up myself.

logging/logging.go
```diff
@@ -58,10 +59,30 @@ func DisableLogging() {
...
+	// Define level-handling logic.
+	highPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl >= zapcore.ErrorLevel
+	})
+	lowPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl < zapcore.ErrorLevel
+	})
+
+	// Lock console
+	consoleDebugging := zapcore.Lock(os.Stdout)
```
enocom, line 71: We'll still want to send all logs to stderr unless a user has enabled the `log_debug_stdout` flag. The approach you had earlier passing the bool into `EnableStructuredLogs` still makes sense here. If the `log_debug_stdout` flag is set, we would use stdout on this line. Otherwise, it would be stderr.

enocom (changes requested):
Looking good. Now we just need to preserve existing stderr logging.

logging/logging.go
kundralaci: True, I've just fixed it.

logging/logging.go
```diff
@@ -57,11 +58,37 @@ func DisableLogging() {
...
+		return lvl < zapcore.ErrorLevel
+	})
+
+	// Lock consoles
+	consoleErrors := zapcore.Lock(os.Stderr)
+
+	var consoleDebugging zapcore.WriteSyncer
+	if logDebugStdout {
+		consoleDebugging = zapcore.Lock(os.Stdout)
+	} else {
```
enocom, line 76: We can simplify this statement and avoid the `else` clause to keep the normal code path at a minimal indentation.

``` go
consoleErrors := zapcore.Lock(os.Stderr)
consoleDebugging := consoleErrors
if logDebugStdout {
  consoleDebugging = zapcore.Lock(os.Stdout)
}
```

See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

enocom (changes requested):
One small nit and then this is ready to go. Thanks again for this!

logging/logging.go
```diff
@@ -57,11 +58,37 @@ func DisableLogging() {
...
+	consoleEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+
+	// Join the outputs, encoders, and level-handling functions into
+	// zapcore.Cores, then tee them together.
+	core := zapcore.NewTee(
+		zapcore.NewCore(consoleEncoder, consoleErrors, highPriority),
+		zapcore.NewCore(consoleEncoder, consoleDebugging, lowPriority),
+	)
+
+	// Create logger, also add caller and stacktrace
```
kurtisvg, line 90: nit: Same for these comments. They can be either removed or updated to add context on why we are doing the specific line.

kurtisvg (changes requested):

logging/logging.go
kundralaci: Thank you for pointing this out, I rarely had a chance to code in go lately.

logging/logging.go
```diff
@@ -57,11 +58,37 @@ func DisableLogging() {
...
+func EnableStructuredLogs(logDebugStdout bool) (func(), error) {
+	// Define level-handling logic.
+	highPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl >= zapcore.ErrorLevel
+	})
+	lowPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl < zapcore.ErrorLevel
+	})
+
+	// Lock consoles
```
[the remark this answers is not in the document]
kundralaci: Alright. As I'm neither quite familiar with zap, I've just removed the redundant comments and added a link in the comments to the official example on which this is based.

logging/logging.go
kundralaci: Added a link in the comments to the official example on which this is based.

logging/logging.go
```diff
@@ -57,15 +58,40 @@ func DisableLogging() {
 
 // EnableStructuredLogs replaces all logging functions with structured logging
 // variants.
-func EnableStructuredLogs() (func(), error) {
-	logger, err := zap.NewProduction()
-	if err != nil {
-		return func() {}, err
+func EnableStructuredLogs(logDebugStdout bool) (func(), error) {
+	// Configuration of zap is based on its Adavnced Configuration example,
```
enocom, line 62: s/Adavnced/Advanced/

enocom: s/example,/example./

logging/logging.go
```diff
@@ -57,15 +58,40 @@ func DisableLogging() {
...
+	// Define level-handling logic.
+	highPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl >= zapcore.ErrorLevel
+	})
+	lowPriority := zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+		return lvl < zapcore.ErrorLevel
+	})
+
+	// Lock wraps a WriteSyncer in a mutex to make it safe for concurrent use. In
+	// particular, *os.Files must be locked before use.
```
enocom, line 74: *os.File types

logging/logging.go
```diff
@@ -57,15 +58,40 @@ func DisableLogging() {
...
+		consoleDebugging = zapcore.Lock(os.Stdout)
 	}
+
+	consoleEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+	core := zapcore.NewTee(
+		zapcore.NewCore(consoleEncoder, consoleErrors, highPriority),
+		zapcore.NewCore(consoleEncoder, consoleDebugging, lowPriority),
+	)
+
+	// By deafult, caller and stacktrace are not included, so add them here
```
enocom, line 87: s/deafult/default/

enocom (commented):
Just a few typos and then I'm happy to approve. Thanks for this!

logging/logging.go
kundralaci: Sorry for that, my bad.

enocom (approved):
Thanks for the contribution @kundralaci!

kurtisvg (approved):

logging/logging.go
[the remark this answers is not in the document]
kurtisvg: Looks great, thank you!

PR #748, opened 2021-04-20
fix: return early when cert refresh fails

### LINKED ISSUE #745: SIGSEGV when first refresh operation fails
## Bug Description

CloudSQL Proxy throws SIGSEGV once the first MySQL client connection is made and the proxy is started with "-ip_address_types=PRIVATE" and the CloudSQL instance does not have a private IP.  

```
./cloud_sql_proxy -instances=PROJECT_ID:us-central1:XYZ=tcp:3306 -verbose -ip_address_types=PRIVATE

2021/04/20 12:18:11 Rlimits for file descriptors set to {Current = 8500, Max = 9223372036854775807}
2021/04/20 12:18:11 using credential file for authentication; path="/Users/……./cred.json"
2021/04/20 12:18:12 Listening on 127.0.0.1:3306 for PROJECT_ID:us-central1:XYZ
2021/04/20 12:18:12 Ready for new connections
2021/04/20 12:18:22 New connection for "PROJECT_ID:us-central1:XYZ"
2021/04/20 12:18:22 refreshing ephemeral certificate for instance PROJECT_ID:us-central1:XYZ
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x155a56f]

goroutine 83 [running]:
github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy.(*Client).startRefresh.func1(0xc00032a540, 0xc0002e6180, 0x7ffeefbff5ad, 0x48, 0x45d964b800)
	/workspace/proxy/proxy/client.go:285 +0x38f
created by github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy.(*Client).startRefresh
	/workspace/proxy/proxy/client.go:264 +0x85


(venv) [user@machine:~/software] $ cloud_sql_proxy -version
Cloud SQL Auth proxy: 1.21.0+darwin.amd64
```

## How to reproduce (see above)

  1. Download and run proxy version 1.21.0+darwin.amd64
  2. Try to connect to DB without private IP via proxy (using private IP flag), resulting in signal SIGSEGV: segmentation violation

## Environment

1. OS type and version: <YOUR_OS_HERE>
2. Cloud SQL Proxy version (`./cloud_sql_proxy -version`): Cloud SQL Auth proxy: 1.21.0+darwin.amd64

enocom:
Prior to this commit, the client assumed all attempts to refresh a
certificate would succeeed and therefore panic when an attempt failed.
This commit ensures the client returns early when errors occur.

Fixes #745

### DIFF
--- a/proxy/proxy/client.go
+++ b/proxy/proxy/client.go
@@ -272,7 +272,7 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
 			logging.Errorf("failed to refresh the ephemeral certificate for %s, returning previous cert instead: %v", instance, err)
 			addr, cfg, ver, err = old.addr, old.cfg, old.version, old.err
 		}
-		c.cfgCache[instance] = cacheEntry{
+		e := cacheEntry{
 			lastRefreshed: time.Now(),
 			err:           err,
 			addr:          addr,
@@ -280,7 +280,15 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
 			cfg:           cfg,
 			done:          done,
 		}
+		c.cfgCache[instance] = e
 		c.cacheL.Unlock()
+
+		if !isValid(e) {
+			// Note: Future refreshes will not be scheduled unless another
+			// connection attempt is made.
+			logging.Errorf("failed to refresh the ephemeral certificate for %v: %v", instance, err)
+			return
+		}
 
 		certExpiration := cfg.Certificates[0].Leaf.NotAfter
 		now := time.Now()
--- a/proxy/proxy/client_test.go
+++ b/proxy/proxy/client_test.go
@@ -31,8 +31,8 @@ import (
 const instance = "instance-name"
 
 var (
-	errFakeDial = errors.New("this error is returned by the dialer")
-	forever     = time.Date(9999, 0, 0, 0, 0, 0, 0, time.UTC)
+	sentinelError = errors.New("sentinel error")
+	forever       = time.Date(9999, 0, 0, 0, 0, 0, 0, time.UTC)
 )
 
 type fakeCerts struct {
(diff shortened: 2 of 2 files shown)


proxy/proxy/client.go
```diff
@@ -272,22 +274,32 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
...
 		timeToRefresh := certExpiration.Sub(now) - refreshCfgBuffer
 		if timeToRefresh <= 0 {
-			// If a new certificate expires before our buffer has expired, we should wait a bit and schedule a new refresh to much closer to the expiration's date
-			// This situation probably only occurs when the oauth2 token isn't refreshed before the cert is, so by scheduling closer to the expiration we can hope the oauth2 token is newer.
+			// If a new certificate expires before our buffer has expired, we
+			// should wait a bit and schedule a new refresh to much closer to
+			// the expiration's date.
+			// This situation probably only occurs when the oauth2 token isn't
+			// refreshed before the cert is, so by scheduling closer to the
+			// expiration we can hope the oauth2 token is newer.
```
kurtisvg, line 302: nit: Not sure this comment needed to be changed

proxy/proxy/client.go
```diff
@@ -257,8 +257,10 @@ func isExpired(cfg *tls.Config) bool {
 	return time.Now().After(cfg.Certificates[0].Leaf.NotAfter)
 }
 
-// startRefresh kicks off a refreshCfg asynchronously, that updates the cacheEntry and closes the returned channel once the refresh is completed. This function
-// should only be called from the scope of "cachedCfg", which controls the logic around throttling refreshes.
+// startRefresh kicks off a refreshCfg asynchronously, that updates the
+// cacheEntry and closes the returned channel once the refresh is completed.
+// This function should only be called from the scope of "cachedCfg", which
```
kurtisvg, line 262: nit: not sure this needs to be changed

proxy/proxy/client.go
```diff
@@ -272,22 +274,32 @@ func (c *Client) startRefresh(instance string, refreshCfgBuffer time.Duration) c
...
 			version:       ver,
 			cfg:           cfg,
 			done:          done,
 		}
+		c.cfgCache[instance] = e
 		c.cacheL.Unlock()
 
+		if !isValid(e) {
+			logging.Errorf("failed to refresh the ephemeral certificate for %v: %v", instance, err)
+			return
```
kurtisvg, line 290: We should probably add a note here about how this means future attempts won't be rescheduled unless another connection attempt is made

kurtisvg (approved):
In the future, let's try to keep more general code health improvements in a separate PRs.

enocom:
I've reverted the comment line wrapping and will keep future code health changes in separate PRs.

kurtisvg (approved):

PR #752, opened 2021-04-21
chore: use version.txt in build

enocom:
This commit removes the hard-coded version string in cloud_sql_proxy.go
and instead pulls it with a linker flag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,9 @@ WORKDIR /go/src/cloudsql-proxy
 COPY . .
 
 RUN go get ./...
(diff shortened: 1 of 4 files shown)


enocom:
It seems this approach is what we're already using for the [GCS uploads](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/main/.build/gcs_upload.yaml#L30). Provided I haven't overlooked something, this commit will make our release process a bit easier.

shubha-rajan (approved):

kurtisvg (changes requested):
Let's talk about this before merging

shubha-rajan:
Looks good! you can probably get rid of the step that checks for version mismatches in the lint GitHub action, since now we're using version.txt as the single source of truth.

PR #766, opened 2021-04-28
chore: add note about maximum socket length

### LINKED ISSUE #747: unix-socket maximal path length exceeded when cloudsql-proxy tries to open a unix socket in the directory specified by `-dir` parameter
## Description
When we start cloudsql-proxy using unix sockets using this command 
```
./cloud_sql_proxy -dir=/cloudsql -instances=INSTANCE_CONNECTION_NAME
-credential_file=PATH_TO_KEY_FILE &
```

cloudsql-proxy opens a unix socket in the path `/cloudsql/INSTANCE_CONNECTION_NAME`. However unix-socket paths have a path lenght limit dependent on the platform it runs. If the unix socket path length exceeds the limit cloudsql-proxy throws the following error :- 
`bind: invalid argument`.

While this could be solved by overriding the `-dir` parameter and provinding an absolute path in the `-instance` parameter, the issue of unix socket path limit should be documented somewhere, so that the users are aware of this.

enocom:
See https://unix.stackexchange.com/a/367012/26069 for details.

Fixes #747.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,9 @@ cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:0.0.0.0:5432
 
 ``` bash
 # The proxy will mount a Unix domain socket at /cloudsql/<INSTANCE_CONNECTION_NAME>
-# Note: The directory specified by `-dir` must exist.
+# Note: The directory specified by `-dir` must exist and the socket file path
+# (i.e., dir plus INSTANCE_CONNECTION_NAME) must be under 108 characters on
+# Linux (or sometimes even less depending on the platform).
 cloud_sql_proxy -dir=/cloudsql -instances=<INSTANCE_CONNECTION_NAME>
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -92,7 +92,9 @@ cloud_sql_proxy -instances=<INSTANCE_CONNECTION_NAME>=tcp:0.0.0.0:5432
 
 ``` bash
 # The proxy will mount a Unix domain socket at /cloudsql/<INSTANCE_CONNECTION_NAME>
-# Note: The directory specified by `-dir` must exist.
+# Note: The directory specified by `-dir` must exist and the socket file path
+# (i.e., dir plus INSTANCE_CONNECTION_NAME) must be under 108 characters on
+# Linux (or sometimes even less depending on the platform).
```
kurtisvg, line 97:
```suggestion
# Note: The directory specified by `-dir` must exist and the socket file path
# (i.e., dir plus INSTANCE_CONNECTION_NAME) must be under your platform's
# limit (typically 108 characters on many Unix systems, but varies by platform) 
```

kurtisvg (approved):
Slight formatting nit, otherwise LGTM

shubha-rajan (approved):

PR #767, opened 2021-04-28
chore(ci): run tests on Windows

### LINKED ISSUE #392: Add continuous testing on Windows

shubha-rajan:
## Change Description

Please provide a detailed description on what changes your PR will have.

## Checklist

## Relevant issues:

- Fixes #392


cmd/cloud_sql_proxy/proxy_test.go
```diff
@@ -98,6 +99,12 @@ func TestCreateInstanceConfigs(t *testing.T) {
 			continue
 		}
 		if err != nil {
+			if runtime.GOOS == "windows" {
+				if err.Error() == "unsupported network: unix" || err.Error() == "FUSE not supported on this system" {
+					t.Logf("CreateInstanceConfigs got error when %s on Windows: %v", v.desc, err)
+					return // these errors are expected on Windows
```
enocom, line 105: Should this be `continue` instead of `return`?

.kokoro/tests/run_tests_windows.sh
```diff
@@ -0,0 +1,34 @@
...
+  # Move into project directory
+  cd github/cloud-sql-proxy
+  # source secrets
+  source "${KOKORO_GFILE_DIR}/TEST_SECRETS.sh"
+  export GOOGLE_APPLICATION_CREDENTIALS="${KOKORO_GFILE_DIR}/testing-service-account.json"
+fi
+
+echo -e "******************** Running tests... ********************\n"
+go test -v ./...
+echo -e "******************** Tests complete.  ********************\n"
```
enocom, line 34: nit: missing newline

cmd/cloud_sql_proxy/proxy_test.go
```diff
@@ -98,6 +99,12 @@ func TestCreateInstanceConfigs(t *testing.T) {
 			continue
 		}
 		if err != nil {
+			if runtime.GOOS == "windows" {
```
enocom, line 102: Granted this test is already is bad shape with all the boolean values, but would it be better to explicitly mark each test as `supportedOnWindows` such that we don't have to check error strings? That way it would be easy to write:

``` go
if runtime.GOOS == "windows" && !v.supportedOnWindows {
  t.Skip("skipping test for Unix sockets or FUSE")
  continue
}
```

shubha-rajan: That would be much cleaner. Should we do this with the other test too, or is checking the network value fine?

kurtisvg: We can also use the `!build windows` compile tags to skip tests on Windows

shubha-rajan: That's what's being done for the fuse and limits tests. The problem here is that we don't want to skip all of the tests in the module. Just some of the inputs for the table tests

kurtisvg: Ah understood - we can also consider just refactoring those to different files. I'll defer to Eno on what makes the most sense organizationally.

enocom: It would be nice to clean up this test such that we could use build tags. For now, we'll at least have Windows-tests explicitly marked so that when we (read: me) refactor the tests, it's an easy job to move things out as needed.

cmd/cloud_sql_proxy/proxy_test.go
```diff
@@ -47,49 +48,55 @@ func TestCreateInstanceConfigs(t *testing.T) {
...
-			"dir", false, nil, "md", false, false,
+			"dir", false, nil, "md", false, false, false,
 		}, {
 			"setting -instances_metadata",
-			"", false, nil, "md", true, false,
+			"", false, nil, "md", true, false, true,
 		},
 	} {
+		if runtime.GOOS == "windows" && !v.supportedOnWindows {
+			t.Skip("skipping test for Unix sockets or FUSE")
```
enocom, line 97: I missed that this entire test isn't wrapped in a subtest like we have below with `t.Run(...)`. So this line actually skips any remaining tests in the table. Probably the simplest thing to do here is just remove this line and keep the `continue` below.

shubha-rajan: Fixed. I missed it too.
Probably because two tests in the same file shouldn't be formatted differently for what seems like no reason.

enocom (approved):
:tada: Windows tests!

cmd/cloud_sql_proxy/proxy_test.go
enocom: Haha. Yep. We have some cleanup to do in future passes (on my list).

PR #774, opened 2021-05-03
chore: include example resource requests

### LINKED ISSUE #103: Update k8s examples to include example resource requests

enocom:
Fixes #103.

### DIFF
--- a/examples/kubernetes/proxy_with_sa_key.yaml
+++ b/examples/kubernetes/proxy_with_sa_key.yaml
@@ -70,7 +70,7 @@ spec:
             memory: "2Gi"
             # More database IO require more CPUs. Less database IO require less
             # CPU. Adjust this value based on your application's requirements.
-            cpu:    "2"
+            cpu:    "1"
       # [START cloud_sql_proxy_k8s_volume_secret]
       volumes:
       - name: <YOUR-SA-SECRET-VOLUME>
--- a/examples/kubernetes/proxy_with_sa_key.yaml
+++ b/examples/kubernetes/proxy_with_sa_key.yaml
@@ -67,7 +67,7 @@ spec:
             # this value based on your application's requirements.
             memory: "2Gi"
             # The proxy's CPU use scales linearly with the amount of IO between
-            # the database and the application. Adjust this value values on your
+            # the database and the application. Adjust this value based on your
             # application's requirements.
             cpu:    "1"
       # [START cloud_sql_proxy_k8s_volume_secret]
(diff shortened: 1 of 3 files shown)


examples/kubernetes/proxy_with_sa_key.yaml
```diff
@@ -57,6 +57,20 @@ spec:
           mountPath: /secrets/
           readOnly: true
           # [END cloud_sql_proxy_k8s_volume_mount]
+        # Resource configuration depends on an application's requirements. The
+        # number of connections is correlated to memory use. The amount of IO
+        # across each connection is correlated to CPU use. Users should adjust
+        # the following values based on what their application needs.
+        # For details, see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```
kurtisvg, line 64: 1. Let's avoid stating the same thing twice
2. "Users" -> "You" to match the rest of the comments

```suggestion
        # Resource configuration depends on an application's requirements.  You
        # should adjust the following values based on what their application needs.
        # For details, see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```

examples/kubernetes/proxy_with_sa_key.yaml
```diff
@@ -57,6 +57,20 @@ spec:
...
+        resources:
+          requests:
+            # More connections require more memory. Fewer connections require
```
kurtisvg, line 67: Generally we prefer active over passive: 

```
The proxy's memory use scales linearly with the number of active connections. Fewer open connections 
will use less memory. Adjust this value based on your applications's requirements. 
```

examples/kubernetes/proxy_with_sa_key.yaml
```diff
@@ -57,6 +57,20 @@ spec:
...
+            # less memory. Adjust this value based on your application's
+            # requirements.
+            memory: "2Gi"
+            # More database IO require more CPUs. Less database IO require less
```
kurtisvg, line 71: Same here as above.

kurtisvg (approved):

examples/kubernetes/proxy_with_sa_key.yaml
enocom: Fixed

examples/kubernetes/proxy_with_sa_key.yaml
enocom: Fixed.

examples/kubernetes/proxy_with_sa_key.yaml
enocom: Fixed.

PR #776, opened 2021-05-03
chore: add example of running the proxy as a service

### LINKED ISSUE #242: Add documentation for running cloudsql-proxy as a service in Kubernetes
for cases such as cronJobs.

enocom:
Fixes #242.


examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
+# Using the Cloud SQL Auth Proxy with PgBouncer
```
kurtisvg, line 1: Does the filepath for this make sense? Should we have both of these under the k8s folder? Maybe one folder for "sidecar" and one for "service"?

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+In this example, you will deploy [PgBouncer][] with the proxy as a sidecar, in
+addition to configuring encryption between the application and PgBouncer.
+
+## Warning
+
+This is an advanced example. Running PgBouncer may be a significant operational
+burden. Nonetheless, running the proxy with PgBouncer may be an appropriate
+choice when running the proxy as a sidecar no longer works effectively at your
+scale.
```
kurtisvg, line 15: I think this section needs to be expanded a bit. Ideally I'd like it to include that:
1. The sidecar is the recommended choice for most users because ... (simpler, less overhead, more secure, lower latency)
2. The service pattern is useful when you are (at _very_ large scale, need a middleware connection pooler, run into SQL Admin API quota problems). 

Bonus points if we can include a relative link to the sidecar example.

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+Next, we specify volume mounts in our PgBouncer container where the secrets will
+be stored:
+
+> [`pgbouncer_deployment.yaml`](pgbouncer_deployment.yaml#L31-L41))
+
+``` yaml
+- name: pgbouncer
+  image: <PG-BOUNCER-CONTAINER>
+  ports:
+  - containerPort: 8080
```
kurtisvg, line 105: Are we sure this should be 8080?

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+  ports:
+  - protocol: TCP
+    port: 5432
+    targetPort: 5432
+```
+
+With the PgBouncer service and deployment done, we are ready to point our
+application at it.
+
+## Point your application at the service
```
kurtisvg, line 187:
```suggestion
## Configure your application 
```

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+addition to configuring encryption between the application and PgBouncer.
+
+## Warning
+
+This is an advanced example. Running PgBouncer may be a significant operational
+burden. Nonetheless, running the proxy with PgBouncer may be an appropriate
+choice when running the proxy as a sidecar no longer works effectively at your
+scale.
+
+## Generate Certificates for PgBouncer
```
kurtisvg, line 17: I would consider moving the following steps as substeps under `## Initial Setup`:
1. Generate Certificates for PgBouncer
2. Save the certificates as secrets
3. Containerize PgBouncer

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+Some users may prefer to containerize PgBouncer themselves. For this example,
+we will make use of an open source container, [edoburu/pgbouncer][edoburu]. One
+nice benefit of `edoburu/pgbouncer` is that it will generate all the PgBouncer
+configuration based on environment variables passed to the container.
+
+## Deploy a PgBouncer Deployment and Service
+
+With PgBouncer containerized, we will now create a deployment with PgBouncer and
+the proxy as a sidecar.
```
kurtisvg, line 70:
```suggestion
the Cloud SQL Auth-proxy as a sidecar.
```

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+## Containerize PgBouncer
+
+Some users may prefer to containerize PgBouncer themselves. For this example,
+we will make use of an open source container, [edoburu/pgbouncer][edoburu]. One
+nice benefit of `edoburu/pgbouncer` is that it will generate all the PgBouncer
+configuration based on environment variables passed to the container.
+
+## Deploy a PgBouncer Deployment and Service
```
kurtisvg, line 67:
```suggestion
## Deploying PgBouncer as a Service
```

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+For the PgBouncer deployment, we add the proxy as a sidecar, starting it on port
+5431:
+
+> [`pgbouncer_deployment.yaml`](pgbouncer_deployment.yaml#L70-L76))
+
+``` yaml
+- name: cloud-sql-proxy
+  image: gcr.io/cloudsql-docker/gce-proxy:1.22.0
```
kurtisvg, line 158: Can we add a comment here encouraging users to update frequently

examples/pgbouncer/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+Then, we configure environment variables for connecting to the database, this
+time including a `CA_CERT`:
+
+> [`deployment.yaml`](deplyment.yaml#L32-L53))
+
+``` yaml
+env:
+- name: DB_HOST
+  value: "<YOUR-SERVICE-NAME>.default.svc.cluster.local"
```
kurtisvg, line 223: This uses the default namespace, which might trip some folks up

kurtisvg (changes requested):

examples/pgbouncer/README.md
enocom: That's a good idea.

README.md
```diff
@@ -344,5 +344,5 @@ Install via Nuget, follow these
 [releases]: https://github.com/GoogleCloudPlatform/cloudsql-proxy/releases
 [roles-and-permissions]: https://cloud.google.com/sql/docs/mysql/roles-and-permissions
 [service-account]: https://cloud.google.com/iam/docs/service-accounts
-[sidecar-example]: https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/master/examples/kubernetes#run-the-cloud-sql-proxy-as-a-sidecar
+[sidecar-example]: https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/master/examples/sidecar#run-the-cloud-sql-proxy-as-a-sidecar
```
kurtisvg, line 347: Should probably be either `kubernetes/sidecar` or `k8s-service`

enocom: Will do. I like the shallow option: `k8s-service` and `k8s-sidecar`.

examples/pgbouncer/README.md
enocom: Good catch. This should be 5432.

examples/pgbouncer/README.md
enocom: Done.

examples/pgbouncer/README.md
enocom: Made it "Cloud SQL Auth Proxy ...".

examples/pgbouncer/README.md
enocom: Done.

examples/pgbouncer/README.md
enocom: Added a comment to catch the eye.

examples/pgbouncer/README.md
enocom: Added a comment both here and in the corresponding YAML file.

examples/pgbouncer/README.md
enocom: Done.

examples/pgbouncer/README.md
enocom: Done.

examples/k8s-service/README.md
```diff
@@ -0,0 +1,275 @@
...
+Running PgBouncer with the Cloud SQL Auth Proxy may pose a significant
+operational burden and should be undertaken with caution given the attendant
+complexity.
+
+In general, we recommend [running the proxy as a sidecar][sidecar] to your
+application because it is simple, there is less overhead, it is secure out of
+the box, and there is less latency involved.
+
+However, the service pattern is useful when you are at very large scale, when
+you clearly need a database connection pooler, and when you are running into SQL
```
kurtisvg, line 22:
```suggestion
However, the service pattern is useful when you are at very large scale, when
client side connection pooling isn't sufficient, or when you are running into Cloud SQL
```

examples/k8s-service/README.md
```diff
@@ -0,0 +1,275 @@
...
+## Initial Setup
+
+Before we deploy PgBouncer with the Cloud SQL Auth Proxy, there are three
+initial steps to take.
+
+### Generate Certificates for PgBouncer
+
+First, you will need to generate certificates to encrypt the connection between
+the application and PgBouncer. We recommend using [CFSSL][] to handle
+certificate generation. Note: this example uses self-signed certificates. In
```
kurtisvg, line 34: Do we recommend it or is that just what this example is using?

kurtisvg (approved):

examples/k8s-service/README.md
enocom: I'll verify what best-practices are and update the wording to make it clear.

PR #791, opened 2021-05-10
fix: change to static base container

### LINKED ISSUE #762: Triage container vulnerabilities
## Bug Description

I'm a big fun of the proxy, but we recently started to monitor vulnerabilities in OSS containers, including the proxy, using Snyk (full disclosure: I am working for Snyk).

When testing the latest version, I can see it has 21 vulnerabilities, 3 reported as high.

Maybe someone can take a look on them please? Ensure nothing is exploitable?

## How to reproduce

  1.  snyk container test --json-file-output==output.json gcr.io/cloudsql-docker/gce-proxy:1.22.0  

You can find the full list of vulns [here](https://gist.github.com/omerlh/94950633c114790357bb6e71e2a09ed7)

enocom:
In addition, this commits ensures we compile a static binary by
disabling CGO, using the built-in net package, and omitting the DWARF
symbol table used for debug.

Fix #762.


enocom:
To test this, I ran the proxy both with workload identity and a credentials file in my sample app. In both cases, the proxy connected successfully to the database and my app worked without issue.

enocom:
A couple points:

The `netgo` tag isn't needed anymore: https://golang.org/doc/go1.5#net.

Also, I've removed the `-w` flag. Until we have a need to make smaller binaries, there's no reason to include that change here.

enocom:
Just to be sure, I've run my same experiment again with the latest commit and confirmed it works.

PR #796, opened 2021-05-14
chore: enable the race detector in test

enocom:


shubha-rajan (approved):
Looks like we have race conditions we need to fix. 😬  Not sure if you want to fix those in this PR or a separate one but the changes to the test scripts LGTM.

enocom:
Yep! I'll fix them (just one I think) in this PR. ;-)

kurtisvg (approved):

PR #817, opened 2021-06-09
fix: force new refresh of certs on connect handshake errors

### LINKED ISSUE #400: Application unable to reconnect to CloudSQL after instance maintenance
## Bug Description

When our CloudSQL instance performs maintenance that interrupts existing connections, our application cannot connect via the proxy until we restart the proxy container.
## How to reproduce

Unfortunately, I'm not sure how to reproduce this as it seems to only happen when our CloudSQL instance performs maintenance that severs existing connections, and prevents further connectivity until we restart the cloudsql-proxy sidecar container (by restarting the pod).

## Environment

- application
    - image node:lts-slim; currently node v12.17.0 (npm v6.14.4)
     - database driver: sidorares/node-mysql2 (v2.1.0)
     - We have both uncaughtException and unhandledRejection for the node process, and both of these result in a process exit, and container restart, after logging the error.
- cloudsql-proxy
     - image: gcr.io/cloudsql-docker/gce-proxy:1.16
     - sidecar to application
- CloudSQL instance (Second Generation) hosted in GCE

orjanbruland:
If the cached certificate is invalid but not expired it will cause
TLS handshakes to fail until the certificate is refreshed.

Invalidate the cached configuration on handshake errors so the next call
to DialContext will trigger a refresh.

Fixes #400


enocom (commented):
Thanks for the PR @orjanbruland. 

We were just discussing #400 this morning. By the way, we're working on a v2 of the dialer which already fixes this issue.

I'd like @kurtisvg to take a close look at this PR as well.

proxy/proxy/client.go
```diff
@@ -444,6 +445,24 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
 
+func (c *Client) invalidateCfg(cfg *tls.Config, instance string) {
+	c.cacheL.RLock()
+	e := c.cfgCache[instance]
+	c.cacheL.RUnlock()
+	if e.cfg != cfg || e.done == nil || !isValid(e) {
+		return
+	}
+	c.cacheL.Lock()
+	e = c.cfgCache[instance]
```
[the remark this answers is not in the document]
orjanbruland: Good point. Added a check below.

kurtisvg (changes requested):

proxy/proxy/client.go
```diff
@@ -416,6 +416,7 @@ func (c *Client) tryConnect(ctx context.Context, addr string, cfg *tls.Config) (
 	ret := tls.Client(conn, cfg)
 	if err := ret.Handshake(); err != nil {
 		ret.Close()
+		c.invalidateCfg(cfg, instance)
```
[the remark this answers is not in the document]
orjanbruland: By moving the operation we would also include dialer errors (unless these are filtered out). Some of the tests pass in a Dialer that returns a `sentinelError`, which would then trigger the refresh logic.

I guess we could add a hook for the tests to override/skip the invalidation logic. What do you think?

proxy/proxy/client.go
```diff
@@ -444,6 +445,26 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
 	}
 }
 
+func (c *Client) invalidateCfg(cfg *tls.Config, instance string) {
+	c.cacheL.RLock()
+	e := c.cfgCache[instance]
+	c.cacheL.RUnlock()
+	if e.cfg != cfg || e.done == nil || !isValid(e) {
```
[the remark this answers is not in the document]
orjanbruland: Sounds good, I've updated the condition to using `needsRefresh`.

proxy/proxy/client.go
```diff
@@ -444,6 +445,26 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
+		return
+	}
+	c.cacheL.Lock()
+	e = c.cfgCache[instance]
+	if e.cfg == cfg && isValid(e) {
```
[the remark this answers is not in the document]
orjanbruland: Updated the logic to match the check above.

proxy/proxy/client.go
```diff
@@ -444,6 +445,26 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
+		select {
+		case <-e.done:
+			e.done = nil
```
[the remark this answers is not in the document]
orjanbruland: Good point. I don't think there is any value in keeping it around. Updated it to delete the key from the map.

proxy/proxy/client.go
```diff
@@ -444,6 +445,32 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
+func (c *Client) invalidateCfg(cfg *tls.Config, instance string) {
+	refreshCfgBuffer := c.RefreshCfgBuffer
+	if refreshCfgBuffer == 0 {
+		refreshCfgBuffer = DefaultRefreshCfgBuffer
+	}
+
+	c.cacheL.RLock()
+	e := c.cfgCache[instance]
+	c.cacheL.RUnlock()
+	if e.cfg != cfg || needsRefresh(e, refreshCfgBuffer) {
```
kurtisvg, line 457: After thinking about this a bit more, I think that my recommendation to use `needsRefresh` here is potentially problematic, because it doesn't account for when a new refresh has started "naturally" (based on time). For example:
1. The cert is ~5 mins from expiration, so a new refresh is triggered [here](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/60cde814b784f97482d0ff7f8221e725a08471be/proxy/proxy/client.go#L297)
1. the handshake errs, `invalidateCfg` is called, but `needsRefresh` returns `false` because a new refresh is already scheduled.

In this specific scenario, I _think_ we still want to invalidate `cfg` for the instance, as we want to block the next connect until the scheduled refresh is complete - not use the old (potentially invalid) info the short term. Does that sound right? 

If so, I think using `!isValid` would be better here than `needsRefresh`.

proxy/proxy/client.go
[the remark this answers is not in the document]
kurtisvg: Can we add a quick comment here (so this isn't mistaken as accidental in the future?)
```
	// the state may have changed between critical sections, so double check
```

proxy/proxy/client.go
```diff
@@ -444,6 +445,32 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
+		c.cacheL.Unlock()
+		return
+	}
+	select {
+	case <-e.done:
+		delete(c.cfgCache, instance)
+	default:
+	}
+	c.cacheL.Unlock()
```
kurtisvg, line 471: nit: should this just be moved to a defer?

proxy/proxy/client.go
```diff
@@ -444,6 +445,32 @@ func (c *Client) selectDialer() func(context.Context, string, string) (net.Conn,
...
+	e = c.cfgCache[instance]
+	if e.cfg != cfg || needsRefresh(e, refreshCfgBuffer) {
+		c.cacheL.Unlock()
+		return
+	}
+	select {
+	case <-e.done:
+		delete(c.cfgCache, instance)
+	default:
+	}
```
kurtisvg, line 470: Following the logic in my comment above RE: `needsRefresh`, I think that this might also be incorrect:

Specifically at this point, the following assertions are true:
1. The current copy of `cfg` is the same one that lead to a handshake error, so it hasn't been invalidated yet
2. The current copy of `cfg` is valid - aka the next connect won't block in `cachedCfg` on it

I think the desired behavior here is just invalidate the config, but preserve any existing refresh. We also want to preserve `lastUpdated`, so that the refresh throttle isn't bypassed. So I think what we want instead of this select might be:
```go
c.cfgCache[instance] = cacheEntry{
    done: e.done,
    lastUpdated: e.lastUpdated
}
```

I think this can replace the `select` altogether, because if `e.done` is complete, the next connect triggers a refresh ([because it's no longer valid, so `needsRefresh` == true](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/60cde814b784f97482d0ff7f8221e725a08471be/proxy/proxy/client.go#L312)). If `e.done` isn't complete, it's also not valid, so will still block as well.

@enocom does that sound right?

kurtisvg (changes requested):
I'm still a bit concerned that introducing this behavior might have some unwanted side effects on SQL Admin API usage or connect attempts hitting the throttle. While the prototype dialer has similar behavior, it uses a bit safer approach with a rate limiter. - @enocom What do you think? Should we try to introduce a similar change here?

proxy/proxy/client.go
orjanbruland: If we remove the select statement as you suggested below, I think it should be sufficient to only check for `e.cfg != cfg` as this means that the config has either been invalidated or refreshed.

kurtisvg: I think at a high level, you only want to return early if:
1. The cfg has completed a refresh and something meaningful has changed (cert? ip?)
2. The cfg has already been invalidated (i.e., the next called to cachedCfg will block until the next refresh is complete)

The `e.cfg != cfg` seems potentially problematic in this scenario:
1. Goroutine A (_GA_): Encounters a handshake err, attempts to invalidate the config
2. GB: Calls `cachedCfg` when `needRefresh == true`, [updates `e.done and e.lastRefreshed`](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/60cde814b784f97482d0ff7f8221e725a08471be/proxy/proxy/client.go#L350)
3. GA: sees that `e.cfg != cfg` since `done` and `lastRefreshed` has changed, returns early
4. GA: Attempts to connect again, calls `cachedCfg`, `needsRefresh == false` since a new refresh was scheduled, but it also doesn't block on the new refresh (because `isValid == true`). Connect happens with the old certificate.

orjanbruland: > GA: sees that e.cfg != cfg since done and lastRefreshed has changed, returns early

I don't think this is the case as `e.cfg != cfg` compares the pointer address of the TLS config. This would return early only when a new `*tls.Config` is created (refresh completed) or when set to nil (invalidated).

kurtisvg: You're correct - I was thinking it was `cachedCfg`, not `*tls.Config`.

enocom:
> While the prototype dialer has similar behavior, it uses a bit safer approach with a rate limiter. - @enocom What do you think? Should we try to introduce a similar change here?

This is the critical section of the proxy. So I'm inclined to make as few changes as possible.

I'd like to run this for a few days and try to break it. @orjanbruland sorry for the slow response on my end. I'm actively looking at this PR now.

proxy/proxy/client.go
enocom: Yes, I agree. This seems like a cleaner approach.

enocom:
@orjanbruland How did you verify the behavior of this PR? 

I've deployed the proxy (v1.23.0) to a Kubernetes cluster with a simple app, triggered a failover, and don't see the issue described in #400.

enocom (approved):

kurtisvg (approved):

enocom:
Before we merge this, I'd like to fix our rate limiting implementation. Given some upcoming vacations, we might miss the next monthly release for this, but will get it into the following one. Apologies for the timing here.

enocom:
These test failures should be fixed by https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/851.

enocom:
@orjanbruland Thanks again for this PR and sorry for missing the last release. This will be in next week's release (v1.24.0).

PR #830, opened 2021-06-28
fix: improve log message when refresh is throttled

enocom:


shubha-rajan (approved):

proxy/proxy/client.go
```diff
@@ -352,7 +352,7 @@ func (c *Client) cachedCfg(ctx context.Context, instance string) (string, *tls.C
 				c.cfgCache[instance] = e
 			} else {
 				// TODO: Investigate returning this as an error instead of just logging
-				logging.Errorf("Throttling refreshCfg(%s): it was only called %v ago", instance, time.Since(e.lastRefreshed))
+				logging.Infof("re-using config for %s from %v ago", instance, time.Since(e.lastRefreshed))
```
kurtisvg, line 355: I think we need to work on messaging for this - we should be clear that we throttled the retry here, since it may cause repeated errors

kurtisvg (changes requested):

proxy/proxy/client.go
enocom: How about "throttling refresh attempt, re-using config for %s from %v ago"?

kurtisvg (approved):

PR #841, opened 2021-07-20
Generate key on connect

### LINKED ISSUE #840: Allow lazy RSA key creation, postponing until connect
## Lazily Generate Keys on Connect
Generating keys takes on the order of 200 ms. Currently this happens when the proxy is initialized.
In some use cases it's not clear at proxy initialization time whether the proxy will be used, delaying
key generation could save a non-trivial amount of time in this case.

## Alternatives Considered
Generating the key in a non-blocking goroutine can also reduce latency, especially if the proxy is
not going to be used immediately after initialization.

joyoza:
## Change Description

Please provide a detailed description on what changes your PR will have.

This change adds `GenerateKeyOnConnect` option to lazily generate the key when needed, at connect time.
Even if this is not set, key generation will happen in a goroutine to avoid blocking the ~200 ms it takes to generate a key.

## Checklist

- [X] Make sure to open an issue as a 
- [X] Ensure the tests and linter pass

## Relevant issues:

- Fixes #840


proxy/certs/certs.go
```diff
@@ -75,6 +76,11 @@ type RemoteOpts struct {
 
 	// Token source for token information used in cert creation
 	TokenSource oauth2.TokenSource
+
+	// GenerateKeyOnConnect, if true, causes the RSA key to be generated lazily
+	// on the first connect to a database. The default behavior is to generate
+	// the key when the CertSource is created.
+	GenerateKeyOnConnect bool
```
kurtisvg, line 83: nit: consider "delayKeyGenerate"

proxy/certs/certs.go
```diff
@@ -249,6 +265,23 @@ func parseCert(pemCert string) (*x509.Certificate, error) {
...
+	if s.key != nil {
+		return s.key
+	}
+	start := time.Now()
+	pkey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err) // very unexpected.
+	}
+	logging.Verbosef("Generated RSA key in %v", time.Since(start))
+	s.key = pkey
```
kurtisvg, line 281: I think this can be optimized with a [`Once`](https://pkg.go.dev/sync#Once.Do) instead of a mutex

proxy/proxy/dial.go
```diff
@@ -78,8 +84,10 @@ type Dialer func(net, addr string) (net.Conn, error)
 func Init(auth *http.Client, connset *ConnSet, dialer Dialer) {
 	dialClient.Lock()
 	dialClient.c = &Client{
-		Port:   port,
-		Certs:  certs.NewCertSource("", auth, true),
+		Port: port,
+		Certs: certs.NewCertSourceOpts(auth, certs.RemoteOpts{
+			GenerateKeyOnConnect: GenerateKeyOnConnect,
```
kurtisvg, line 89: This defaults to false, so is it necessary to include it here? Would it be better to have the user use `InitWithClient` to configure the client instead?

kurtisvg (changes requested):

proxy/certs/certs.go
joyoza: Done

proxy/certs/certs.go
joyoza: Done, yeah that's a bit simpler.

proxy/proxy/dial.go
joyoza: Done, but I had to make `Port` an exported constant so users can see it. Alternately I could set `Client.Port` in `InitWithClient()` if it's zero? The cleaner solution might be an `Init` function for `Client`.

proxy/proxy/dial.go
```diff
@@ -25,7 +25,15 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
-const port = 3307
+// The port that CloudSQL expects the client to connect to.
+const Port = 3307
```
kurtisvg, line 29: Let's just rename this `DefaultPort` if we have to export it.

proxy/proxy/dial.go
```diff
@@ -25,7 +25,15 @@ import (
...
+
+// DelayKeyGenerate (if true) defers the generation of the CertSource's RSA
+// key to the first connection attempt to a Cloud SQL database. This is useful
+// in scenarios when the proxy needs to be created no matter what (e.g. when
+// mounting the FUSE), but may never be used. If this behavior is desired, this
+// var should be set to true before calling proxy.Init.
+var DelayKeyGenerate bool
```
kurtisvg, line 36: This value isn't used anymore, I think we can remove it.

kurtisvg (changes requested):

proxy/proxy/dial.go
joyoza: Whoops, I thought I'd already done that. Done now.

proxy/proxy/dial.go
joyoza: Done.

proxy/certs/certs.go
```diff
@@ -75,6 +76,11 @@ type RemoteOpts struct {
 
 	// Token source for token information used in cert creation
 	TokenSource oauth2.TokenSource
+
+	// DelayKeyGenerate, if true, causes the RSA key to be generated lazily
+	// on the first connect to a database. The default behavior is to generate
```
enocom, line 81: s/connect/connection/

kurtisvg (approved):

enocom (approved):
Minor nit on a docstring. Otherwise, LGTM.

proxy/certs/certs.go
joyoza: Done

PR #850, opened 2021-07-28
chore: use golang.org/x/time/rate for rate limiting

enocom:


proxy/proxy/client.go
```diff
@@ -342,10 +346,18 @@ func (c *Client) cachedCfg(ctx context.Context, instance string) (string, *tls.C
...
+		if limiter == nil {
+			limiter = rate.NewLimiter(rate.Every(throttle), 1)
+			c.limiters[instance] = limiter
+		}
 		if needsRefresh(e, refreshCfgBuffer) {
-			if time.Since(e.lastRefreshed) >= throttle {
+			if limiter.Allow() {
 				// start a new refresh and update the cachedEntry to reflect that
 				e.done = c.startRefresh(instance, refreshCfgBuffer)
 				e.lastRefreshed = time.Now()
```
enocom, line 363: We could remove this value as it's only used in the logging message.

kurtisvg: I'm okay with it being in the logging message.

kurtisvg (approved):

PR #861, opened 2021-08-02
fix: use Google Cloud structured log format

### LINKED ISSUE #860: cloudsql proxy structured log entries incompatible with google cloud logging
## Bug Description

The option `-structured_logs` the cloudsql proxy writes out single line json documents for each log entry. I expected this to be picked up by Google Cloud Logging, but the fields do not match the [specification](https://cloud.google.com/logging/docs/structured-logging#special-payload-fields).

It should have a field `severity` with the proper encoding of the value (in caps) and a field `message` which contains the log message. The option writes out the fields `level` and `msg`.

If these fields are added, Google Cloud Logging will reflect the correct severity level even if it is written to stderr.

## Example code (or command)

```
$ cloud_sql_proxy -structured_logs -log_debug_stdout=true
{"level":"info","ts":1627910610.2727964,"caller":"cloud_sql_proxy/proxy.go:161","msg":"New connection for \"binx-io-dev:europe-west1:binx-io-3o\""}
{"level":"info","ts":1627910611.800329,"caller":"cloud_sql_proxy/cloud_sql_proxy.go:634","msg":"Received TERM signal. Waiting up to 0s before terminating."}
```

## Stacktrace
see above.

## How to reproduce

  1. run cloud_sql_proxy with -structured_logs
  2. check output and google cloud logging

## Environment

1. OS type and version: Alpine
2. Cloud SQL Proxy version (`./cloud_sql_proxy -version`): 1.23.0

mvanholsteijn:
## Change Description

This fix will generate log messages as json documents that are compatible with [Google Cloud Logging](https://cloud.google.com/logging/docs/structured-logging#special-payload-fields). This fix ensure that Google Cloud Logging will pick up the severity and text message, even if the log is written on stderr.

## Checklist

- [X] Make sure to open an issue as a 
- [X] Ensure the tests and linter pass
- [X] Appropriate documentation is updated (if necessary)

There was no test for EnableStructuredLogs() and it is hard to create one. So I left it this way. I did test the flag manually. Given the complexity of the change, I hope this is ok.

## Relevant issues:

- Fixes #860

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -78,12 +78,17 @@ func EnableStructuredLogs(logDebugStdout, verbose bool) (func(), error) {
 		consoleDebugging = zapcore.Lock(os.Stdout)
 	}
 
-	consoleEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+	config := zap.NewProductionEncoderConfig()
+	config.LevelKey = "severity"
+	config.MessageKey = "message"
+	config.TimeKey = "timestamp"
+	config.EncodeLevel = zapcore.CapitalLevelEncoder
+	config.EncodeTime = zapcore.ISO8601TimeEncoder
+	consoleEncoder := zapcore.NewJSONEncoder(config)
 	core := zapcore.NewTee(
 		zapcore.NewCore(consoleEncoder, consoleErrors, highPriority),
 		zapcore.NewCore(consoleEncoder, consoleDebugging, lowPriority),
 	)
-
 	// By default, caller and stacktrace are not included, so add them here
 	logger := zap.New(core, zap.AddCaller(), zap.AddStacktrace(zapcore.ErrorLevel))
 
--- a/README.md
+++ b/README.md
@@ -276,13 +276,11 @@ this flag to true.  Defaults to false.
 
 #### `-structured_logs`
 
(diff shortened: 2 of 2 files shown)


logging/logging.go
```diff
@@ -58,7 +58,7 @@ func DisableLogging() {
 
 // EnableStructuredLogs replaces all logging functions with structured logging
 // variants.
-func EnableStructuredLogs(logDebugStdout, verbose bool) (func(), error) {
+func EnableStructuredLogs(logDebugStdout, googleLogFormat, verbose bool) (func(), error) {
```
enocom, line 61: The structured logging should have worked with Cloud Logging when it was introduced. So, I think of this as a bug fix and so it's fine to just update the format within this function without introducing another flag.

enocom: Let's remove the flag, and alter the format inline, adding a comment explaining that it adheres to what Cloud Logging requires. How does that sound?

enocom (changes requested):
Thanks for the contribution! Just a minor change in how we do this suggested below.

logging/logging.go
mvanholsteijn: great! I did it this way to ensure that I did not break any existing log metrics/alerts people may have created based on the original format. I did a force push of the update.

enocom (approved):
Looks great. Thanks!

PR #865, opened 2021-08-04
docs: Remove warnings regarding IAM login support for MySQL

AdamBSteele:
IAM login now supported by MySQL, per [docs](https://cloud.google.com/sql/docs/mysql/iam-logins). Also in MySQL documentation: "...we recommend you use the Cloud SQL Auth proxy when using IAM database authentication for long-lived processes or applications that rely on connection pooling."


kurtisvg:
Hey @AdamBSteele,

Thanks for taking the time to open a PR. It's not super clear from the current doc page, but there are two ways to do IAM DB AuthN - manually (supplying the Oauth2 token as a password) and automatically (using a Cloud SQL connector (such as the proxy) with the automatic option enabled). In this case, MySQL currently supports manual authentication, but not automatic. 

There are some bigger updates coming to that page to make this more clear.

PR #869, opened 2021-08-05
Add instance dialing to health check

tifftoff:
## Change Description

Add instance dialing to health check readiness test. Allow readiness to succeed only when all instances can be dialed successfully.

## Checklist

- [ ] Make sure to open an issue as a [bug/issue] (https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/new/choose) 
  before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea.


enocom:
We're going to remake this PR.

PR #877, opened 2021-08-09
fix: ensure proxy shuts down gracefully on SIGTERM

### LINKED ISSUE #876: The proxy doesn't shutdown gracefully when running in fuse mode
When the proxy receives a SIGTERM, it [immediately exits](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/5dd6d25e5abd590990dd6237ec303516bb73b72a/cmd/cloud_sql_proxy/cloud_sql_proxy.go#L658) preventing any of the `defer` statements registered in the main goroutine from running. In fuse mode this is particular egregious because the fuse mount is not cleaned up and has to be manually unmonuted.

enocom:
Fixes #876

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -442,7 +442,7 @@ func gcloudProject() ([]string, error) {
 	return []string{cfg.Configuration.Properties.Core.Project}, nil
 }
 
-func startProxy() int {
+func runProxy() int {
 	flag.Parse()
 
 	if *version {
@@ -659,6 +659,6 @@ func startProxy() int {
 }
 
 func main() {
-	code := startProxy()
+	code := runProxy()
 	os.Exit(code)
 }
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -442,29 +442,12 @@ func gcloudProject() ([]string, error) {
 	return []string{cfg.Configuration.Properties.Core.Project}, nil
 }
 
-// Main executes the main function of the proxy, allowing it to be called from tests.
-//
-// Setting timeout to a value greater than 0 causes the process to panic after
-// that amount of time. This is to sidestep an issue where sending a Signal to
-// the process (via the SSH library) doesn't seem to have an effect, and
-// closing the SSH session causes the process to get leaked. This timeout will
-// at least cause the proxy to exit eventually.
-func Main(timeout time.Duration) {
-	if timeout > 0 {
-		go func() {
-			time.Sleep(timeout)
-			panic("timeout exceeded")
-		}()
-	}
-	main()
-}
-
-func main() {
+func startProxy() error {
 	flag.Parse()
 
 	if *version {
 		fmt.Println("Cloud SQL Auth proxy:", semanticVersion())
-		return
+		return nil
 	}
 
 	if *logDebugStdout {
@@ -479,7 +462,7 @@ func main() {
 		cleanup, err := logging.EnableStructuredLogs(*logDebugStdout, *verbose)
 		if err != nil {
 			logging.Errorf("failed to enable structured logs: %v", err)
-			os.Exit(1)
+			return err
 		}
 		defer cleanup()
 	}
@@ -501,7 +484,7 @@ func main() {
 	if *host != "" && !strings.HasSuffix(*host, "/") {
 		logging.Errorf("Flag host should always end with /")
 		flag.PrintDefaults()
-		return
+		return errors.New("invalid command")
 	}
 
 	// TODO: needs a better place for consolidation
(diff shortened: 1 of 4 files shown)


cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -651,11 +652,10 @@ func main() {
 		}()
 
 		err := proxyClient.Shutdown(*termTimeout)
-		if err == nil {
-			os.Exit(0)
+		if err != nil {
+			logging.Errorf("Error during SIGTERM shutdown: %v", err)
+			os.Exit(1)
```
kurtisvg, line 657: This os.Exit here - does this still ensure `cancel()` is called?

enocom: I double checked and interestingly, no. A call to `os.Exit` exits immediately ignoring any defers.

Cf. https://stackoverflow.com/questions/27629380/how-to-exit-a-go-program-honoring-deferred-calls

enocom:
This change ended up being larger than I expected. With the calls to os.Exit, we were actually failing to run any previous defer statements.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -666,5 +652,15 @@ func main() {
...
+	proxyClient.RunContext(ctx, connSrc)
+	if err, ok := <-shutdown; ok {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	if err := startProxy(); err != nil {
+		os.Exit(1)
```
kurtisvg, line 664: Some of the other `os.Exits` return different error codes return a 2 - this would possible be a breaking change?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -650,12 +634,14 @@ func main() {
 			}
 		}()
 
+		cancel()
```
kurtisvg, line 637: If this is canceled before proxyClient.Shutdown, won't it end all of the connections that you've passed it to?

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -442,29 +442,12 @@ func gcloudProject() ([]string, error) {
...
 }
 
-// Main executes the main function of the proxy, allowing it to be called from tests.
-//
-// Setting timeout to a value greater than 0 causes the process to panic after
-// that amount of time. This is to sidestep an issue where sending a Signal to
-// the process (via the SSH library) doesn't seem to have an effect, and
-// closing the SSH session causes the process to get leaked. This timeout will
-// at least cause the proxy to exit eventually.
-func Main(timeout time.Duration) {
```
kurtisvg, line 452: Main is exported, so this change is breaking as well?

kurtisvg (changes requested):

cmd/cloud_sql_proxy/cloud_sql_proxy.go
enocom: Fortunately, the `main` package cannot be imported anywhere, so in this case, we're the only possible consumers of this function, and it's unused in test.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
enocom: I'll update the `startProxy` function to return an exit code, to maintain the different error codes.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
enocom: Yes it would. I suppose we should give clients `termTimeout` to close themselves, and then cancel the context after that warning. I'll make this run on function exit.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -650,12 +634,14 @@ func main() {
...
-		if err == nil {
-			os.Exit(0)
+		if err != nil {
+			logging.Errorf("Error during SIGTERM shutdown: %v", err)
+			shutdown <- 2
+			return
 		}
-		logging.Errorf("Error during SIGTERM shutdown: %v", err)
-		os.Exit(2)
+		close(shutdown)
```
kurtisvg, line 644: nit: should this be deferred earlier?

kurtisvg (approved):

enocom:
Ran the linting manually.

kurtisvg (approved):

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -442,29 +442,12 @@ func gcloudProject() ([]string, error) {
...
-		go func() {
-			time.Sleep(timeout)
-			panic("timeout exceeded")
-		}()
-	}
-	main()
-}
-
-func main() {
+func startProxy() int {
```
kurtisvg, line 445: nit: This is probably more like `runProxy` since it doesn't return until complete

kurtisvg (approved):

cmd/cloud_sql_proxy/cloud_sql_proxy.go
enocom: Done.

PR #892, opened 2021-08-14
chore: make fuse tests reliable

### LINKED ISSUE #856: fuse tests are unreliable
The fuse tests sometimes fail and sometimes don't, even though we've not changed any of the fuse code for a long time now. This is the tracking issue to fix the flakey tests.

enocom:
The fuse tests were unreliable for two reasons:
1. All tests used the same two directories to mount and unmount FUSE file systems, and
2. The FUSE mount takes some amount of time, so promptly trying to close the connection was subject to race conditions.

This PR addresses both issues:
1. It uses distinct directories for each test, and
2. It attempts to close the connection at most 10 times before failing.

I've manually run the tests 15 times to verify (at least by best effort) that this fix does in fact work.

Fixes #856

### DIFF
--- a/proxy/fuse/fuse_test.go
+++ b/proxy/fuse/fuse_test.go
@@ -83,7 +83,7 @@ func TestReadme(t *testing.T) {
 	}
 	defer fuse.Close()
 
-	data, err := os.ReadFile(filepath.Join(dir, "README"))
+	data, err := ioutil.ReadFile(filepath.Join(dir, "README"))
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


enocom:
Test Run Record (manually invoking the Kokoro build)
1. :heavy_check_mark: 
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO
1. TODO

shubha-rajan (approved):
Thanks for fixing these!

PR #923, opened 2021-09-08
chore: configure Go 1.17 in Kokoro

enocom:


kurtisvg (approved):

kurtisvg:
Do we want to target go 1.17 for this? Or should we test all "supported" go versions

PR #953, opened 2021-09-29
fix: replace library for FUSE support

### LINKED ISSUE #937: FUSE fails to mount on macOS Big Sur (Apple Silicon and x86)
The proxy fails to start on macOS Big Sur when using the `--fuse` flag. Full details below:

```
# Apple Silicon Mac                                                                                                                                            
# macOS 11.5.2                                                                                                                                                 
                                                                                                                                                               
% ./cloud_sql_proxy -fuse -dir /tmp/cloudsql                                                                                                                   
2021/09/16 13:41:30 Rlimits for file descriptors set to {Current = 8500, Max = 9223372036854775807}                                                            
2021/09/16 13:41:31 Mounting /tmp/cloudsql...                                                                                                                  
2021/09/16 13:41:32 Generated RSA key in 102.874334ms                                                                                                          
kext load failed: -603947007                                                                                                                                   
2021/09/16 13:41:33 WARNING: Mount failed - attempting to unmount dir to resolve..., dir=/tmp/cloudsql                                                         
2021/09/16 13:41:33 Unmount failed: unmount /tmp/cloudsql: no such file or directory                                                                           
kext load failed: -603947007                                                                                                                                   
2021/09/16 13:41:33 Could not start fuse directory at "/tmp/cloudsql": cannot mount "/tmp/cloudsql": exit status 255                                           
                                                                                                                                                               
% ./cloud_sql_proxy -version                                                                                                                                   
Cloud SQL Auth proxy: 1.25.0+darwin.arm64                                                                                                                      
                                                                                                                                                               
% brew info osxfuse                                                                                                                                            
osxfuse: 3.11.2                                                                                                                                                
https://osxfuse.github.io/                                                                                                                                     
/opt/homebrew/Caskroom/osxfuse/3.11.2 (26 files, 6.9MB)                                                                                                        
From: https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/osxfuse.rb  
```

```
# x86_64 Mac                                                                                                                                                   
# macOS 11.5.1                                                                                                                                                 
                                                                                                                                                               
% ./cloud_sql_proxy -fuse -dir /tmp/cloudsql                                                                                                                   
2021/09/16 14:00:32 Rlimits for file descriptors set to {Current = 8500, Max = 9223372036854775807}                                                            
2021/09/16 14:00:33 Mounting /tmp/cloudsql...                                                                                                                  
2021/09/16 14:00:33 Generated RSA key in 110.094138ms                                                                                                          
2021/09/16 14:00:37 WARNING: Mount failed - attempting to unmount dir to resolve..., dir=/tmp/cloudsql                                                         
2021/09/16 14:00:37 Unmount failed: unmount /tmp/cloudsql: operation not permitted                                                                             
2021/09/16 14:00:40 Could not start fuse directory at "/tmp/cloudsql": cannot mount "/tmp/cloudsql": exit status 1                                             
                                                                                                                                                               
% ./cloud_sql_proxy -version                                                                                                                                   
Cloud SQL Auth proxy: 1.25.0+darwin.amd64                                                                                                                      
                                                                                                                                                               
% brew info osxfuse                                                                                                                                            
osxfuse: 3.11.2                                                                                                                                                
https://osxfuse.github.io/                                                                                                                                     
/usr/local/Caskroom/osxfuse/3.11.2 (26 files, 6.9MB)                        
```

Currently, the proxy uses [bazil.org/fuse](https://github.com/bazil/fuse), which [removed support for macOS](https://github.com/bazil/fuse/commit/60eaf8f021ce00e5c52529cdcba1067e13c1c2c2). Instead, we might try replacing the FUSE internals with https://github.com/hanwen/go-fuse.

enocom:
This commit refactors the internals of proxy/fuse to use
github.com/hanwen/go-fuse/v2 instead of bazil.org/fuse, partly in
pursuit of better macOS support.

Fixes #937.


enocom:
I tested this manually on macOS Big Sur (Apple M1 & x86) and confirmed it worked in both cases. 

The FUSE mount will fail to create a Unix socket if the path exceeds the system limit (see https://github.com/golang/go/issues/6895) and the resulting error isn't very friendly (`bind: invalid argument`), but I'll fix that on a PR after this one.

Also, the Kokoro build for Go 1.16 on mac is stalled, because the installation of osxfuse requires a restart. We'll have to figure out how to work with that problem in a future change, too.

enocom:
As a side note, installing macfuse on M1 machines is quite an undertaking. After using `brew install macfuse`, I had to restart into recovery mode, [allow kernel extensions](https://helpdesk.egnyte.com/hc/en-us/articles/360055258492-Desktop-App-M1-silicon-support), restart again, and only then would the proxy start in fuse mode.

proxy/fuse/fuse.go
```diff
@@ -93,98 +80,94 @@ func NewConnSrc(mountdir, tmpdir string, client *proxy.Client, connset *proxy.Co
...
 	if ch := r.dst; ch != nil {
-		ch <- proxy.Conn{instance, c}
+		ch <- proxy.Conn{Instance: instance, Conn: c}
 	} else {
 		logging.Errorf("Ignored new conn request to %q: system has been closed", instance)
 	}
 	r.RUnlock()
 }
 
-func (r *fsRoot) Forget() {
```
kurtisvg, line 180: Just to confirm - fsRoot isn't exported, so this shouldn't be a breaking change?

proxy/fuse/fuse.go
```diff
@@ -332,56 +307,76 @@ func (r *fsRoot) listenerLifecycle(l net.Listener, instance, path string) {
...
-func (r *fsRoot) ReadDirAll(context.Context) ([]fuse.Dirent, error) {
-	ret := []fuse.Dirent{
-		{Name: "README", Type: fuse.DT_File},
+// Readdir implements fs.NodeReaddirer and returns a list of files for each
+// instance to which the proxy is actively connected. In addition, the list
+// includes a README.
+func (r *fsRoot) Readdir(ctx context.Context) (fs.DirStream, syscall.Errno) {
+	activeConns := r.connset.IDs()
+	entries := []fuse.DirEntry{
+		{Name: "README", Mode: 0777 | fuse.S_IFREG},
```
kurtisvg, line 316: Does `0777` imply right on the file? Should it be read only?

proxy/fuse/fuse.go
```diff
@@ -332,56 +307,76 @@ func (r *fsRoot) listenerLifecycle(l net.Listener, instance, path string) {
...
+	return ds, fs.OK
 }
 
-// readme implements the REAME file found in the mounted folder. It is a
-// static read-only text file.
-type readme struct{}
+// readme represents a static read-only text file.
+type readme struct {
+	fs.Inode
+}
```
kurtisvg, line 331: Can you explain why we are using `fs.Inode` here instead of in the interface below? (it doesn't look like any of the functions actually use `readme` so I'm curious we we're using a pointer).

kurtisvg (changes requested):

PR #959, opened 2021-09-30
feat: verify FUSE is installed on macOS / linux

### LINKED ISSUE #994: Verify FUSE is installed on macOS
Presently, we always assume FUSE is installed on macOS. Instead, we should check before proceeding.

See https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/bd8225377bfa276a3352ec22a7887a823f195d5b/proxy/fuse/fuse.go#L60.

enocom:
This commit addresses a longstanding [TODO](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/b0184c4b9823bbf70249f92410b5e37cca729359/proxy/fuse/fuse.go#L56) and is related to https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/953.

Fixes #994.

### DIFF
--- a/proxy/fuse/fuse_darwin.go
+++ b/proxy/fuse/fuse_darwin.go
@@ -3,7 +3,7 @@ package fuse
 import (
 	"os"
 
-	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/logging"
+	"github.com/GoogleCloudPlatform/cloudsql-proxy/logging"
 )
 
 const (
--- a/proxy/fuse/fuse_darwin.go
+++ b/proxy/fuse/fuse_darwin.go
@@ -1,6 +1,10 @@
 package fuse
 
-import "os"
+import (
+	"os"
+
+	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/logging"
+)
 
 const (
 	macfusePath = "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse"
--- a/proxy/fuse/fuse_darwin.go
+++ b/proxy/fuse/fuse_darwin.go
@@ -1,0 +1,24 @@
+package fuse
+
+import "os"
+
+const (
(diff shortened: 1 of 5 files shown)


proxy/fuse/fuse_darwin.go
```diff
@@ -0,0 +1,28 @@
+package fuse
+
+import (
+	"os"
+
+	"github.com/GoogleCloudPlatform/cloudsql-proxy/logging"
+)
+
+const (
+	macfusePath = "/Library/Filesystems/macfuse.fs/Contents/Resources/mount_macfuse"
```
enocom, line 10: This comes from https://github.com/hanwen/go-fuse/blob/0f728ba15b38579efefc3dc47821882ca18ffea7/fuse/mount_darwin.go#L121-L124.

enocom:
Here's what the error message looks like when FUSE is not installed on mac:

```
./cloud_sql_proxy -dir /Users/enocom/mnt -fuse
2021/09/30 14:18:09 Rlimits for file descriptors set to {Current = 8500, Max = 9223372036854775807}
2021/09/30 14:18:10 FUSE support on macOS depends on osxfuse or macfuse.
Neither were found on your machine. For installation instructions,
see https://osxfuse.github.io.
2021/09/30 14:18:10 FUSE not supported on this system
```

proxy/fuse/fuse_linux.go
```diff
@@ -0,0 +1,6 @@
+package fuse
+
+// Supported returns true if the current system supports FUSE.
+func Supported() bool {
+	return true
```
kurtisvg, line 5: Is there a way to check if FUSE is installed on Linux?

kurtisvg (changes requested):

proxy/fuse/fuse_darwin.go
kurtisvg: Can we make this a comment in the code as well?

kurtisvg (approved):

PR #960, opened 2021-09-30
fix: return early when socket length is too long

enocom:


proxy/fuse/fuse.go
```diff
@@ -43,15 +43,25 @@ import (
...
 	"golang.org/x/net/context"
 )
 
+// maxAllowsSocketPathLength is the maximum socket path length. The value
+// differs between platforms. Typically Linux supports 108 characters, but here
+// we tune the value down slightly to support macOS as well. Note: the value
+// leaves room for one additional null character to terminate the string.
+//
+// See https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars
+const maxAllowsSocketPathLength = 103
```
kurtisvg, line 63: This is platform dependent, I'm not sure it make sense to hardcode a value.

kurtisvg (changes requested):

PR #973, opened 2021-10-07
feat: add support for specifying cert duration

enocom:
This commit adds a flag to the proxy such that callers may specify the
duration of each ephemeral certificate. By default, the duration remains
1 hour. Optionally a caller may increase that duration up to 24 hours.


enocom:
I'm going to close this. We can re-open it when we're ready for this change.

PR #1003, opened 2021-11-03
docs: suggest macfuse over osxfuse

arbourd:
This is probably the preferred option for macOS 12.0 and M1 chips. 

I noticed that the kokoro build is also installing `osxfuse`, but am unaware of what version of macOS it might be using (not a Googler).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,7 +92,7 @@ sudo apt-get install fuse
 On macOS, use:
 
 ```
-brew install --cask osxfuse
+brew install --cask macfuse
 ```
 
 ### How to run just unit tests


enocom (approved):

enocom:
Thanks @arbourd! Separately, I'm working on updating our macos build which I hope will include updating to macfuse.

arbourd:
The `conventionalcommits.org` status went from green to red. Do I need to fix that to make it mergable?

enocom:
No worries. I just changed the PR title to include a docs prefix.

arbourd:
oh, woops. only fixed the commit. thanks @enocom

PR #1040, opened 2021-12-08
feat: add support for ReadTime in Admin API requests

### LINKED ISSUE #977: Add support for ReadTime in Ephemeral Cert request
A recent PR switched the proxy over to using the [Connect API](https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/connect/generateEphemeralCert). Part of that API includes a `ReadTime` field that allows for stale reads.

This feature request is to add support for `ReadTime` in the [request to generate an ephemeral certificate](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/15608f066cae475012f86d5d10ef83a0ad122cf7/proxy/certs/certs.go#L212-L214).

enocom:
Now when a SQL Admin API requests fails, we'll supply a read time set to
30 seconds before the present time to alleviate failures.

Fixes #977.

### DIFF
--- a/proxy/certs/certs.go
+++ b/proxy/certs/certs.go
@@ -160,17 +160,19 @@ const (
 	backoffRetries = 5
 )
 
+// now returns the current time in UTC. It is defined as a var so tests can
+// replace it with a fixed return value.
 var now = func() time.Time {
 	return time.Now().UTC()
 }
 
-func backoffAPIRetry(desc, instance string, do func(timestamp time.Time) error) error {
+func backoffAPIRetry(desc, instance string, do func(staleRead time.Time) error) error {
 	var (
 		err error
-		ts  time.Time
+		t   time.Time
 	)
 	for i := 0; i < backoffRetries; i++ {
-		err = do(ts)
+		err = do(t)
 		gErr, ok := err.(*googleapi.Error)
 		switch {
 		case !ok:
@@ -192,7 +194,7 @@ func backoffAPIRetry(desc, instance string, do func(timestamp time.Time) error) 
 		logging.Errorf("Error in %s %s: %v; retrying in %v", desc, instance, err, sleep)
 		time.Sleep(sleep)
 		// Create timestamp 30 seconds before now for stale read requests
-		ts = now().Add(-30 * time.Second)
+		t = now().Add(-30 * time.Second)
 	}
 	return err
 }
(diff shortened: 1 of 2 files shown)


proxy/certs/certs.go
```diff
@@ -160,10 +160,17 @@ const (
 	backoffRetries = 5
 )
 
-func backoffAPIRetry(desc, instance string, do func() error) error {
-	var err error
+var now = func() time.Time {
+	return time.Now().UTC()
+}
```
kurtisvg, line 165: Is this function really needed? It seems like just a bit of indirection

proxy/certs/certs.go
```diff
@@ -310,7 +322,10 @@ func (s *RemoteCertSource) Remote(instance string) (cert *x509.Certificate, addr
 	req := s.serv.Connect.Get(p, regionName)
 
 	var data *sqladmin.ConnectSettings
-	err = backoffAPIRetry("get instance", instance, func() error {
+	err = backoffAPIRetry("get instance", instance, func(timestamp time.Time) error {
```
kurtisvg, line 325: nit: consider changing timestamp to "stale_read_time" (since that's what it actually is)

proxy/certs/certs.go
kurtisvg: It looks like we are using this for testing. Can we add a comment here to clarify that?

proxy/certs/certs_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+func patchNowFunc(t *testing.T, nowTime string) func() {
+	oldNow := now
+
+	aTime, err := time.Parse(time.RFC3339, nowTime)
+	if err != nil {
+		t.Fatal(err)
+	}
+	now = func() time.Time { return aTime }
+	return func() { now = oldNow }
```
kurtisvg, line 56: I'm not sure if it matters, but this is definitely not threadsafe

kurtisvg (approved):

proxy/certs/certs.go
enocom: Will do.

proxy/certs/certs.go
enocom: Good idea.

proxy/certs/certs_test.go
enocom: I'll add a note to that effect. It's hard to get a good test assertion when time is involved without taking an approach like this. However, it also means any test changing the `now` function will potentially collide with another concurrent test run.

enocom: Updated the time assertions for a better result.

kurtisvg (approved):

PR #1044, opened 2021-12-09
feat: add support for specifying a quota project

enocom:
This commit adds a CLI flag quota_project such that users may provide a
project against which the SQL Admin API quota is calculated. Typically,
this flag will be used for internal development purposes.

Note: there is an API option WithQuotaProject that does the same thing.
However, that option is not compatible with WithHTTPClient and so we
manually add the quota project instead.

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -397,7 +397,7 @@ func (t quotaProjectTransport) RoundTrip(req *http.Request) (*http.Response, err
 	return t.base.RoundTrip(req)
 }
 
-// configureQuotaProject returns an HTTP client that uses the provided project
+// configureQuotaProject configures an HTTP client to use the provided project
 // for quota calculations for all requests.
 func configureQuotaProject(c *http.Client, project string) {
 	// Copy the given client's tripper. Note that tripper can be nil, which is equivalent to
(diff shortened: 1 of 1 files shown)


cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -374,6 +376,42 @@ func authenticatedClient(ctx context.Context) (*http.Client, oauth2.TokenSource,
...
+func (t quotaProjectTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Header == nil {
+		req.Header = make(http.Header)
+	}
+	req.Header.Add("X-Goog-User-Project", t.project)
+	return t.base.RoundTrip(req)
+}
+
+// configureQuotaProject returns a HTTP client that adds a given user project for all
+// requests.
```
kurtisvg:
```suggestion
// configureQuotaProject configures an HTTP client to use a given user project for all 
// requests.
```

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -374,6 +376,42 @@ func authenticatedClient(ctx context.Context) (*http.Client, oauth2.TokenSource,
...
+func (t quotaProjectTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Header == nil {
+		req.Header = make(http.Header)
+	}
+	req.Header.Add("X-Goog-User-Project", t.project)
+	return t.base.RoundTrip(req)
+}
+
+// configureQuotaProject returns an HTTP client that uses the provided project
+// for quota calculations for all requests.
```
kurtisvg, line 401:
```suggestion
// configureQuotaProject configures an HTTP client that uses the provided project
// for quota calculations for all requests.
```

kurtisvg (approved):

cmd/cloud_sql_proxy/cloud_sql_proxy.go
enocom: Done.

PR #1046, opened 2021-12-10
feat: allow multiple -instances flags

### LINKED ISSUE #1030: FR: Support multiple --instances flag
It'd be nice if clousd_sql_proxy accepts multiple --instances flags combined.
In k8s Deployment config, let's say if you need to specify 10 instances

```
command:
  - "/cloud_sql_proxy"
  - "--instances=some-project:some-zone:instance-name-1=tcp:3366"
  - "--instances=some-project:some-zone:instance-name-2=tcp:3367"
  - ...
```

is much easier to read than everything jammed together in one line.
If this sounds good please let me know and I'm happy to send a PR.

tsawada:
Fixes #1030

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -494,7 +494,7 @@ to listen on a custom address by providing a host, e.g., '=tcp:0.0.0.0:port'. If
 no value is provided for 'tcp', one socket file per instance is opened in 'dir'.
 For convenience, this flag may be specified multiple times.
 You may use the INSTANCES environment variable for the same effect. Using both will
-use value from flag, Not compatible with -fuse.`,
+use the value from the flag, Not compatible with -fuse.`,
 	)
 
 	flag.Parse()
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -75,7 +75,7 @@ Unix socket-based connections.`)
 		`Open sockets for each Cloud SQL Instance in the projects specified
 (comma-separated list)`,
 	)
-	instances instancesValue // -instances flag is defined in runProxy()
+	instances   instancesValue // -instances flag is defined in runProxy()
 	instanceSrc = flag.String("instances_metadata", "", `If provided, it is treated as a path to a metadata value which
 is polled for a comma-separated list of instances to connect to. For example,
 to use the instance metadata value named 'cloud-sql-instances' you would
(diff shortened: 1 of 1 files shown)


enocom (changes requested):
A couple of minor comments. Otherwise, this looks fine.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -295,6 +289,15 @@ func userAgentFromVersionString() string {
 
 const accountErrorSuffix = `Please create a new VM with Cloud SQL access (scope) enabled under "Identity and API access". Alternatively, create a new "service account key" and specify it using the -credential_file parameter`
 
+func (i *stringListValue) String() string {
```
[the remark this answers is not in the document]
tsawada: ah, forgot that modern languages allow forward declarations!

tsawada:
Thanks @enocom for your review! Please take another look.

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -481,6 +486,17 @@ func gcloudProject() ([]string, error) {
...
 
 func runProxy() int {
+	flag.Var(&instances, "instances",
+		`Comma-separated list of fully qualified instances (project:region:name)
+to connect to. If the name has the suffix '=tcp:port', a TCP server is opened
+on the specified port on localhost to proxy to that instance. It is also possible
+to listen on a custom address by providing a host, e.g., '=tcp:0.0.0.0:port'. If
+no value is provided for 'tcp', one socket file per instance is opened in 'dir'.
+For convenience, this flag may be specified multiple times.
+You may use INSTANCES environment variable for the same effect. Using both will
```
enocom, line 496:
```suggestion
You may use the INSTANCES environment variable for the same effect. Using both will
```

enocom (approved):
A couple of minor wording details.

Thanks for this contribution!

cmd/cloud_sql_proxy/cloud_sql_proxy.go
```diff
@@ -73,15 +73,7 @@ Unix socket-based connections.`)
 		`Open sockets for each Cloud SQL Instance in the projects specified
 (comma-separated list)`,
 	)
-	instances = flag.String("instances", "",
```
akinfermo, line 76: Why was the instances flag moved to the runProxy() function?

enocom: It's a side effect of how `flag.Var` works in effect. We need a type that implements the `Value` interface, so we can override the `Set` behavior. Since `flag.Var` doesn't return a value, moving the definition was a reasonable approach.


https://pkg.go.dev/flag#Var

akinfermo: Thanks for the clarification, Eno!

PR #1048, opened 2021-12-10
fix: increase rateLimit burst size to 2

enocom:


proxy/proxy/client.go
```diff
@@ -544,6 +544,7 @@ func (c *Client) invalidateCfg(cfg *tls.Config, instance string) {
 		done:          e.done,
 		lastRefreshed: e.lastRefreshed,
 	}
+	c.limiters[instance] = nil
```
kurtisvg, line 547: I don't think this is a change we want to make. I think that even (especially?) in the event of a failure the ratelimiter should be active to make sure we don't excessively retry.

kurtisvg (changes requested):

proxy/proxy/client.go
enocom: I think that means that when a new dial attempt fails on invalid config, they'd have to wait 30 seconds to try again. In other words, I think invalid config shouldn't be cause to make a caller wait before trying again.

kurtisvg: We're using rate.limiter, which means it's supposed to allow a "surge" of a X amount as long as the average rate doesn't exceed the time provided: https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/08f42cabbd0b441bb020d7c56b861f8048f07156/proxy/proxy/client.go#L420

So I think what we really need to do is up the bucket size so that we can "burst" several refreshes in a row and then slow the role after that (if that many fail there is a different problem)

enocom: That's a better idea. I'll do that.

kurtisvg (approved):

kurtisvg (approved):

PR #1053, opened 2021-12-14
chore: add integration test for resetting SSL

enocom:


enocom:
This test will fail until rebased onto https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/1048.

tests/dialer_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	defer tx.Rollback()
+
+	resetSSL(client)
+
+	// Re-dial twice, once to invalidate config, once to establish connection
+	var attempts int
+	for {
+		t.Log("Re-dialing instance")
+		_, err = proxyClient.DialContext(context.Background(), os.Getenv("POSTGRES_CONNECTION_NAME"))
```
kurtisvg, line 120: Should this env var be a reference to an existing var? I think you can pass these in via flags as well

kurtisvg (approved):

tests/dialer_test.go
enocom: Yes, that's a good point.

PR #1058, opened 2021-12-22
fix: improve logging and cleanup

kurtisvg:
## Change Description

Update it so we're using `cmd.Print...` everywhere over `log.Print...`, and cleaning up some roughly related cleanup.

### DIFF
--- a/cmd/proxy.go
+++ b/cmd/proxy.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


cmd/proxy.go
```diff
@@ -134,18 +102,41 @@ func (s *socketMount) serve(ctx context.Context) error {
 		}
 		// handle the connection in a separate goroutine
 		go func() {
-			log.Printf("[%s] accepted connection from %s\n", s.inst, cConn.RemoteAddr())
-			sConn, err := s.dialer.Dial(ctx, s.inst)
+			pc.cmd.Printf("[%s] accepted connection from %s\n", s.inst, cConn.RemoteAddr())
+
+			// give a max of 30 seconds to connect to the instance
+			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
enocom, line 108: What's the relationship between this context and the on passed in above? Should we be building the new context here with the one passed in?

enocom (approved):

PR #1068, opened 2022-01-10
fix: invalidated config should retain error

### LINKED ISSUE #1065: Panic in liveness endpoint
## Bug Description

When this google cloud incident started https://status.cloud.google.com/incidents/NMcnk6aE8xMHHwRGmyry, our cloudsql proxy started being killed by liveness probes. They are configured according to this recommendation https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/main/examples/k8s-health-check#cloud-sql-proxy-health-checks. That's fine and expected, but the error message was very confusing:

```
[Health Check] Liveness failed: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)
```

There seems to be a panic somewhere in cloudsql proxy, but no more information was printed.

## Environment

1. OS type and version: GKE
2. Cloud SQL Proxy version (`./cloud_sql_proxy -version`): gcr.io/cloudsql-docker/gce-proxy:1.27.0-buster@sha256:5d3d9bb25542b5f786cdb983d052b5f9145c9abd0b7128688f545ae9ce060fca

enocom:
When a TLS handshake fails, the client would invalidate the
configuration without recording the associated error. When the liveness
check runs, it would panic when trying to print the invalidated
configuration's error because the error was nil. This commit ensures
that when the proxy invalidates a configuration, it includes the error.

Fixes #1065.

### DIFF
--- a/proxy/proxy/connect_tls_117.go
+++ b/proxy/proxy/connect_tls_117.go
@@ -38,7 +38,7 @@ func (c *Client) connectTLS(
 	// this file is conditionally compiled on only Go versions >= 1.17.
 	if err := ret.HandshakeContext(ctx); err != nil {
 		_ = ret.Close()
-		c.invalidateCfg(cfg, instance)
+		c.invalidateCfg(cfg, instance, err)
 		return nil, err
 	}
 	return ret, nil
--- a/proxy/proxy/connect_tls_117.go
+++ b/proxy/proxy/connect_tls_117.go
@@ -20,6 +20,7 @@ package proxy
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 )
 
@@ -38,6 +39,7 @@ func (c *Client) connectTLS(
 	// this file is conditionally compiled on only Go versions >= 1.17.
 	if err := ret.HandshakeContext(ctx); err != nil {
 		_ = ret.Close()
+		err = fmt.Errorf("config invalidated after TLS handshake failed, error = %w", err)
 		c.invalidateCfg(cfg, instance, err)
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


proxy/proxy/connect_tls_117.go
```diff
@@ -38,7 +38,7 @@ func (c *Client) connectTLS(
 	// this file is conditionally compiled on only Go versions >= 1.17.
 	if err := ret.HandshakeContext(ctx); err != nil {
 		_ = ret.Close()
-		c.invalidateCfg(cfg, instance)
+		c.invalidateCfg(cfg, instance, err)
```
kurtisvg, line 41: Let's add some context here on that this because of an invalidated config

kurtisvg (changes requested):

proxy/proxy/connect_tls_117.go
```diff
@@ -38,7 +39,8 @@ func (c *Client) connectTLS(
 	// this file is conditionally compiled on only Go versions >= 1.17.
 	if err := ret.HandshakeContext(ctx); err != nil {
 		_ = ret.Close()
-		c.invalidateCfg(cfg, instance)
+		err = fmt.Errorf("config invalidated after TLS handshake failed, error = %w", err)
```
kurtisvg, line 42: nit: consider moving the wrapping to add context ("config invalidated because error:") inside the invalidate function

kurtisvg (approved):

proxy/proxy/connect_tls_117.go
enocom: Good idea. Should have thought of that myself.

PR #1081, opened 2022-01-25
chore: update proxy version in all docs

enocom:


kurtisvg (approved):
We should open a bug to track automating this

PR #1104, opened 2022-02-17
chore: stop updating google.golang.org/api

enocom:
As of https://github.com/googleapis/google-api-go-client/pull/1428 and v0.69.0, the minimum Go version is Go 1.15.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,7 +12,8 @@
     "gomodTidy"
   ],
   "ignoreDeps": [
-      "golang.org/x/net"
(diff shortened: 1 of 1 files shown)


kurtisvg (approved):
Two open questions:
1. How are we going to remember to unfreeze this? 
2. Do we need to reconsider our minimum version if they go client is jumping to 1.15?

enocom:
As for unfreezing this, if we support only the latest four versions, we're just waiting for Go 1.18 to unfreeze this.

PR #1105, opened 2022-02-18
docs: update the install from source instructions

mariosnikolaou:
Latest Go versions use "go install" instead of "go get".


enocom (approved):
Thanks for this!

PR #1127, opened 2022-03-07
fix: print usage only on erroneous invocation

enocom:


enocom:
Here's what happens without this change:

```
$ ./cloudsql-proxy my-project:us-central1:pg my-project:us-central1:mysql
[my-project:us-central1:pg] Listening on 127.0.0.1:5000
[my-project:us-central1:mysql] Listening on 127.0.0.1:5001
The proxy has started successfully and is ready for new connections!
^CSIGINT signal received. Shuting down...
Error: SIGINT signal received
Usage:
  cloud_sql_proxy instance_connection_name... [flags]

Flags:
  -h, --help   help for cloud_sql_proxy

```

cmd/root.go
```diff
@@ -46,12 +46,21 @@ func New() *cobra.Command {
...
+		RunE:         runSignalWrapper,
 	}
 }
 
 // runSignalWrapper watches for SIGTERM and SIGINT and interupts execution if necessary.
 func runSignalWrapper(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		cmd.Help()
+		os.Exit(1)
+	}
```
kurtisvg, line 63: So I agree we don't always want the usage, but I think that only printing it when no args are set isn't maybe the best idea either. Passing an invalid arg will no longer trigger the help message here. 

Take a look at [this comment ](https://github.com/spf13/cobra/issues/340#issuecomment-374617413) - they recommended setting "SilenceUsuage" from within RunE after the args are parsed to avoid having it print for errors once a certain point is reached.

cmd/root.go
```diff
@@ -46,12 +46,21 @@ func New() *cobra.Command {
...
+		SilenceUsage: true,
+		RunE:         runSignalWrapper,
 	}
 }
 
 // runSignalWrapper watches for SIGTERM and SIGINT and interupts execution if necessary.
 func runSignalWrapper(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		cmd.Help()
+		os.Exit(1)
```
kurtisvg, line 62: I think `os.Exit(1)` will cause problems when we are testing because it'll attempt to exit the Go environment.

kurtisvg (changes requested):

cmd/root.go
enocom: Nice find. I'll update with that approach.

cmd/root.go
```diff
@@ -52,6 +52,9 @@ any client SSL certificates.`,
 
 // runSignalWrapper watches for SIGTERM and SIGINT and interupts execution if necessary.
 func runSignalWrapper(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return errInvalidInvocation
+	}
```
kurtisvg, line 57: So we can leave this for now, but I think we'll probably end up yanking it out at some point when things like "fuse" or even config files are added since args will no longer be required.

cmd/root.go
```diff
@@ -110,5 +113,6 @@ func runSignalWrapper(cmd *cobra.Command, args []string) error {
 	default:
 		cmd.PrintErrf("The proxy has encountered a terminal error: %v\n", err)
 	}
+	cmd.SilenceUsage = true
```
kurtisvg, line 116: So my hot take is that probably needs to happen in `newProxyClient()`, after we've parsed the instance connection string. Otherwise things like mount errors will cause the usage to print when maybe the don't need too. Thoughts?

PR #1128, opened 2022-03-08
feat: add support for address (and a) flag

enocom:


kurtisvg (changes requested):
High level thoughts:

1. I think we should look at using [viper](https://github.com/spf13/viper) if we can, because it'll make many feature requests (like config files) way easier later on
2. We had talked about the port using "auto" by default to default to an instance specific IP (for example first mysql instance = 3306, second mysql = 3307, postgres = 5432). Do you think we should abandon that approach and just pick a static port?

enocom:
Re-purposing this PR to be for the `address` flag. We'll revisit the port flag in a future PR. This commit sets the stage to easily add Viper support.

To add new flags the process looks like this:

1. Declare a new variable (e.g., `flagFoo`)
2. Add a `PersistantFlag` onto the root command object.
3. Assign the variable into the `proxy.Config` type.
4. Use the new field inside `proxy.NewClient`

kurtisvg (approved):

PR #1129, opened 2022-03-08
fix: don't send on a closed channel

enocom:
When lauching the proxy with multiple instances, a second goroutine
would attempt to send on a closed channel. This commit resolves the
problem by leaving the channel open and preventing a goroutine leak by
attempting a best effort send.

### DIFF
--- a/cmd/proxy.go
+++ b/cmd/proxy.go
@@ -67,7 +67,11 @@ func (pc *proxyClient) serve(ctx context.Context) error {
 			err := pc.serveSocketMount(ctx, mnt)
 			if err != nil {
 				select {
-				// Best effort attempt to send error
+				// Best effort attempt to send error.
+				// If this send fails, it means the reading goroutine has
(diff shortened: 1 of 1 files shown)


cmd/proxy.go
```diff
@@ -62,13 +62,16 @@ func (pc *proxyClient) serve(ctx context.Context) error {
...
 	exitCh := make(chan error)
-	defer close(exitCh)
 	for _, m := range pc.mnts {
 		go func(mnt *socketMount) {
 			err := pc.serveSocketMount(ctx, mnt)
 			if err != nil {
-				exitCh <- err
-				return
+				select {
+				// Best effort attempt to send error
```
kurtisvg, line 70: nit: I'd consider extending this comment a bit more to clarify that the only time this doesn't send "instantly" if an error has already been sent, and (I think) we only care about the first one.

kurtisvg (approved):

PR #1130, opened 2022-03-08
fix: close proxy client after mount failures

enocom:


cmd/proxy.go
```diff
@@ -48,7 +48,7 @@ func newProxyClient(ctx context.Context, cmd *cobra.Command, args []string) (*pr
 		m := &socketMount{inst: inst}
 		addr, err := m.listen(ctx, "tcp4", net.JoinHostPort("127.0.0.1", fmt.Sprint(port+i)))
 		if err != nil {
-			return nil, fmt.Errorf("[%s] Unable to mount socket: %v", inst, err)
+			return pc, fmt.Errorf("[%s] Unable to mount socket: %v", inst, err)
```
kurtisvg, line 51: IMO, we are asking for trouble to sometimes return `nil` and sometimes return an object when `err != nil`. Can we just close pc here instead?

kurtisvg (changes requested):

kurtisvg (approved):

PR #1139, opened 2022-03-17
feat: add port flag and query param

enocom:


cmd/root.go
```diff
@@ -104,16 +107,35 @@ func parseConfig(conf *proxy.Config, args []string) error {
 			if err != nil {
 				return newBadCommandError(fmt.Sprintf("could not parse query: %q", res[1]))
 			}
-			if len(q["address"]) != 1 {
-				return newBadCommandError(fmt.Sprintf("address query param should be only one value: %q", q["address"]))
+
+			a, ok := q["address"]
+			if ok {
```
kurtisvg, line 112: nit: should this be a single line?

kurtisvg (approved):

cmd/root.go
enocom: Yes. I'll fix below too.

PR #1140, opened 2022-03-17
feat: add support for automatic port resolution

enocom:
This commit adds support for resolving a Cloud SQL instance's port based
on its database engine. Postgres instances will bind to 5432, MySQL to
3306, and SQL Server to 1433. Subsequent bindings will increment from
the appropriate database engine port (e.g., 5432, 5433, 5434, etc).

This behavior may be disabled by using either the port flag, or by
specifying a port query parameter on a per instance basis.


internal/proxy/mock_test.go
```diff
@@ -0,0 +1,315 @@
+// Copyright 2022 Google LLC
```
enocom, line 1: This file is largely the same from the Go connector. I've dropped what we don't use and added support for specifying the DB engine.

internal/proxy/proxy.go
```diff
@@ -60,38 +61,83 @@ type Client struct {
...
 		a := conf.Addr
 		if inst.Addr != "" {
 			a = inst.Addr
 		}
-		nextPort := port + inc
-		if inst.Port != 0 {
-			nextPort = inst.Port
-		} else {
-			inc++
+		version, err := d.EngineVersion(ctx, inst.Name)
```
enocom, line 117: We could do this work in a goroutine, so we mount faster.

cmd/root.go
```diff
@@ -172,7 +173,13 @@ func runSignalWrapper(cmd *Command) error {
 	startCh := make(chan *proxy.Client)
 	go func() {
 		defer close(startCh)
-		p, err := proxy.NewClient(ctx, cmd.Command, cmd.conf)
+		d, err := cloudsqlconn.NewDialer(ctx)
+		if err != nil {
+			defer d.Close()
```
kurtisvg, line 178: Should we close the dialer if err != nil? Doesn't that mean that dialer = nil?

internal/proxy/mock_test.go
kurtisvg: Is it really beneficial for us to mock the entire admin API here - would it be better to try to mock the "Dialer" part itself? We don't need to test the Dialer functionality, but can we build a similar "mock" style thing here, but for the Dialer instead of the admin api?

internal/proxy/proxy.go
```diff
@@ -60,38 +61,113 @@ type Client struct {
...
+			}
+			addr, err := m.listen(ctx, "tcp", net.JoinHostPort(a, fmt.Sprint(np)))
+			if err != nil {
+				err := fmt.Errorf("[%v] Unable to mount socket: %v", i.Name, err)
+				mntCh <- mountRes{err: err}
+				return
+			}
+			cmd.Printf("[%s] Listening on %s\n", i.Name, addr.String())
+			mntCh <- mountRes{mnt: m}
+		}(inst)
```
kurtisvg, line 149: This bit concerns me a bit, because I think it makes the port used non-deterministic. This could cause the "auto" port behavior to be inconsistent, which I don't think we want. 

I think what we mostly want to do is run `d.EngineVersion` in parallel - can we just do a quick for loop to "warm up" the call (and disregard the result, we just want it to start the refresh), and then when we call it below it'll block until ready?

internal/proxy/proxy.go
```diff
@@ -60,38 +61,113 @@ type Client struct {
...
-		if inst.Port != 0 {
-			nextPort = inst.Port
-		} else {
-			inc++
+		go func(i InstanceConnConfig) {
+			m := &socketMount{inst: i.Name}
+			a := conf.Addr
+			if i.Addr != "" {
+				a = i.Addr
+			}
```
kurtisvg, line 126: nit: Is it possible to set this when the query string is parsed? Something like "if not ok, set default value"?

kurtisvg (changes requested):

cmd/root.go
enocom: I don't know what I was thinking. I'll remove this.

internal/proxy/proxy.go
enocom: Sounds good.

internal/proxy/mock_test.go
enocom: No. We can wrap the dialer itself in an interface. Just wanted to go the full way and see how it looked, and I agree it seems unnecessary.

internal/proxy/proxy.go
enocom: As in, when we don't see an address query string, we should just parse the instance's address as the default value? I think that would remove the need for a top level address field.

internal/proxy/proxy.go
```diff
@@ -54,44 +55,110 @@ type Config struct {
...
+		// Unexpected engine version, use global port setting instead.
+		return c.nextPort()
 	}
-	c := &Client{cmd: cmd, dialer: d}
+}
 
-	port := conf.Port
-	var inc int
+// DBDialer is a dialer for a Cloud SQL instance.
+type DBDialer interface {
```
kurtisvg, line 108: nit: Should it be DbDialer?

internal/proxy/proxy.go
```diff
@@ -54,44 +55,110 @@ type Config struct {
...
+	var wg sync.WaitGroup
+	for _, inst := range conf.Instances {
+		wg.Add(1)
+		go func(i InstanceConnConfig) {
+			// Initiate refresh operation
+			d.EngineVersion(ctx, inst.Name)
+			wg.Done()
+		}(inst)
+	}
+	wg.Wait()
```
kurtisvg, line 131: I _think_ (unless I'm missing something) that the waitgroup is overkill - calls to EngineVersion will block until the refresh is done, so I don't think we need to wait for them all to complete before referencing.

kurtisvg (approved):

internal/proxy/proxy.go
enocom: DBD is a bit hard to read. We could just call it a Dialer.

kurtisvg: SGTM

internal/proxy/proxy.go
enocom: I'd like to hold off on this change for now. We're checking the port's global config vs instance config in this same function, and for now keeping this similar code together seems preferable, but I could be wrong.

internal/proxy/proxy.go
enocom: Its only merit that I can think of is to communicate intention (i.e., when mounting the sockets, we want immediate return values from the call). But thinking about the performance, I agree the wait group doesn't do anything better.

kurtisvg: I think in that case, it's better to just leave a comment explaining the intent rather than rely on the wg here.

enocom: The wait group just means the waiting is consolidated to `wg.Wait`. Whereas without it, we will spend some time waiting for `d.EngineVersion` to return on its second invocation.

internal/proxy/proxy.go
kurtisvg: SGTM, we can punt this until later.

PR #1149, opened 2022-03-29
feat: add support for token-based auth

enocom:


internal/proxy/proxy_test.go
```diff
@@ -179,3 +179,30 @@ func TestClientInitialization(t *testing.T) {
...
+			config:  proxy.Config{Token: "my-token"},
+			wantLen: 1,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			if got := tc.config.DialerOpts(); tc.wantLen != len(got) {
+				t.Errorf("want len = %v, got = %v", tc.wantLen, len(got))
+			}
```
kurtisvg, line 205: This test doesn't feel like it's doing a particularly thorough job - it's checking the length but not the contents? I'm a little concerned it might be too brittle 

Would this be a better candidate for an e2e test?

kurtisvg (approved):

internal/proxy/proxy_test.go
enocom: I agree -- this is a feeble test. I'll add an e2e test in the next PR and remove this.

PR #1151, opened 2022-03-29
feat: add support for credentials file

enocom:


.envrc.example
```diff
@@ -15,3 +15,5 @@ export SQLSERVER_CONNECTION_NAME="project:region:instance"
 export SQLSERVER_USER="sqlserver-user"
 export SQLSERVER_PASS="sqlserver-password"
 export SQLSERVER_DB="sqlserver-db-name"
+
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json
```
kurtisvg, line 19: Does this actually need to be set, or can we fake this env var for that specific test?

cmd/root.go
```diff
@@ -79,6 +79,8 @@ any client SSL certificates.`,
 	// Global-only flags
 	cmd.PersistentFlags().StringVarP(&c.conf.Token, "token", "t", "",
 		"Bearer token used for authorization.")
+	cmd.PersistentFlags().StringVarP(&c.conf.CredentialsFile, "credentials-file", "c", "",
```
kurtisvg, line 82: does "c" make sense here, or should it be "cf"?

internal/proxy/proxy.go
```diff
@@ -58,10 +61,20 @@ type Config struct {
...
+	// 3. gcloud auth
+	switch {
+	case c.Token != "":
 		opts = append(opts, cloudsqlconn.WithTokenSource(
 			oauth2.StaticTokenSource(&oauth2.Token{AccessToken: c.Token}),
 		))
+	case c.CredentialsFile != "":
+		opts = append(opts, cloudsqlconn.WithCredentialsFile(
+			c.CredentialsFile,
+		))
```
kurtisvg, line 77: I think we should throw an error if the user specifies both. Silently ignoring one makes it harder to understand what's wrong.

kurtisvg: It would also be good if we can log which one we are using on start.

testsV2/postgres_test.go
```diff
@@ -56,5 +59,55 @@ func TestPostgresTcp(t *testing.T) {
...
+		"postgres", dsn, postgresPort, "")
+}
+
+func TestAuthWithCredentialsFile(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping Postgres integration tests")
+	}
+	requirePostgresVars(t)
+
+	path, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS")
```
kurtisvg, line 94: Is it possible that maybe we can fake this one somehow? Like write a fake service account, set the env var just for this test, and check the header on our mocks? 

This is one integration test I'd prefer not to have given it means we have to have a service account key laying around.

kurtisvg (changes requested):

cmd/root.go
enocom: cf sounds good to me.

testsV2/postgres_test.go
enocom: When we add a custom dialer, we could remove this. Until then, we don't have a way to inject a custom dialer, or HTTP client.

internal/proxy/proxy.go
enocom: I've put logging here, and moves the check for both values up into `parseConfig` -- that's where other argument checking is happening.

.envrc.example
enocom: For the e2e test to work, we'll need this.

enocom: It's also already present in our CI environment.

testsV2/postgres_test.go
enocom: So I was wrong: a custom dialer does not make this easy to test, as setting up a dialer also entails setting up authentication and so skips the code path where we append dialer options.

I also tried to remove the e2e test in favor of a more targeted unit test. That test was around the proxy.Client and involved setting up a mock HTTP server. This approach worked well for testing that the token is added to the header (I could just check the actual headers sent to the HTTP server).

For a test of the credentials file, however, the unit test works less well, as we need to also put together a fake OAuth2 server. When we pass a credential file dialer option, the underlying library reads the file and then tries to fetch a token at the URL specified in the credential file. While it is totally possible to assemble a fake OAuth2 server that returns tokens, I think the final resulting test depends on too many moving pieces, exercises a code path just below the top level, and ultimately doesn't provide as good test coverage.

So in the end, although I'd love to remove the service account dependency, a good and reliable test for this needs to be end to end. This is exactly the kind of feature that if it broke, it would have a very large impact on our customers, and so I want a robust test that guarantees the CLI flag, the parsing of args, the dialer options, and the ultimate authentication works. 

As a side note, if we switch to WIF, the [auth action](https://github.com/google-github-actions/auth) does support adding a `GOOGLE_APPLICATION_DEFAULT` credentials environment variable. So these tests would run in that setup as well.

cmd/root.go
enocom: I like the single letter if only to match what I imagine most CLIs do -- long phrase for a flag and a corresponding single letter. So far, I've been just using whatever the long phrases starts with to pick a single letter.

testsV2/mysql_test.go
```diff
@@ -58,5 +59,9 @@ func TestMysqlTcp(t *testing.T) {
 		Addr:                 "127.0.0.1:3306",
 		Net:                  "tcp",
 	}
-	proxyConnTest(t, *mysqlConnName, "mysql", cfg.FormatDSN(), mysqlPort, "")
+	ctx, cancel := context.WithTimeout(context.Background(), connTestTimeout)
+	defer cancel()
+	proxyConnTest(t,
+		ctx,
```
enocom, line 65: Adding this context allows our tests to print a better error message when a CLI flag isn't wired up correctly. Otherwise the test just hangs indefinitely when a CLI flag is missing.

internal/proxy/proxy.go
```diff
@@ -61,18 +64,28 @@ type Config struct {
...
+	Error(msg string, data ...interface{})
 }
 
-func (c *Config) DialerOpts() []cloudsqlconn.Option {
+func (c *Config) DialerOpts(l Logger) []cloudsqlconn.Option {
 	var opts []cloudsqlconn.Option
-	if c.Token != "" {
+	switch {
+	case c.Token != "":
+		l.Log("Using provided token for authentication\n")
```
kurtisvg, line 80: nit: Can we use the pattern 

- "Authorizing with the `-token` flag"
- "Authorizing with the credential file at %q"

we should also probably log a "Authorizing using the Application Default credentials" if nothing is passed.

internal/proxy/proxy.go
```diff
@@ -61,18 +64,28 @@ type Config struct {
...
+// messages.
+type Logger interface {
+	// Log reports an informational message.
+	Log(msg string, data ...interface{})
+	// Error reports an error message.
+	Error(msg string, data ...interface{})
 }
 
-func (c *Config) DialerOpts() []cloudsqlconn.Option {
+func (c *Config) DialerOpts(l Logger) []cloudsqlconn.Option {
```
kurtisvg, line 76: Does it make sense to do the logging here? Shouldn't we move it to where the parsing/config is set?

testsV2/connection_test.go
```diff
@@ -18,14 +18,13 @@ import (
...
 
-// proxyConnTest is a test helper to verify the proxy works with a basic connectivity test.
-func proxyConnTest(t *testing.T, connName, driver, dsn string, port int, dir string) {
-	ctx := context.Background()
-
-	args := []string{connName}
+const connTestTimeout = time.Minute
 
+// proxyConnTest is a test helper to verify the proxy works with a basic connectivity test.
+func proxyConnTest(t *testing.T, ctx context.Context, args []string, driver, dsn string, port int, dir string) {
```
kurtisvg, line 27: Why did we move ctx/args out of this function into the callers? The point of this function is to avoid writing the same test 3 times, but now we've just done that with ctx.

testsV2/mysql_test.go
kurtisvg: I don't have a problem with adding a timeout, but think it can be done in proxyConnTest.

cmd/root.go
```diff
@@ -177,6 +184,18 @@ func parseConfig(conf *proxy.Config, args []string) error {
...
+	*cobra.Command
+}
+
+func (c cmdLogger) Log(msg string, data ...interface{}) {
+	c.Printf(msg, data...)
+}
+
+func (c cmdLogger) Error(msg string, data ...interface{}) {
+	c.PrintErrf(msg, data...)
+}
```
kurtisvg, line 197: I'm not sure I follow why we are adding this now.

cmd/root.go
```diff
@@ -129,6 +131,11 @@ func parseConfig(conf *proxy.Config, args []string) error {
 		return newBadCommandError(fmt.Sprintf("not a valid IP address: %q", conf.Addr))
 	}
 
+	// If both token and credentials file were set, error.
+	if conf.Token != "" && conf.CredentialsFile != "" {
+		return newBadCommandError("ambiguous auth configuration: both token and credentials file were set")
```
kurtisvg, line 136: nit: suggest wording like "Cannot specify `--token` and `--credentials-file` flags at the same time". 

There are likely going to be a few more cases of conflicting flags, we should just be consistent

kurtisvg (changes requested):

.gitignore
```diff
@@ -8,3 +8,7 @@
 # Compiled binary
 /cmd/cloud_sql_proxy/cloud_sql_proxy
 /cloud_sql_proxy
+# v2 binary
+/cloudsql-proxy
```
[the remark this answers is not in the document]
enocom: Yes, I agree. We'll need to update the build scripts to output `cloud_sql_proxy`. This is for local dev-flow convenience. If I run `go build` it will use the module name `cloudsql-proxy`. Granted, I could add the `-o cloud_sql_proxy` locally, but `go build` is much easier. And so I add this line here to avoid committing the dev artifact.

cmd/root.go
enocom: Removed this in favor of logging in parseConfig.

internal/proxy/proxy.go
enocom: Moved this up a level.

internal/proxy/proxy.go
enocom: Nice. Done.

.gitignore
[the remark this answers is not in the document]
kurtisvg: FYI you can just make a local-only gitignore: https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

kurtisvg (changes requested):
I think you need to push some changes :)

testsV2/postgres_test.go
```diff
@@ -56,5 +60,59 @@ func TestPostgresTcp(t *testing.T) {
...
+	}
+	if err := os.Unsetenv("GOOGLE_APPLICATION_CREDENTIALS"); err != nil {
+		t.Fatalf("failed to unset GOOGLE_APPLICATION_CREDENTIALS")
+	}
+	return tok, path, func() {
+		os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", path)
+	}
+}
+
+func TestAuthWithToken(t *testing.T) {
```
kurtisvg, line 90: Are these postgres specific tests or should we be testing them more widely?

testsV2/postgres_test.go
```diff
@@ -56,5 +60,59 @@ func TestPostgresTcp(t *testing.T) {
...
 
 	dsn := fmt.Sprintf("user=%s password=%s database=%s sslmode=disable", *postgresUser, *postgresPass, *postgresDb)
-	proxyConnTest(t, *postgresConnName, "postgres", dsn, postgresPort, "")
+	proxyConnTest(t, []string{*postgresConnName}, "postgres", dsn, postgresPort, "")
+}
+
+// removeAuthEnvVar retrieves an OAuth2 token and a path to a service account key
+// and then unsets GOOGLE_APPLICATION_CREDENTIALS. It returns a cleanup function
+// that restores the original setup.
+func removeAuthEnvVar(t *testing.T) (*oauth2.Token, string, func()) {
```
kurtisvg, line 69: Is this necessary for the other 2 tests? if we specific the flag, it should override the GOOGLE_APPLICATION_CREDENTIALS

testsV2/postgres_test.go
enocom: Ultimately, we need to connect to a database, but which database doesn't matter. Can you think of a better place to put these?

testsV2/postgres_test.go
enocom: I've removed the GOOGLE_APPLICATION_CREDENTIALS for each auth test, to ensure pass for the correct reason. If we retain the envvar, in theory the test could pass but be using the ADC.

.gitignore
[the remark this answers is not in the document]
enocom: For peculiar dev workflows, I think a local-only gitignore is perfectly reasonable. Out of the box, a `go build` on the v2 branch will produce this exact binary. So I think we should ignore it globally. What if I added a comment to that effect, e.g., "Ignore the binary produced by building the module"?

testsV2/postgres_test.go
enocom: I've added tests to all three engines.

kurtisvg (approved):

PR #1158, opened 2022-04-01
feat: add support for a custom dialer

enocom:


cmd/root.go
```diff
@@ -49,11 +50,25 @@ type Command struct {
 	conf *proxy.Config
 }
 
+// Option is a function that configures a Command.
+type Option func(*Command)
```
kurtisvg, line 54: nit: Should this be a `Command`, or should it be `proxy.Config{}`?

kurtisvg (approved):

enocom:
Previous builds have failed on port in-use errors, so I've updated this to allow overriding the default ports.

kurtisvg (approved):

PR #1161, opened 2022-04-01
feat: add --version flag

kurtisvg:
Add support for `--version`.


cmd/root.go
```diff
@@ -178,7 +192,8 @@ func runSignalWrapper(cmd *Command) error {
 	startCh := make(chan *proxy.Client)
 	go func() {
 		defer close(startCh)
-		d, err := cloudsqlconn.NewDialer(ctx, cmd.conf.DialerOpts()...)
+		opts := append(cmd.conf.DialerOpts(), cloudsqlconn.WithUserAgent(userAgent))
```
enocom, line 195: Would it be worth adding a test for this, or perhaps adding the user agent to the proxy config, such that `DialerOpts` returns it?

enocom (approved):

cmd/root.go
kurtisvg: I don't think we'll have a good way to test without https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/157

PR #1165, opened 2022-04-07
docs: add note about redirecting logs to stdout

### LINKED ISSUE #1143: Logs show up as errors when deploying the sidecar example
When running the proxy as sidecar in GKE as described in the example, all logs written by the proxy are flagged as errors in Cloud Logging. This can be prevented by using the flag `-structured_logs`. In my opinion it would be helpful for others to add some additional notes about this behaviour.

Relevant snippets from the [example](https://github.com/GoogleCloudPlatform/cloudsql-proxy/tree/main/examples/k8s-sidecar) and [Google Cloud Doc](https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine#run_the_as_a_sidecar):

```yaml
  - name: cloud-sql-proxy
    # It is recommended to use the latest version of the Cloud SQL proxy
    # Make sure to update on a regular schedule!
    image: gcr.io/cloudsql-docker/gce-proxy:1.28.0 # make sure the use the latest version
    command:
      - "/cloud_sql_proxy"

      # If connecting from a VPC-native GKE cluster, you can use the
      # following flag to have the proxy connect over private IP
      # - "-ip_address_types=PRIVATE"

      # Replace DB_PORT with the port the proxy should listen on
      # Defaults: MySQL: 3306, Postgres: 5432, SQLServer: 1433
      - "-instances=<INSTANCE_CONNECTION_NAME>=tcp:<DB_PORT>"

    # [START cloud_sql_proxy_k8s_volume_mount]
      # This flag specifies where the service account key can be found
      - "-credential_file=/secrets/service_account.json"
 ```

### PR DESCRIPTION
**Pull request by enocom**
Fixes #1143.

### DIFF
--- a/examples/k8s-sidecar/proxy_with_sa_key.yaml
+++ b/examples/k8s-sidecar/proxy_with_sa_key.yaml
@@ -59,7 +59,7 @@ spec:
           # environments, anything printed to stderr is consider an error. To
           # disable this behavior and write all logs to stdout (except errors
           # which will still go to stderr), use:
-          # - "-log_debug_stdout"
+          - "-log_debug_stdout"
 
           # Replace DB_PORT with the port the proxy should listen on
           # Defaults: MySQL: 3306, Postgres: 5432, SQLServer: 1433
--- a/examples/k8s-sidecar/proxy_with_sa_key.yaml
+++ b/examples/k8s-sidecar/proxy_with_sa_key.yaml
@@ -55,6 +55,12 @@ spec:
           # following flag to have the proxy connect over private IP
           # - "-ip_address_types=PRIVATE"
 
+          # By default, the proxy will write all logs to stderr. In some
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: examples/k8s-sidecar/proxy_with_sa_key.yaml [hunk 1]
```diff
@@ -55,6 +55,12 @@ spec:
           # following flag to have the proxy connect over private IP
           # - "-ip_address_types=PRIVATE"
 
+          # By default, the proxy will write all logs to stderr. In some
+          # environments, anything printed to stderr is consider an error. To
+          # disable this behavior and write all logs to stdout (except errors
+          # which will still go to stderr), use:
+          # - "-log_debug_stdout"
```
kurtisvg, line 62: nit: I'd suggest just uncommenting this - it's probably the behavior we want to suggest

kurtisvg (approved):

PR #1166, opened 2022-04-07
feat: add support for gcloud auth 

enocom:


cmd/root.go
```diff
@@ -227,8 +237,12 @@ func runSignalWrapper(cmd *Command) error {
 		// Otherwise, initialize a new one.
 		d := cmd.conf.Dialer
 		if d == nil {
-			opts := append(cmd.conf.DialerOpts(), cloudsqlconn.WithUserAgent(userAgent))
-			var err error
+			opts, err := cmd.conf.DialerOpts()
+			if err != nil {
+				shutdownCh <- fmt.Errorf("error initializing dialer: %v", err)
+				return
+			}
```
kurtisvg, line 244: I'm feeling a bit iffy about returning an error here for some reason - we already have a switch statement on L147. Is there any reason why we shouldn't just set the DialerOpts (including the gcloud token source) after we log that we are using said source? ?

internal/gcloud/gcloud.go
```diff
@@ -0,0 +1,133 @@
...
+	GcloudOk GcloudStatusCode = iota
+	GcloudNotFound
+	// generic execution failure error not specified above.
+	GcloudExecErr
+)
+
+type GcloudError struct {
+	GcloudError error
+	Status      GcloudStatusCode
+}
```
kurtisvg, line 65: Should we update this to match our other errors with an exit code?

internal/gcloud/gcloud.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+type GcloudStatusCode int
+
+const (
+	GcloudOk GcloudStatusCode = iota
+	GcloudNotFound
+	// generic execution failure error not specified above.
+	GcloudExecErr
```
kurtisvg, line 59: I feel like these might be better as different errors rather than a single error type with a status code. The status is not exposed to the user (or helpful in determining what went wrong).

testsV2/postgres_test.go
```diff
@@ -86,3 +90,49 @@ func TestPostgresAuthWithCredentialsFile(t *testing.T) {
...
+	}
+	requirePostgresVars(t)
+
+	// The following configures gcloud using only GOOGLE_APPLICATION_CREDENTIALS.
+	configureGcloud := func(t *testing.T) func() {
+		dir, err := ioutil.TempDir("", "cloudsdk*")
+		if err != nil {
+			t.Fatalf("failed to create temp dir: %v", err)
+		}
+		os.Setenv("CLOUDSDK_CONFIG", dir)
```
kurtisvg, line 106: Can we add a comment for why we need to do this?

internal/gcloud/gcloud_test.go
```diff
@@ -0,0 +1,80 @@
...
+		t.Skip("skipping gcloud integration tests")
+	}
+
+	configureGcloud := func(t *testing.T) func() {
+		// The following configures gcloud using only GOOGLE_APPLICATION_CREDENTIALS.
+		dir, err := ioutil.TempDir("", "cloudsdk*")
+		if err != nil {
+			t.Fatalf("failed to create temp dir: %v", err)
+		}
+		os.Setenv("CLOUDSDK_CONFIG", dir)
```
kurtisvg, line 39: Can we beef up this comment a bit - what does "CLOUDSDK_CONFIG" have to do with any of it?

kurtisvg (changes requested):

cmd/root.go
```diff
@@ -110,6 +110,8 @@ any client SSL certificates.`,
 		"Bearer token used for authorization.")
 	cmd.PersistentFlags().StringVarP(&c.conf.CredentialsFile, "credentials-file", "c", "",
 		"Path to a service account key to use for authentication.")
+	cmd.PersistentFlags().BoolVarP(&c.conf.GcloudAuth, "gcloud-auth", "g", false,
+		"Use gcloud's config helper to retrieve a token for authentication.")
```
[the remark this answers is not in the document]
enocom: I've updated this string to note that we're using the gcloud config's user credentials.

cmd/root.go
```diff
@@ -131,16 +133,24 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+		return newBadCommandError("ambiguous auth configuration: both credentials file and gcloud auth were set")
+	}
 
 	switch {
 	case conf.Token != "":
 		cmd.Printf("Authorizing with the -token flag\n")
 	case conf.CredentialsFile != "":
 		cmd.Printf("Authorizing with the credentials file at %q\n", conf.CredentialsFile)
+	case conf.GcloudAuth:
+		cmd.Println("Authorizing with gcloud config")
```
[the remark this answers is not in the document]
enocom: :+1:

cmd/root.go
enocom: Yeah, putting the dialer opts directly on the config makes more sense now. I'll remove this section.

internal/gcloud/gcloud.go
enocom: I've cleaned up this file and deleted a lot of unnecessary stuff. We'll have a proper exit code with a good error now.

internal/gcloud/gcloud.go
enocom: Agreed -- I've just removed these and we'll use the errors already defined in the cmd package.

testsV2/postgres_test.go
enocom: Done. In short, it's to provide a temporary directory to configure gcloud without messing with any existing config. I've added that to the comment.

internal/gcloud/gcloud_test.go
enocom: Done.

internal/gcloud/gcloud.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// configHelper implements oauth2.TokenSource via the `gcloud config config-helper` command.
+type configHelper struct{}
+
+// Token helps gcloudTokenSource implement oauth2.TokenSource.
+func (configHelper) Token() (*oauth2.Token, error) {
+	gcloudCmd, err := Path()
+	if err != nil {
+		return nil, err
```
kurtisvg, line 60: Do these errors need more context? Will the make sense when boiled up and returned to the user?

kurtisvg (approved):

internal/gcloud/gcloud.go
enocom: Here's the error that occurs when a user doesn't have gcloud:

```
Error: exec: "gcloud": executable file not found in $PATH
```

I think that's pretty good.

PR #1182, opened 2022-04-22
feat: add support for unix sockets

enocom:


cmd/root.go
```diff
@@ -130,20 +132,29 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	if len(args) == 0 {
 		return newBadCommandError("missing instance_connection_name (e.g., project:region:instance)")
 	}
-	// First, validate global config.
+
+	userSet := func(f string) bool {
```
kurtisvg, line 136: nit: this might be more clear as "userHasSet", on my first take I thought it was a set of user info

cmd/root.go
```diff
@@ -185,7 +196,18 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 				return newBadCommandError(fmt.Sprintf("could not parse query: %q", res[1]))
 			}
 
-			if a, ok := q["address"]; ok {
+			a, aok := q["address"]
```
kurtisvg, line 199: Why did you move these up? They seem cleaner scoped to the if statement.

internal/proxy/proxy.go
```diff
@@ -133,32 +143,76 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf
...
-		default:
-			np = pc.nextDBPort(version)
+
+		var (
+			// network is one of "tcp" or "unix"
+			network string
+			// address is either a TCP host port, or a Unix socket
+			address string
+		)
+		if (conf.UnixSocket != "" || inst.UnixSocket != "") &&
```
kurtisvg, line 159: nit: I think maybe we should invert this, and do TCP first since that's the "default" experience. 

it would also be nice to have a comment here explaining what we are checking with these 4 conditions.

internal/proxy/proxy.go
```diff
@@ -133,32 +143,76 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf
...
+			(inst.Addr == "" && inst.Port == 0) {
```
kurtisvg, line 160: Do we check anywhere if the user has specified conflicting query parameters?

internal/proxy/proxy.go
```diff
@@ -133,32 +143,76 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf
...
+			network = "unix"
+
+			dir := conf.UnixSocket
+			if dir == "" {
+				dir = inst.UnixSocket
+			}
+			// Attempt to make the directory if it does not already exist.
+			if _, err := os.Stat(dir); err != nil {
+				if err = os.Mkdir(dir, 0750); err != nil {
```
kurtisvg, line 169: Why these file permissions? It looks like do this inconsistently in the proxy, and use both `0755` and `0777` in different places.

cmd/root.go
```diff
@@ -120,6 +120,8 @@ any client SSL certificates.`,
 		"Address on which to bind Cloud SQL instance listeners.")
 	cmd.PersistentFlags().IntVarP(&c.conf.Port, "port", "p", 0,
 		"Initial port to use for listeners. Subsequent listeners increment from this value.")
+	cmd.PersistentFlags().StringVarP(&c.conf.UnixSocket, "unix-socket", "u", "",
+		`Enables Unix sockets for all listeners using the provided directory.`)
```
kurtisvg, line 124: Should this default to `/cloudsql` if set with no argument?

kurtisvg (changes requested):
It looks like this is passing on Windows - is nothing special required for us to support this on Windows environments?

cmd/root.go
enocom: Sounds good to me.

cmd/root.go
enocom: Just below, I check that both address and unix, or address and port have not been set.

internal/proxy/proxy.go
enocom: Sounds good. I've added a comment as well. We might be able to tighten up this boolean expression as well.

internal/proxy/proxy.go
enocom: Yes, see above in the diff where we check that an instance query parameter doesn't include both address and unix, or port and unix.

internal/proxy/proxy.go
enocom: We should probably should make these consistent elsewhere. I think we'd want to allow the user read-write-execute permissions (7), group might deserve the same or even just read permissions (5), and other seems like they shouldn't have any permissions whatsoever (0). So, for any Unix socket or Fuse file, I'd expect either 750, or even 700, although I'm less certain what the right group permission should be.

cmd/root.go
enocom: Since we require users to opt-in to Unix sockets, omitting the default is how we determine if the user did in fact set the flag.

enocom: And even if we did set a default, we'd never use it given the opt-in nature of the flag.

enocom: So I tried this out. If I supply a default and pass the flag without any value, e.g., `--unix-socket` without a directory, Cobra errors. So the default only applies if a user doesn't pass the flag. So I think we're OK to leave this as an empty string.

enocom:
I've updated the Postgres e2e test to use pgx to help ease the transition to v2. As part of that, I discovered pgx does not support Unix sockets on Windows, but fixing that would be only a small amount of work. I've filed https://github.com/jackc/pgx/issues/1199 and otherwise disabled the test on Windows for now.

kurtisvg (approved):

cmd/root.go
```diff
@@ -130,20 +132,29 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	if len(args) == 0 {
 		return newBadCommandError("missing instance_connection_name (e.g., project:region:instance)")
 	}
-	// First, validate global config.
+
+	userHasSet := func(f string) bool {
```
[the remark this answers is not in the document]
enocom: We'll update to flag groups as soon as they're released. https://github.com/spf13/cobra/issues/1626.

internal/proxy/proxy.go
```diff
@@ -134,31 +143,84 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf
...
+		//   an instance-level Unix socket is NOT set
+		//   (e.g.,  I didn't set a Unix socket globally or for this instance)
+		// OR
+		//   a global Unix socket directory IS set, AND
+		//   an instance-level TCP address or port IS set
+		//   (e.g., I want Unix sockets globally, but I'm overriding it for this instance)
+		// use a TCP listener.
+		// Otherwise, use a Unix socket.
+		if (conf.UnixSocket == "" && inst.UnixSocket == "") ||
+			conf.UnixSocket != "" && (inst.Addr != "" || inst.Port != 0) {
```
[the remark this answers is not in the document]
enocom: One of the boolean expressions was unnecessary here, so I've simplified it to:

``` golang
if (conf.UnixSocket == "" && inst.UnixSocket == "") || (inst.Addr != "" || inst.Port != 0) {
  // use tcp
}
```

I think that helps.

internal/proxy/proxy.go
enocom: Did some research on this. https://www.postgresql.org/docs/current/runtime-config-connection.html#:~:text=The%20default%20permissions%20are%200777,revoking%20read%20or%20execute%20permissions.)

In short, the default is 0777 with 0770 being an acceptable alternative. Since there's a chance some users rely on file permissions to control access to the database, I'm going to stick with user and group access, but no access for other, i.e., 0770.

PR #1193, opened 2022-05-06
feat: make Docker images ARM-friendly

shubha-rajan:
## Change Description

Enable multiarchitecture builds in Cloud Build to make our Docker images compatible with ARM64 Linux and MacOS.

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -22,7 +22,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 RUN go get ./...
-RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     go build -ldflags "-X main.metadataString=container.alpine" -o cloud_sql_proxy ./cmd/cloud_sql_proxy
 
 # Final stage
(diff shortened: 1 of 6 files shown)


Dockerfile.alpine
```diff
@@ -13,13 +13,17 @@
...
 
 WORKDIR /go/src/cloudsql-proxy
 COPY . .
 
+ARG TARGETOS
+ARG TARGETARCH
+
 RUN go get ./...
-RUN go build -ldflags "-X main.metadataString=container.alpine" -o cloud_sql_proxy ./cmd/cloud_sql_proxy
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
```
enocom, line 25: My uninformed opinion is that we should always build static binaries, including for Alpine and Buster. However, this is a change from previous versions. Do we need to enable this for the multi arch builds?

enocom (approved):
I built all three containers in my test project, confirmed that the manifest for each includes both architectures, and that each container can start the proxy without issue on Linux. I don't have an M1 machine, but trust the Docker will do the right thing.

image

Dockerfile.alpine
shubha-rajan: Tested this removing the `CGO_ENABLED=0` flag on my M1 mac and it works, so I agree, let's remove from the diff

PR #1212, opened 2022-06-01
feat: add flags to enable Cloud Monitoring and Trace

enocom:


cmd/root.go
```diff
@@ -63,6 +65,9 @@ func Execute() {
 type Command struct {
 	*cobra.Command
 	conf *proxy.Config
+
+	telemetryProject string
+	metricPrefix     string
```
kurtisvg, line 70: should this be telemetryPrefix?

cmd/root.go
```diff
@@ -114,6 +119,10 @@ any client SSL certificates.`,
 		"Path to a service account key to use for authentication.")
 	cmd.PersistentFlags().BoolVarP(&c.conf.GcloudAuth, "gcloud-auth", "g", false,
 		"Use gcloud's user configuration to retrieve a token for authentication.")
+	cmd.PersistentFlags().StringVarP(&c.telemetryProject, "telemetry-project", "j", "",
+		"Enable telemetry with the provided project ID to use for metrics and traces.")
+	cmd.PersistentFlags().StringVarP(&c.metricPrefix, "telemetry-prefix", "x", "",
+		"Prefix to use for metrics.")
```
kurtisvg, line 125: Can we get a better description here.

cmd/root.go
```diff
@@ -114,6 +119,10 @@ any client SSL certificates.`,
 		"Path to a service account key to use for authentication.")
 	cmd.PersistentFlags().BoolVarP(&c.conf.GcloudAuth, "gcloud-auth", "g", false,
 		"Use gcloud's user configuration to retrieve a token for authentication.")
+	cmd.PersistentFlags().StringVarP(&c.telemetryProject, "telemetry-project", "j", "",
+		"Enable telemetry with the provided project ID to use for metrics and traces.")
```
kurtisvg, line 123: Should this specifically call out Cloud Monitoring and Cloud Trace?

cmd/root.go
```diff
@@ -249,11 +258,35 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	return nil
 }
 
+func configureTelemetry(projectID, metricPrefix string) (func(), error) {
```
kurtisvg, line 261: Can we make more clear this is for Stackdriver (or Cloud Trace / Metrics)?

kurtisvg (changes requested):

cmd/root.go
enocom: Yes -- better to be consistent.

cmd/root.go
enocom: Done. I've also made it so each may be enabled independently of the other.

cmd/root.go
enocom: Done.

cmd/root.go
```diff
@@ -249,11 +280,52 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	return nil
 }
 
+// configureTelemetry configures Cloud Trace and Cloud Monitoring based on user
+// selection of each feature. If neither have been enabled, this function is a
+// no-op.
+func configureTelemetry(cmd *Command) (func(), error) {
+	if !cmd.telemetryTracing && !cmd.telemetryMetrics {
+		return func() {}, nil
+	}
```
kurtisvg, line 289: nit: Seems like maybe you could do this check outside and just avoid calling the function if true

cmd/root.go
```diff
@@ -249,11 +280,52 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+		s := trace.ProbabilitySampler(1 / float64(cmd.telemetryTracingSampleRate))
+		trace.ApplyConfig(trace.Config{DefaultSampler: s})
+		trace.RegisterExporter(sd)
+	}
+	if cmd.telemetryMetrics {
+		err = sd.StartMetricsExporter()
+		if err != nil {
+			return func() {}, err
+		}
+	}
```
kurtisvg, line 307: similarly, this feels like maybe better off as just being 2 seperate functions.

kurtisvg (changes requested):

cmd/root.go
```diff
@@ -192,6 +209,19 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	}
 	conf.DialerOpts = opts
 
+	// If a user has set telemetry-project, but hasn't enabled one of metrics or
+	// traces, fail.
+	if userHasSet("telemetry-project") &&
+		!userHasSet("telemetry-metrics") &&
+		!userHasSet("telemetry-traces") {
+		return newBadCommandError("cannot specify --telemetry-project without also setting --telemetry-metrics or --telemetry-traces")
```
kurtisvg, line 217: (nit:) Do we really need to split this line 3 ways? It's shorter than the next line, which we don't split. 

If we have to split it, I'd rather we use a variable over 2 lines rather than 3 imo

kurtisvg (approved):

kurtisvg (approved):

PR #1214, opened 2022-06-03
feat: add support for Automatic IAM AuthN

enocom:
This commit adds a global flag and an instance-level query parameter.
The primary use cases are:

1. Enable IAM AuthN for all instances with the global flag
2. Enable IAM AuthN on a per instance basis with the query param


cmd/root.go
```diff
@@ -122,6 +122,8 @@ any client SSL certificates.`,
 		"Initial port to use for listeners. Subsequent listeners increment from this value.")
 	cmd.PersistentFlags().StringVarP(&c.conf.UnixSocket, "unix-socket", "u", "",
 		`Enables Unix sockets for all listeners using the provided directory.`)
+	cmd.PersistentFlags().BoolVarP(&c.conf.IAMAuthN, "iam-authn-login", "i", false,
```
kurtisvg, line 125: how about "--auto-iam-authn"?

It could maybe even be "auto-iam" or "auto-authn" but they feel a bit too short.

cmd/root.go
```diff
@@ -239,8 +244,15 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 					return newBadCommandError(fmt.Sprintf("unix query param should be only one value: %q", a))
 				}
 				ic.UnixSocket = u[0]
+			}
 
+			if iam, ok := q["iam-authn-login"]; ok {
+				if len(iam) != 1 {
+					return newBadCommandError(fmt.Sprintf("iam authn login param should be only one value: %q", iam))
+				}
+				ic.IAMAuthN = true
```
kurtisvg, line 253: what if `iam-auth-login=false`?

internal/proxy/proxy.go
```diff
@@ -42,6 +42,9 @@ type InstanceConnConfig struct {
 	// connected to the Cloud SQL instance. If set, takes precedence over Addr
 	// and Port.
 	UnixSocket string
+	// IAMAuthN enables Automatic IAM Authentication for the instance.
```
kurtisvg, line 45:
```suggestion
	// IAMAuthN enables automatic IAM DB Authentication for the instance.
```

internal/proxy/proxy.go
```diff
@@ -67,6 +70,10 @@ type Config struct {
 	// connected to any Instances. If set, takes precedence over Addr and Port.
 	UnixSocket string
 
+	// IAMAuthN enables Automatic IAM Authentication for all instances.
```
kurtisvg, line 73:
```suggestion
	// IAMAuthN enables automatic IAM DB Authentication for all instances.
```

internal/proxy/proxy.go
```diff
@@ -209,7 +216,11 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf
 			}
 		}
 
-		m := &socketMount{inst: inst.Name}
+		var opts []cloudsqlconn.DialOption
+		if inst.IAMAuthN {
+			opts = append(opts, cloudsqlconn.WithDialIAMAuthN(true))
```
kurtisvg, line 221: Do we error our somewhere if it's an unsupported engine? e.i. SQL Server? 

maybe that needs to be in the dialer

kurtisvg (changes requested):

cmd/root.go
enocom: auto-iam-authn is good -- it matches the feature and is easy to remember.

internal/proxy/proxy.go
enocom: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/216

cmd/root.go
```diff
@@ -239,8 +244,27 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
 				ic.UnixSocket = u[0]
+			}
 
+			if iam, ok := q["auto-iam-authn"]; ok {
+				if len(iam) != 1 {
+					return newBadCommandError(fmt.Sprintf("auto iam authn param should be only one value: %q", iam))
+				}
+				enableIAMAuthN := true
+				disableIAMAuthn := false
+				switch iam[0] {
```
kurtisvg, line 255: semi-nit: Do we care about case sensitivity?

cmd/root.go
```diff
@@ -239,8 +244,27 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
 				}
 				ic.UnixSocket = u[0]
+			}
 
+			if iam, ok := q["auto-iam-authn"]; ok {
+				if len(iam) != 1 {
+					return newBadCommandError(fmt.Sprintf("auto iam authn param should be only one value: %q", iam))
+				}
+				enableIAMAuthN := true
+				disableIAMAuthn := false
```
kurtisvg, line 254: nit: should you move these into the case blocks, so the alloc doesn't happen unless it's used?

kurtisvg (approved):

cmd/root.go
enocom: I think we do -- but right now all our query params are case sensitive. So I'm inclined to address this problem in one go.

cmd/root.go
enocom: yes good idea.

PR #1215, opened 2022-06-04
feat: add flags to enable prometheus metrics

enocom:
This commit adds two flags (prometheus-namespace and prometheus-port)
which enable a Prometheus exporter exposed at localhost:9090/metrics.
The port is configurable with prometheus-port. When enabled, the metrics
endpoint will serve all metrics collected in the Go connector in the
Prometheus format.


enocom:
Here's an example of what the metrics look like when scraped by Prometheus:

image

cmd/root_test.go
```diff
@@ -349,3 +354,48 @@ func TestCommandWithCustomDialer(t *testing.T) {
...
 }
+
+func TestPrometheusMetricsEndpoint(t *testing.T) {
+	c := NewCommand(WithDialer(&spyDialer{}))
+	// Keep the test output quiet
+	c.SilenceUsage = true
+	c.SilenceErrors = true
+	c.SetArgs([]string{
+		"--prometheus-namespace", "prometheus",
+		"--prometheus-port", "9999",
```
kurtisvg, line 365: Should we have a test where we don't set port?

cmd/root.go
```diff
@@ -114,6 +119,10 @@ any client SSL certificates.`,
 		"Path to a service account key to use for authentication.")
 	cmd.PersistentFlags().BoolVarP(&c.conf.GcloudAuth, "gcloud-auth", "g", false,
 		"Use gcloud's user configuration to retrieve a token for authentication.")
+	cmd.PersistentFlags().StringVarP(&c.prometheusNamespace, "prometheus-namespace", "n", "",
+		"Enable Prometheus for metric collection using the provided namespace")
+	cmd.PersistentFlags().StringVarP(&c.prometheusPort, "prometheus-port", "s", "9090",
```
kurtisvg, line 124: I'm a little concerned about `-n` and `-s` as the shortcode here. For one, I'm not sure why `-s` is for the port, and two I'm worried about these flags taking the short flags of more useful or common flags in the future. Should we not have shortcode on these for now?

cmd/root.go
```diff
@@ -256,6 +269,32 @@ func runSignalWrapper(cmd *Command) error {
...
+		if err != nil {
+			return err
+		}
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", e)
+		addr := fmt.Sprintf("localhost:%s", cmd.prometheusPort)
+		server := &http.Server{Addr: addr, Handler: mux}
+		go func() {
+			select {
+			case <-ctx.Done():
```
kurtisvg, line 285: Down below (L305), we're using `cmd.Context().Done()`. Which should we be using?

cmd/root.go
```diff
@@ -256,6 +269,32 @@ func runSignalWrapper(cmd *Command) error {
...
+				if err := server.Shutdown(context.Background()); err != nil {
```
kurtisvg, line 286: Is background the right thing to use here? If so, should it have a timeout so this doesn't hang the proxy?

If ctx.Done() is already closed, doesn't that mean the server is already shutdown? Would shutdownCh already have been sent a value then? Maybe this needs to be logged indiscriminately.

cmd/root.go
```diff
@@ -256,6 +269,32 @@ func runSignalWrapper(cmd *Command) error {
...
+					shutdownCh <- fmt.Errorf("failed to stop prometheus HTTP server: %v", err)
+				}
+			}
+		}()
+		go func() {
+			if err := server.ListenAndServe(); err != nil {
```
kurtisvg, line 292: Should we check for `http.ErrServerClosed` here?

kurtisvg (changes requested):

cmd/root.go
enocom: Good idea.

cmd/root.go
enocom: Probably not. I added a context with a 1 second timeout. And yes, logging makes more sense as you suggest.

cmd/root.go
enocom: Let's use the child context that we create at the top of this function based on the cmd Context.

cmd/root.go
enocom: Good point. I've removed the short flags. We can do the same in the other telemetry PR too.

cmd/root.go
```diff
@@ -114,6 +120,10 @@ any client SSL certificates.`,
 		"Path to a service account key to use for authentication.")
 	cmd.PersistentFlags().BoolVarP(&c.conf.GcloudAuth, "gcloud-auth", "g", false,
 		"Use gcloud's user configuration to retrieve a token for authentication.")
+	cmd.PersistentFlags().StringVar(&c.prometheusNamespace, "prometheus-namespace", "",
+		"Enable Prometheus for metric collection using the provided namespace")
+	cmd.PersistentFlags().StringVar(&c.prometheusPort, "prometheus-port", "9090",
```
kurtisvg, line 125: I think we chatted about this and decided we only wanted want HTTP server, even for things like health checks and this endpoint. 

If so, we probably want to remove this flag to something like "--http-port" or "--http-server-port"

We also probably want to abstract it out and register endpoints as needed, but maybe that can go in a different PR.

kurtisvg (approved):

cmd/root.go
enocom: Sounds good. I'll update here.

enocom: I agree -- let's extract the HTTP server on the next flag that needs its own. I'll handle the endpoint registration in that future PR.

PR #1240, opened 2022-06-21
feat: add max connections flag

enocom:
The max connections flags allows callers to limit the number of outgoing
connections to the Cloud SQL backend. Max connections are tracked on a
global level across all instances.

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/cloudsqlconn"
@@ -73,6 +74,11 @@ type Config struct {
 	// IAMAuthN enables automatic IAM DB Authentication for all instances.
 	// Postgres-only.
 	IAMAuthN bool
+
+	// MaxConnections are the maximum number of connections the Client may
+	// establish to the Cloud SQL server side proxy before refusing additional
+	// connections. A zero-value indicates no limit.
+	MaxConnections uint32
 
 	// Instances are configuration for individual instances. Instance
 	// configuration takes precedence over global configuration.
@@ -130,13 +136,22 @@ func (c *portConfig) nextDBPort(version string) int {
 	}
 }
 
-// Client represents the state of the current instantiation of the proxy.
+// Client proxies connections from a local client to the remote server side
+// proxy for multiple Cloud SQL instances.
 type Client struct {
 	cmd    *cobra.Command
 	dialer cloudsql.Dialer
 
 	// mnts is a list of all mounted sockets for this client
 	mnts []*socketMount
+
+	// connCount tracks the number of all open connections from the Client to
+	// all Cloud SQL instances.
+	connCount uint32
+
+	// maxConns is the maximum number of allowed connections tracked by
+	// connCount. If not set, there is no limit.
+	maxConns uint32
 }
 
 // NewClient completes the initial setup required to get the proxy to a "steady" state.
@@ -232,10 +247,17 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, cmd *cobra.Command, conf 
 		cmd.Printf("[%s] Listening on %s\n", inst.Name, addr.String())
 		mnts = append(mnts, m)
 	}
-	return &Client{mnts: mnts, cmd: cmd, dialer: d}, nil
+	c := &Client{
+		mnts:     mnts,
+		cmd:      cmd,
+		dialer:   d,
+		maxConns: conf.MaxConnections,
+	}
+	return c, nil
 }
 
-// Serve listens on the mounted ports and beging proxying the connections to the instances.
+// Serve starts proxying connections for all configured instances using the
+// associated socket.
 func (c *Client) Serve(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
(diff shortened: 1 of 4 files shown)


internal/proxy/proxy.go
```diff
@@ -289,6 +311,15 @@ func (c *Client) serveSocketMount(ctx context.Context, s *socketMount) error {
 		go func() {
 			c.cmd.Printf("[%s] accepted connection from %s\n", s.inst, cConn.RemoteAddr())
 
+			count := atomic.AddUint32(&c.connCount, 1)
+			defer atomic.AddUint32(&c.connCount, ^uint32(0))
+
+			if c.maxConns > 0 && count > c.maxConns {
+				c.cmd.Printf("max connections (%v) exceeded, refusing new connection\n", c.maxConns)
+				cConn.Close()
```
hessjcg, line 319: The proxy is closing connections. This gives the client no indication that the server is too busy. I would expect the proxy to behave in a way that provides better hints to the client that it is not accepting connections. 

Two options I would consider better behaved:

- refuse to accept the connection by closing the listening socket. Clients will get "Connection Refused" or "EOF" from their net.Dial() call. This is fail-fast. 
- allow the inbound connections to queue up in the inbound server connection queue by only calling `s.listener.Accept()` when connections < maxConnections. On the client side, net.Dial() will hang until the server accepts the connection or client-side tcp connect timeout is exceeded.  This is fail-slow.

I wrote a simple example executable demonstrating how to implement both server-side options in go: 
[main.go.txt](https://github.com/GoogleCloudPlatform/cloudsql-proxy/files/8962108/main.go.txt)

internal/proxy/proxy_test.go
```diff
@@ -241,6 +259,50 @@ func TestClientInitialization(t *testing.T) {
...
+	go c.Serve(context.Background())
+
+	conn1, err1 := net.Dial("tcp", "127.0.0.1:5000")
+	if err1 != nil {
+		t.Fatalf("net.Dial error: %v", err1)
+	}
+	defer conn1.Close()
+
+	conn2, err2 := net.Dial("tcp", "127.0.0.1:5000")
+	if err2 != nil {
```
hessjcg, line 286: When the server is too busy, this is the call that should fail. not `conn.Read()`

hessjcg (changes requested):
I'm missing a lot of context, but i'm not sure that this is the right way to implement refusing connections when the server is too busy. I left more detailed notes in the comment. Let's discuss so you can fill me in on the design decisions that I'm not aware of.

internal/proxy/proxy.go
```diff
@@ -289,6 +311,15 @@ func (c *Client) serveSocketMount(ctx context.Context, s *socketMount) error {
 		go func() {
 			c.cmd.Printf("[%s] accepted connection from %s\n", s.inst, cConn.RemoteAddr())
 
+			count := atomic.AddUint32(&c.connCount, 1)
+			defer atomic.AddUint32(&c.connCount, ^uint32(0))
```
kurtisvg, line 315: a comment here would be good

internal/proxy/proxy.go
```diff
@@ -289,6 +311,15 @@ func (c *Client) serveSocketMount(ctx context.Context, s *socketMount) error {
 		go func() {
 			c.cmd.Printf("[%s] accepted connection from %s\n", s.inst, cConn.RemoteAddr())
 
+			count := atomic.AddUint32(&c.connCount, 1)
```
kurtisvg, line 314: Is 32 the right choice? Looks like we are using 64 in v1 at the momment

internal/proxy/proxy.go
kurtisvg: So this is a reasonable thought, but the way it's currently implemented does match what's being done in v1: https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/957d1609ad96bfed77b3744f1c11a762010bc06e/proxy/proxy/client.go#L177 (that doesn't mean we shouldn't discuss if this is the behavior we want though).

For the current behavior, clients do get a "server closed the connection" and if they check the proxy logs, they'll see why the connection was closed. 

For the first alternative, I think that closing/opening the socket will introduce a lot of overhead (it's essentially a syscall). We'd have to reopen/close (possible several) sockets all at once any time a connection opens or closes. It's also potentially harder to debug, as there won't be a way for the proxy to tell when this caused a client side connection to fail. 

For the second alternative, this seems similar to how [netutil.LimitListener](https://cs.opensource.google/go/x/net/+/263ec571:netutil/listen.go;l=16) works. The one caveat is that this for just 1 socket, where we need to share across all of them. This probably has the same drawbacks -- it'll end up being hard to debug and might cause issues.

So my hot take is the current behavior seems preferable. thoughts?

kurtisvg (changes requested):

internal/proxy/proxy.go
hessjcg: Let's rule out Option 1. I agree with your reasoning about syscalls.

Going ahead with an implementation matching V1 is viable.  Changing to an Option 2 implementation that respects a global active connection limit is also viable and for me, closer to what I expect well-behaved proxy to do in theory. 

But more important than my opinion, what has happened in practice? In our experience with the V1 proxy, has the limit and connection behavior lead to support calls with confused customers or do they not really notice? Maybe V1 behavior is good enough for the real world.

enocom: > Maybe V1 behavior is good enough for the real world.

This is my view. Granted, we don't have a lot of data on how frequently these flags are used. So we could choose:

1. to forgo implementing this flag until someone asks for it, or
2. match v1 behavior.

The guiding light here is that we want the migration to v2 to be as friction-less as possible, even if that means having flags that are no used heavily.

internal/proxy/proxy_test.go
enocom: The discussion above addresses this.

internal/proxy/proxy.go
enocom: I guess I'll just match v1. uint64 seems to allow an impossibly high amount of connections, but I probably should toe the line of matching v1 behavior.

internal/proxy/proxy.go
```diff
@@ -196,13 +202,22 @@ func (c *portConfig) nextDBPort(version string) int {
...
 type Client struct {
 	cmd    *cobra.Command
 	dialer cloudsql.Dialer
 
 	// mnts is a list of all mounted sockets for this client
 	mnts []*socketMount
+
+	// connCount tracks the number of all open connections from the Client to
+	// all Cloud SQL instances.
+	connCount uint64
```
kurtisvg, line 216: Do we need to ensure this is 64bit aligned to use atomic.sync? 

If so, should we add a test to ensure it doesn't change? https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/550/files#diff-de11c0b4b635f0cda6d02b369cf94915ec3595d09d5115753f8e9d9d19dcf0dcR290-R296

kurtisvg (approved):

PR #1242, opened 2022-06-21
feat: private-ip flag for v2

hessjcg:
This creates a private-ip address flag for the work in progress.


cmd/root.go
```diff
@@ -314,6 +312,38 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+//  name the name of the option in the query string
+//  q the map of the parsed query string parameters from url.ParseQuery()
+// returns
+//   nil if no value was found
+//   true if the value is 't' or 'true' case insensitive
+//   false if the value is 'f' or 'false' case insensitive
+func parseBoolOpt(name string, q map[string][]string) (*bool, error) {
+	if iam, ok := q[name]; ok {
+		if len(iam) != 1 {
+			return nil, newBadCommandError(fmt.Sprintf("auto iam authn param should be only one value: %q", iam))
```
kurtisvg, line 325: need to generalize this beyond "auto-iam-authn"

cmd/root.go
```diff
@@ -314,6 +312,38 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+		switch v {
+		case "true", "t":
+			enable := true
+			return &enable, nil
+		case "false", "f":
+			disable := false
+			return &disable, nil
+		default:
+			return nil, newBadCommandError(
+				fmt.Sprintf("auto iam authn query param should be true or false, got: %q",
```
kurtisvg, line 337: Same here

cmd/root.go
```diff
@@ -287,24 +292,17 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+			if err != nil {
+				return err
 			}
+			ic.IAMAuthN = iam
+
+			privateIP, err := parseBoolOpt("private-ip", q)
+			if err != nil {
+				return err
+			}
+			ic.PrivateIP = privateIP
```
kurtisvg, line 305: any reason not to just assign these directly to the config?

kurtisvg (changes requested):

cmd/root.go
hessjcg: done

cmd/root.go
hessjcg: done

cmd/root.go
```diff
@@ -148,6 +148,8 @@ any client SSL certificates.`,
 		`Enables Unix sockets for all listeners using the provided directory.`)
 	cmd.PersistentFlags().BoolVarP(&c.conf.IAMAuthN, "auto-iam-authn", "i", false,
 		"Enables Automatic IAM Authentication for all instances")
+	cmd.PersistentFlags().BoolVarP(&c.conf.PrivateIP, "private-ip", "", false,
```
enocom, line 151: `BoolVarP` means you'll provide a long flag and a short flag. We should decide if private-ip warrants a short flag. I think we could skip the short flag for now and just use `BoolVar`

cmd/root.go
```diff
@@ -148,6 +148,8 @@ any client SSL certificates.`,
...
+		"Enables Automatic IAM Authentication for all instances")
```
enocom, line 152: This will need to change.

cmd/root.go
hessjcg: done

hessjcg:
This still needs an integration test.

cmd/root.go
```diff
@@ -285,6 +278,38 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
 }
 
+// parseBoolOpt parses a boolean option from the query string
+//  name the name of the option in the query string
+//  q the map of the parsed query string parameters from url.ParseQuery()
+// returns
+//   nil if no value was found
+//   true if the value is 't' or 'true' case insensitive
+//   false if the value is 'f' or 'false' case insensitive
+func parseBoolOpt(name string, q map[string][]string) (*bool, error) {
```
enocom, line 288: q should go first in the argument list and be a `url.Values` type.

cmd/root.go
```diff
@@ -285,6 +278,38 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
 	return nil
 }
 
+// parseBoolOpt parses a boolean option from the query string
+//  name the name of the option in the query string
```
enocom, line 282: This and everything below is available from the type signature and therefore unnecessary, with the exception of t/true and f/false.

cmd/root.go
```diff
@@ -285,6 +278,38 @@ func parseConfig(cmd *cobra.Command, conf *proxy.Config, args []string) error {
...
+//  q the map of the parsed query string parameters from url.ParseQuery()
+// returns
+//   nil if no value was found
+//   true if the value is 't' or 'true' case insensitive
+//   false if the value is 'f' or 'false' case insensitive
+func parseBoolOpt(name string, q map[string][]string) (*bool, error) {
+	if iam, ok := q[name]; ok {
```
enocom, line 289: This should be `!ok`. Typically the error flow is indented and the happy path is not.

cmd/root_test.go
```diff
@@ -378,6 +456,7 @@ func TestNewCommandWithErrors(t *testing.T) {
 			desc: "enabling a Prometheus port without a namespace",
 			args: []string{"--htto-port", "1111", "proj:region:inst"},
 		},
+		//TODO error when privateip=bad-string
```
enocom, line 459: dev artifact?

internal/proxy/proxy.go
```diff
@@ -47,6 +47,10 @@ type InstanceConnConfig struct {
 	// IAMAuthN enables automatic IAM DB Authentication for the instance.
 	// Postgres-only. If it is nil, the value was not specified.
 	IAMAuthN *bool
+
+	// PrivateIP tells the proxy to attempt to connect to the db instance's
+	// private ip address instead of the public IP address
```
enocom, line 52: ip should be IP

internal/proxy/proxy.go
```diff
@@ -79,12 +83,16 @@ type Config struct {
...
+	// for all instances.
+	PrivateIP bool
+
 	// Instances are configuration for individual instances. Instance
 	// configuration takes precedence over global configuration.
 	Instances []InstanceConnConfig
 
 	// Dialer specifies the dialer to use when connecting to Cloud SQL
-	// instances.
+	// instances. Used by serverless.
```
enocom, line 95: Let's leave this out for now. We need a separate commit explaining that the "public" interface is actually subject to change and is for internal use only, e.g., "Use at your own risk".

structure.md
```diff
@@ -0,0 +1,111 @@
+# Structure
```
enocom: This probably wasn't mean to be included in the commit.

We will need to write something similar with the table for a v1 to v2 migration guide, though.

internal/proxy/proxy_test.go
```diff
@@ -242,6 +244,31 @@ func TestClientInitialization(t *testing.T) {
...
+func (d *spyDialer) Dial(ctx context.Context, inst string, opts ...cloudsqlconn.DialOption) (net.Conn, error) {
+	log.Printf("spyDialer.Dial() called")
+	d.LastOpts = opts
+	d.LastInst = inst
+	conn, _ := net.Pipe()
+	return conn, nil
+}
+
+func (d *spyDialer) EngineVersion(ctx context.Context, inst string) (string, error) {
+	log.Printf("spyDialer.EngineVersion() called")
```
enocom: Let's keep these log statements out.

internal/proxy/proxy_test.go
```diff
@@ -242,6 +244,31 @@ func TestClientInitialization(t *testing.T) {
...
+
+func (d *spyDialer) Close() error {
+	log.Printf("spyDialer.Close() called")
+	// do nothing
+	return nil
+}
+
+func (d *spyDialer) Dial(ctx context.Context, inst string, opts ...cloudsqlconn.DialOption) (net.Conn, error) {
+	log.Printf("spyDialer.Dial() called")
+	d.LastOpts = opts
```
enocom: These need to be protected behind a mutext as the dialer and test goroutines won't be the same.

internal/proxy/proxy_test.go
```diff
@@ -242,6 +244,31 @@ func TestClientInitialization(t *testing.T) {
 	}
 }
 
+type spyDialer struct {
```
enocom: Let's move this up to live next to fakeDialer.

internal/proxy/proxy.go
```diff
@@ -282,6 +287,29 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
 	return &Client{mnts: mnts, cmd: cmd, dialer: d}, nil
 }
 
+// dialOptions interprets appropriate dial options for a particular instance
+// configuration
+func dialOptions(conf *Config, inst *InstanceConnConfig) []cloudsqlconn.DialOption {
```
enocom, line 292: This is a different pattern from proxy.Confg.DialerOptions. Let's pick one pattern and use it for both Option and DialOption.

internal/proxy/proxy.go
```diff
@@ -282,6 +287,29 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
...
+	} else {
+		opts = append(opts, cloudsqlconn.WithDialIAMAuthN(conf.IAMAuthN))
+	}
+
+	// related to --private-ip=(bool) option
+	if inst.PrivateIP != nil && *inst.PrivateIP ||
+		inst.PrivateIP == nil && conf.PrivateIP {
+		opts = append(opts, cloudsqlconn.WithPrivateIP())
+	} else {
+		opts = append(opts, cloudsqlconn.WithPublicIP())
```
enocom, line 307: We default to public, so this else block can probably be deleted.

internal/proxy/proxy.go
```diff
@@ -282,6 +287,29 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
...
 
+// dialOptions interprets appropriate dial options for a particular instance
+// configuration
+func dialOptions(conf *Config, inst *InstanceConnConfig) []cloudsqlconn.DialOption {
+	var opts []cloudsqlconn.DialOption
+
+	// related to --auto-iam-authn=(bool)
+	if inst.IAMAuthN != nil {
+		opts = append(opts, cloudsqlconn.WithDialIAMAuthN(*inst.IAMAuthN))
+	} else {
```
enocom, line 298: This else block is also unneeded.

cmd/root_test.go
```diff
@@ -201,6 +202,23 @@ func TestNewCommandArguments(t *testing.T) {
...
+		{
+			desc: "using the private-ip flag",
+			args: []string{"--private-ip", "proj:region:inst"},
+			want: withDefaults(&proxy.Config{
+				PrivateIP: true,
+			}),
+		},
+		{
+			desc: "using the private-ip flag query param",
+			// the query param's presence equates to true
```
enocom, line 214: This is an old comment and can be deleted.

cmd/root_test.go
```diff
@@ -491,3 +570,124 @@ func TestPrometheusMetricsEndpoint(t *testing.T) {
 		t.Fatalf("expected a 200 status, got = %v", resp.StatusCode)
 	}
 }
+
+func TestParseBoolOpt(t *testing.T) {
```
enocom, line 574: I'd rather test this through the public API as you have above.

enocom (changes requested):

cmd/root.go
hessjcg: Done.

cmd/root.go
hessjcg: done

cmd/root.go
hessjcg: done

internal/proxy/proxy.go
hessjcg: Even though WithPublicIP() is the default for the cloudsqlconn library today, I would rather make it explicit in our code. Too often, the upstream library team changes a default and unexpectedly breaks something. Are you ok if I leave this as-is?

internal/proxy/proxy.go
hessjcg: Ok, removed.

cmd/root_test.go
hessjcg: done.

enocom (approved):
two minor things

cmd/root_test.go
```diff
@@ -287,6 +304,66 @@ func TestAutoIAMAuthNQueryParams(t *testing.T) {
...
+		want *bool
+	}{
+		{
+			desc: "when the query string is absent",
+			args: []string{"proj:region:inst"},
+			want: nil,
+		},
+		{
+			desc: "when the query string is true",
+			args: []string{"proj:region:inst?private-ip=true"},
```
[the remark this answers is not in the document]
hessjcg: The only 2 planned boolean args on an instance are `auto-iam-authn` and `private-ip.`
I can adapt the parseBooleanValue() function to treat an empty url parameter as true.

cmd/root_test.go
hessjcg: removed.

internal/proxy/proxy.go
```diff
@@ -282,6 +256,58 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
 	return &Client{mnts: mnts, cmd: cmd, dialer: d}, nil
 }
 
+// dialerOptions builds appropriate list of options from the Config
+// values for use by cloudsqlconn.NewClient()
+func dialerOptions(c *Config) ([]cloudsqlconn.Option, error) {
```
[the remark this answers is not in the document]
hessjcg: I was trying to make the call signatures for these functions consistent: 
 ```
func (c *Config) DialerOptions() ([]cloudsqlconn.Option, error)  {}
func dialOptions(conf *Config, inst *InstanceConnConfig) []cloudsqlconn.DialOption {} 
``` 

So when Eno and I discussed the inconsistency, we landed on this:
 ```
func dialerOptions(c *Config) ([]cloudsqlconn.Option, error)  {}
func dialOptions(conf *Config, inst *InstanceConnConfig) []cloudsqlconn.DialOption {} 
``` 

The alternative to make `DialOptions` a  member functions of *Config, which seemed worse to me because
`DialOptions` is more transformation on Config and InstanceConnConfig, and less a member of Config.
```
func (c *Config) DialerOptions() ([]cloudsqlconn.Option, error)  {}
func (c *Config) DialOptions(inst *InstanceConnConfig) ([]cloudsqlconn.Option, error)  {}
```

internal/proxy/proxy.go
```diff
@@ -282,6 +256,58 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
...
+
+	return opts, nil
+}
+
+// dialOptions interprets appropriate dial options for a particular instance
+// configuration
+func dialOptions(conf *Config, inst *InstanceConnConfig) []cloudsqlconn.DialOption {
+	var opts []cloudsqlconn.DialOption
+
+	// related to --auto-iam-authn=(bool)
```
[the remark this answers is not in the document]
hessjcg: removed.

internal/proxy/proxy.go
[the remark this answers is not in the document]
kurtisvg: The difference between function vs method is largely syntactic / organizational. For me, it more sense to me to use `func (c *Config) DialerOptions()` and `func (i *instanceConnConfig) dialOptions()` (and making InstanceConnConfig contain a reference to it's parent config) because they signify a way of transforming the object to another type of data, similar to how `String()` works. It's also easier to discover (because it's attached to the object it transforms, and lands closer to it in the doc) and leaves the door open to more complex behavior (say in the future it needs to be thread safe, we can add an internal mutex more easily).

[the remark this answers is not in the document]
enocom: I think both approaches are reasonable, but if there's preference for methods then let's do that.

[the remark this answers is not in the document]
hessjcg: Ok, i'll try again with a version closer to @kurtisvg's suggestion.

internal/proxy/proxy.go
```diff
@@ -88,6 +96,25 @@ type Config struct {
 	Dialer cloudsql.Dialer
 }
 
+// DialOptions interprets appropriate dial options for a particular instance
+// configuration
+func (c *Config) DialOptions(i *InstanceConnConfig) []cloudsqlconn.DialOption {
```
hessjcg, line 101: Following the dumb struct smart procedure idiom, I put DialOptions on `Config` instead of `InstanceConnConfig`.  I like it better here because two less tight couplings than what @kurtisvg proposed: 

- `InstanceConnConfig` doesn't need a pointer field to `Config`
-  the code doesn't need to set the a pointer to the correct parent `Config`

internal/proxy/proxy.go
```diff
@@ -88,6 +96,25 @@ type Config struct {
...
+	var opts []cloudsqlconn.DialOption
+
+	if i.IAMAuthN != nil {
+		opts = append(opts, cloudsqlconn.WithDialIAMAuthN(*i.IAMAuthN))
+	}
+
+	if i.PrivateIP != nil && *i.PrivateIP ||
+		i.PrivateIP == nil && c.PrivateIP {
```
kurtisvg, line 109: nit: seems like a line break is unnecessary here

kurtisvg (approved):

internal/proxy/proxy.go
hessjcg: removed.

internal/proxy/proxy.go
```diff
@@ -263,10 +290,7 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
 			}
 		}
 
-		var opts []cloudsqlconn.DialOption
-		if inst.IAMAuthN != nil {
-			opts = append(opts, cloudsqlconn.WithDialIAMAuthN(*inst.IAMAuthN))
-		}
+		opts := conf.DialOptions(&inst)
```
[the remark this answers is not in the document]
hessjcg: changed to value.

PR #1245, opened 2022-06-22
fix: resolve data race on closing client

enocom:
Previously, when a caller would try to close the client that was
serving, there was a data race between the assignment of a listener, and
the removal of the listener within socketMount. This commit ensures that
the listener is created prior to the socket mount and so makes the data
race impossible.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -432,7 +432,11 @@ func runSignalWrapper(cmd *Command) error {
 	case p = <-startCh:
 	}
 	cmd.Println("The proxy has started successfully and is ready for new connections!")
-	defer p.Close()
+	defer func() {
+		if cErr := p.Close(); cErr != nil {
+			cmd.PrintErrln(p.Close())
+		}
+	}()
 
 	go func() {
 		shutdownCh <- p.Serve(ctx)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -394,7 +394,7 @@ func runSignalWrapper(cmd *Command) error {
 	cmd.Println("The proxy has started successfully and is ready for new connections!")
 	defer func() {
 		if cErr := p.Close(); cErr != nil {
-			cmd.PrintErrln(cErr)
+			cmd.PrintErrf("error during shutdown: %v\n", cErr)
 		}
 	}()
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -434,7 +434,7 @@ func runSignalWrapper(cmd *Command) error {
 	cmd.Println("The proxy has started successfully and is ready for new connections!")
 	defer func() {
 		if cErr := p.Close(); cErr != nil {
-			cmd.PrintErrln(p.Close())
+			cmd.PrintErrln(cErr)
 		}
 	}()
 
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -193,11 +193,41 @@ func (c *Client) Serve(ctx context.Context) error {
 	return <-exitCh
 }
 
+// MultiErr is a group of errors wrapped into one.
+type MultiErr []error
+
+// Error returns a single string representing one or more errors.
+func (m MultiErr) Error() string {
+	l := len(m)
+	if l == 1 {
+		return m[0].Error()
+	}
+	var errStr string
+	for i, err := range m {
+		if i == l-1 {
+			errStr += err.Error()
+			continue
+		}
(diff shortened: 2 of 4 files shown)


internal/proxy/proxy.go
```diff
@@ -311,22 +242,92 @@ type socketMount struct {
...
+	return s.listener.Accept()
 }
 
 // close stops the mount from listening for any more connections
-func (s *socketMount) close() error {
-	err := s.listener.Close()
-	s.listener = nil
-	return err
+func (s *socketMount) Close() error {
+	return s.listener.Close()
```
hessjcg, line 330: We used to make the listener null. Why not make it null again?

kurtisvg (changes requested):

internal/proxy/proxy.go
enocom: That was the root of the race condition. Mutating listener across goroutines.

internal/proxy/proxy.go
```diff
@@ -311,22 +242,92 @@ type socketMount struct {
...
 
-// listen causes a socketMount to create a Listener at the specified network address.
-func (s *socketMount) listen(ctx context.Context, network string, address string) (net.Addr, error) {
+func newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst InstanceConnConfig, version string) (*socketMount, error) {
+	var (
+		// network is one of "tcp" or "unix"
+		network string
+		// address is either a TCP host port, or a Unix socket
+		address string
+	)
```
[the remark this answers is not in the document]
enocom: If it weren't for the comments, I think `var network, address string` might be reasonable. But in my view the comments make the block variable declaration preferable.

cmd/root.go
```diff
@@ -392,7 +392,11 @@ func runSignalWrapper(cmd *Command) error {
 	case p = <-startCh:
 	}
 	cmd.Println("The proxy has started successfully and is ready for new connections!")
-	defer p.Close()
+	defer func() {
+		if cErr := p.Close(); cErr != nil {
+			cmd.PrintErrln(cErr)
```
kurtisvg, line 397: Can we add some context that this is an error that happened during shutdown?

internal/proxy/proxy.go
```diff
@@ -307,22 +240,50 @@ func (c *Client) Serve(ctx context.Context) error {
...
+		return m[0].Error()
+	}
+	var errStr string
+	for i, err := range m {
+		if i == l-1 {
+			errStr += err.Error()
+			continue
+		}
+		errStr += err.Error() + ", "
+	}
```
kurtisvg, line 259: This seems potentially problematic since we're adding a lot of strings together and will have exponential runtime growth. Can we use `strings.Join()` instead?

internal/proxy/proxy.go
```diff
@@ -188,95 +188,28 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
...
-		if inst.IAMAuthN != nil {
-			opts = append(opts, cloudsqlconn.WithDialIAMAuthN(*inst.IAMAuthN))
-		}
-		m := &socketMount{inst: inst.Name, dialOpts: opts}
-		addr, err := m.listen(ctx, network, address)
+		m, err := newSocketMount(ctx, conf, pc, inst, version)
 		if err != nil {
 			for _, m := range mnts {
-				m.close()
+				_ = m.Close() // TODO: log this error?
```
kurtisvg, line 207: Seems like we should be consistent and log the error here if we are logging it when the Client closes

internal/proxy/proxy.go
```diff
@@ -222,92 +222,31 @@ func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client,
...
-		m := &socketMount{inst: inst.Name, dialOpts: opts}
-		addr, err := m.listen(ctx, network, address)
+		m, err := newSocketMount(ctx, conf, pc, inst, version)
 		if err != nil {
 			for _, m := range mnts {
-				m.close()
+				mErr := m.Close()
+				if mErr != nil {
+					cmd.PrintErrf("failed to close mount: %v", mErr)
+				}
```
kurtisvg, line 244: nit: 
```suggestion
				
				if mErr := m.Close(); mErr != nil {
					cmd.PrintErrf("failed to close mount: %v", mErr)
				}
```

kurtisvg (approved):

PR #1246, opened 2022-06-22
feat: add support for structured logs

enocom:


hessjcg (approved):
LGTM

cmd/root.go
```diff
@@ -91,42 +98,57 @@ func WithDialer(d cloudsql.Dialer) Option {
...
-				return err
-			}
-			// The arguments are parsed. Usage is no longer needed.
-			cmd.SilenceUsage = true
-			return nil
-		},
-		RunE: func(*cobra.Command, []string) error {
-			return runSignalWrapper(c)
-		},
+any client SSL certificates.`
```
kurtisvg, line 121: Seems like we could still declare these inline during initialization on L101

internal/proxy/proxy.go
```diff
@@ -132,15 +139,16 @@ func (c *portConfig) nextDBPort(version string) int {
 
 // Client represents the state of the current instantiation of the proxy.
 type Client struct {
-	cmd    *cobra.Command
 	dialer cloudsql.Dialer
 
 	// mnts is a list of all mounted sockets for this client
 	mnts []*socketMount
+
+	logger log.Logger
```
kurtisvg, line 147: Shouldn't this say a reference to cmd? What if cmd's logger changes?

kurtisvg (changes requested):

internal/log/log.go
```diff
@@ -0,0 +1,97 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// Logger is the interface used throughout the project for logging.
+type Logger interface {
+	// Infof is for reporting informational messages.
+	Infof(format string, args ...interface{})
+	// Errorf is for reporitng errors.
+	Errorf(format string, args ...interface{})
```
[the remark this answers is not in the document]
enocom: We decided introduce a debug flag that would enable additional debug logging. Fatal and Warning are out for now.

cmd/root.go
enocom: Sorry -- trying to make the relationship between cobra.Command and our command more clear here. This probably didn't help. I've cleaned it up.

cmd/root.go
```diff
@@ -78,6 +80,11 @@ type Command struct {
 	httpPort                   string
 }
 
+// SetLogger overrides the default logger.
+func (c *Command) SetLogger(l log.Logger) {
+	c.conf.Logger = l
```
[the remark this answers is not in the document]
enocom: Yes, that's better. It also means we have to explicitly pass the logger as a parameter to the proxy.NewClient method, but that also seems better.

internal/proxy/proxy.go
enocom: I think our contract here is that prior to calling `Execute`, the logger can change, but after, the command has started and the configuration is set. Side note: Hot reloading will probably create a new command (saving some state from the previous command) and if the logger changes, it will still be done before the next call to `Execute`.

internal/log/log.go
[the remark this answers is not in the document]
enocom: Thought I had an issue for this already filed, but didn't. So I created one here: https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/1255.

enocom:
Looks like we lost some logging around which authentication method is in use. I'll restore it here.

cmd/root.go
```diff
@@ -75,6 +78,11 @@ type Command struct {
 	httpPort                   string
 }
 
+// SetLogger overrides the default logger.
+func (c *Command) SetLogger(l log.Logger) {
+	c.logger = l
```
kurtisvg, line 83: Should this be an Option?

internal/proxy/proxy.go
kurtisvg: I think we'll want hotloading to preserve the existing connections, so my 2c is that the cmd will need to be updated and adapted instead. But we can handle that in a different PR if need be.

internal/proxy/proxy.go
```diff
@@ -198,21 +206,22 @@ func (c *portConfig) nextDBPort(version string) int {
...
 }
 
 // NewClient completes the initial setup required to get the proxy to a "steady" state.
-func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client, error) {
+func NewClient(ctx context.Context, l log.Logger, conf *Config) (*Client, error) {
```
kurtisvg, line 218: Are we sure this shouldn't stay a ref to command?

kurtisvg (changes requested):

cloudsql/cloudsql.go
```diff
@@ -32,3 +32,11 @@ type Dialer interface {
 
 	io.Closer
 }
+
+// Logger is the interface used throughout the project for logging.
+type Logger interface {
```
kurtisvg, line 37: This looks like a public interface now right? Does this mean it's in line for a breaking change?

internal/proxy/proxy.go
```diff
@@ -213,21 +216,21 @@ type Client struct {
...
 
 	// mnts is a list of all mounted sockets for this client
 	mnts []*socketMount
+
+	logger cloudsql.Logger
 }
 
 // NewClient completes the initial setup required to get the proxy to a "steady" state.
-func NewClient(ctx context.Context, cmd *cobra.Command, conf *Config) (*Client, error) {
+func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *Config) (*Client, error) {
```
kurtisvg, line 228: nit: I'm still at of a bit of a loss as to why this is preferable to just passing the cmd in

kurtisvg (approved):

cloudsql/cloudsql.go
enocom: Yes, it is public. I've added `Debugf` to the interface to support https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/1255. Since this interface is meant for google-internal usage only, I think we could technically break it by adding additional methods if necessary. But between info, error, and debug, we have everything we need now.

PR #1256, opened 2022-07-01
feat: add max-sigterm-delay flag

enocom:
This is the v2 version of term_timeout and will make proxy wait at most
for the specified duration beforing existing once a SIGTERM has been
received by the process.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -130,6 +130,10 @@ any client SSL certificates.`,
 	cmd.PersistentFlags().Uint32Var(&c.conf.MaxConnections, "max-connections", 0,
 		`Limits the number of connections by refusing any additional connections.
 When this flag is not set, there is no limit.`)
+	cmd.PersistentFlags().DurationVar(&c.conf.WaitOnClose, "wait-after-sigterm", 0,
+		`Amount of time to wait for any open connections to close before
+shutting down the proxy. When this flag is not set, the proxy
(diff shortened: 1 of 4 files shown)


enocom:
Putting this up to ensure it passes the tests while waiting to merge https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/1240.

cmd/root.go
```diff
@@ -127,6 +127,10 @@ any client SSL certificates.`,
 	cmd.PersistentFlags().Uint64Var(&c.conf.MaxConnections, "max-connections", 0,
 		`Limits the number of connections by refusing any additional connections.
 When this flag is not set, there is no limit.`)
+	cmd.PersistentFlags().DurationVar(&c.conf.WaitOnClose, "wait-after-sigterm", 0,
+		`Amount of time to wait for any open connections to close before
```
kurtisvg, line 131: Nit: consider "max-sigterm-delay"

Maybe change the wording to clarify the behavior as well:
"Maximum amount of time to wait after for any open connections after received a sigterm signal. The proxy will shutdown when open connections hits 0 or when the max time has passed. Defaults to 0s."

kurtisvg (approved):

cmd/root.go
enocom: That's a better name. I've slightly tweaked the description.

PR #1270, opened 2022-07-19
chore(ci): upgrade macOS runner to macos-latest and disable mac fuse tests

### LINKED ISSUE #1225: Upgrade macOS Github Actions runner from macos-10.15 to macos-latest
`macos-10.15` runner is catalina and can install `osxfuse` without having to restart the kernel. Where as `macos-latest` is Big Sur and can't  install `osxfuse` without requiring restart of the kernel.

Need to look into ways of using `osxfuse` or `macfuse` within github action builds that use `macos-latest` runner.

jackwotherspoon:
Updating test builds to `macos-latest` runner version which is macOS-11.

Seems like macOS-11 has tightened its ability around using [kernel extensions](https://support.apple.com/en-ca/guide/deployment/depa5fb8376f/web), making it currently not possible to install `macfuse` in the appropriate way on a github-hosted runner. (i.e. you can not restart the macOS runner into Reduced Security mode)

Closes #1225


enocom (approved):

cmd/cloud_sql_proxy/proxy_test.go
```diff
@@ -42,58 +42,137 @@ func TestCreateInstanceConfigs(t *testing.T) {
...
+			useFuse:                  false,
+			instances:                nil,
+			instancesSrc:             "md",
+			wantErr:                  true,
+			skipFailedInstanceConfig: false,
 		},
 	} {
-		if runtime.GOOS == "windows" && !v.supportedOnWindows {
+		// fuse is not supported in CI for darwin
+		if runtime.GOOS == "darwin" && v.useFuse {
```
kurtisvg, line 171: nit: Would it be better to use t.Run and t.Skip here?

kurtisvg (approved):

kurtisvg:
We should maybe follow up w/ GitHub to see if we can request the macos runners have this be enabled.

cmd/cloud_sql_proxy/proxy_test.go
enocom: I'll send a PR to clean this up.

PR #859, opened 2021-07-30
feat: add health checks to proxy

monazhn:
## Change Description

Introduce health checks to the Cloud SQL proxy client, allowing for the proactive and automatic mitigation of health-related issues. The health checks consist of **startup**, **liveness**, and **readiness** probing, with requests against the proxy container issued via HTTP.

## Checklist

- [x] Make sure to open an issue as a [bug/issue](https://github.com/GoogleCloudPlatform/cloudsql-proxy/issues/new/choose) 
  before writing your code!  That way we can discuss the change, evaluate 
  designs, and agree on the general idea.
- [x] Ensure the tests and linter pass
- [x] Appropriate documentation is updated (if necessary)

## Relevant issues:

- Fixes #137

PR #1271, opened 2022-07-19
feat: add support for health-check flag

enocom:


enocom:
Reading some more about healthchecks [here](https://stackoverflow.com/a/65858310/927514) (and in the linked items), I'm a bit worried about including dial attempts in a readiness check, but given that was the [initial implementation](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/871), I'm inclined to perserve it for now. The liveness probe, I think, should also return to the [original implementation](https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/859) where it's mostly a process health check.

internal/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,110 @@
...
+
+// HandleReadiness ensures the Check has been notified of successful startup,
+// that the proxy has not reached maximum connections, and that all connections
+// are healthy.
+func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	select {
+	case <-c.started:
```
hessjcg, line 77: This looks like the intention is to wait for another part of the application to call Check.NotifyStarted() before responding. There is no reason to wait for something in the application to occur before responding to a k8s check. These should respond immediately with success or failure application state.

Success:  immediately return 200 OK status 
Failure:  endpoint should immediately return a >= 400 status. 

As a hint to any human looking at the response, it might be nice to choose one of the following: 
- 503 Service Unavailable  - this check is failing for normal reasons
- 500 Internal Server Error - something unexpected is going wrong, check the logs

internal/proxy/proxy.go
```diff
@@ -298,9 +297,54 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
+func (c *Client) CheckConnections(ctx context.Context) error {
+	var (
+		wg    sync.WaitGroup
+		errCh = make(chan error, len(c.mnts))
+	)
+	for _, m := range c.mnts {
+		wg.Add(1)
+		go func(inst string) {
+			defer wg.Done()
+			conn, err := c.dialer.Dial(ctx, inst)
```
hessjcg, line 311: If we are doing dial and immediately close every 5 seconds to all the databases from all the pods, is this going to exceed any api request limits?

hessjcg (commented):
I have some questions about the concurrency behavior in the health check. Checks should respond immediately, not wait for an application state before responding.

internal/proxy/proxy.go
```diff
@@ -298,9 +297,54 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
+	var mErr MultiErr
+	for i := 0; i < len(c.mnts); i++ {
+		select {
+		case err := <-errCh:
+			mErr = append(mErr, err)
+		default:
+			continue
+		}
+	}
+	if len(mErr) > 0 {
```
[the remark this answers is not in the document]
enocom: When this feature was initially written, this was something we discussed. Your suggestion seems like a reasonable improvement. @kurtisvg what do you think?

internal/proxy/proxy.go
enocom: Not necessarily. The dialer re-uses connection info under the hood and only refreshes that info every ~hour. For large deployments, though, just using the proxy can hit up against API quota.

The bigger question is if we should be dialing instances at all in the readiness check. I've read that readiness shouldn't be checking connections to downstream services. What would a good readiness check be for the proxy?

internal/healthcheck/healthcheck.go
enocom: The default case in a channel read will do just that -- if there are no values available on `c.started`, we'll trigger the default and return 503.

enocom:
@hessjcg also we have some example settings here: https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/main/examples/k8s-health-check/proxy_with_http_health_check.yaml

internal/healthcheck/healthcheck.go
hessjcg: Ok, I misunderstood the golang select{} language feature. Looks like this will immediately send OK if c.started is closed and immediately send 503 otherwise. Nice work.  I learned something new.

internal/proxy/proxy.go
hessjcg: Well, it depends on what ready means in the case of the proxy? 

When the proxy is a standalone deployment to me the choice is clear: ready means "database connection is ok." 

With proxy as a sidecar container to an application, the application won't receive traffic until both the app container and the sidecar both indicate "ready." Competent kubernetes application developers will have their application also expose a readiness check that will test the database connection before allowing traffic to the application. 

We have two choices:

- **Option A:** Proxy only says it is "ready" when the database can connect. If app developer added a database check to the application's readiness check, then the proxy readiness check is redundant. If app developer does not not add a database connection, the developer is saved from routing traffic to an application that can't connect to its database.
- **Option B:** Proxy says it is "ready" when its configuration is fully loaded and ports are open and listening. If the app developer added a readiness check, then the app won't receive traffic until the database can connect. If not, the app will become "ready" and start receiving traffic regardless of the ability to connect to the db.

Unfortunately, not every developer is a competent kubernetes application developer. Most of our customers rely on us to know "the right thing to do." So I think Option A is a safer option for most customers.

internal/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if open, max := c.proxy.ConnCount(); max > 0 && open == max {
+		c.logger.Errorf("[Health Check] Readiness failed: %v", errMaxConnections)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte(errMaxConnections.Error()))
+		return
+	}
+
+	err := c.proxy.CheckConnections(ctx)
```
kurtisvg, line 92: I think it might be nice to return a list of failed instances here (and maybe why they failed).

It should maybe be logged as well.

internal/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,110 @@
...
+	case <-c.started:
+	default:
+		c.logger.Errorf("[Health Check] Readiness failed: %v", errNotStarted)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte(errNotStarted.Error()))
+		return
+	}
+
+	if open, max := c.proxy.ConnCount(); max > 0 && open == max {
+		c.logger.Errorf("[Health Check] Readiness failed: %v", errMaxConnections)
```
kurtisvg, line 86: nit: consider reporting the max number of connections already made

cmd/root_test.go
```diff
@@ -516,10 +516,6 @@ func TestNewCommandWithErrors(t *testing.T) {
 			desc: "when the iam authn login query param is bogus",
 			args: []string{"proj:region:inst?auto-iam-authn=nope"},
 		},
-		{
```
hessjcg, line 519: Why did you remove this testcase?

internal/proxy/proxy.go
```diff
@@ -298,9 +297,54 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
+		return mErr
+	}
+	return nil
+}
+
+// ConnCount returns the number of open connections and the maximum allowed
+// connections. Returns 0 when the maximum allowed connetions have not been set.
```
kurtisvg, line 340:
```suggestion
// connections. Returns 0 when the maximum allowed connections have not been set.
```

hessjcg (approved):
Nice work on the test coverage. LGTM.

internal/proxy/proxy.go
kurtisvg: Startup check covers "proxy has loaded the config and mounted it's listeners". 

Liveliness checks "the proxy is running"

Readiness checks are when the pod ready to accept traffic. I don't see a better way to do that than to test dial the instances.

internal/proxy/proxy.go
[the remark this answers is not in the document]
kurtisvg: My 2c is that if the user specified 5 instances, they probably want to connect to those 5 instances or else they wouldn't be wasting the quota. 

We talked about different ways to make this customizable (e.g. query parameters that require specific instances to pass or a min number to fail) but ultimately decided that we could add those if people ask for them. I think that's still the case.

kurtisvg (changes requested):

internal/proxy/proxy.go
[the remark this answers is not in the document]
hessjcg: @kurtisvg I'm convinced. Current implementation is the best way to go. We probably want to recommend that  when configured in kubernetes, the readiness check should fail 2 or 3 times in a row before the pod is marked "not ready." This way a transient connection error on a single database doesn't put the connection offline.

cmd/root.go
```diff
@@ -394,7 +398,11 @@ func runSignalWrapper(cmd *Command) error {
 		}()
 	}
 
-	shutdownCh := make(chan error)
+	mux := http.NewServeMux()
+	server := &http.Server{
+		Addr:    fmt.Sprintf("localhost:%s", cmd.httpPort),
+		Handler: mux,
```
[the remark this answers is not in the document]
enocom: We use this `mux` below to register `/metrics` or the heathcheck endpoints if the user has requested either. Granted we could use the `DefaultServeMux` with `http.Handle`, but for me it's more clear to create a mux and register endpoints with it.

cmd/root.go
```diff
@@ -465,18 +452,48 @@ func runSignalWrapper(cmd *Command) error {
...
+		hc := healthcheck.NewCheck(p, cmd.logger)
+		mux.HandleFunc("/startup", hc.HandleStartup)
+		mux.HandleFunc("/readiness", hc.HandleReadiness)
+		mux.HandleFunc("/liveness", hc.HandleLiveness)
+		notify = hc.NotifyStarted
+	}
+
+	// Start the HTTP server is either Prometheus or the health check was
+	// enabled.
+	if cmd.prometheusNamespace != "" || cmd.healthCheck {
```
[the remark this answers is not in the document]
enocom: Since we have to take different actions for prometheus vs health check, we do need two separate values. Nonetheless, looking again at prometheus, the namespace is optional, so I'll update the flag to be a boolean (and add a namespace flag separately), which cleans this up slightly.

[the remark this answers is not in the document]
enocom: On second thought, I'll make this change as a separate PR to keep things focused here.

cmd/root_test.go
enocom: This was a mistake. I'll restore this test.

internal/healthcheck/healthcheck.go
enocom: I've encapsulated that information in the `err` value here, which is a MultiError. Its contents include both the instance name and the reason the dial failed.

internal/healthcheck/healthcheck.go
enocom: I'll update the error to capture that information.

cmd/root.go
```diff
@@ -465,18 +452,48 @@ func runSignalWrapper(cmd *Command) error {
...
+			select {
+			case <-ctx.Done():
+				// Give the HTTP server a second to shutdown cleanly.
+				ctx2, _ := context.WithTimeout(context.Background(), time.Second)
+				if err := server.Shutdown(ctx2); err != nil {
+					cmd.logger.Errorf("failed to shutdown Prometheus HTTP server: %v\n", err)
+				}
+			}
+		}()
+		go func() {
```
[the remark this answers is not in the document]
enocom: Will do.

cmd/root.go
```diff
@@ -465,18 +452,48 @@ func runSignalWrapper(cmd *Command) error {
...
+	}
+
+	// Start the HTTP server is either Prometheus or the health check was
+	// enabled.
+	if cmd.prometheusNamespace != "" || cmd.healthCheck {
+		go func() {
+			select {
+			case <-ctx.Done():
+				// Give the HTTP server a second to shutdown cleanly.
+				ctx2, _ := context.WithTimeout(context.Background(), time.Second)
```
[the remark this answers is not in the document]
enocom: Nope. Will do.

cmd/root_test.go
enocom: Actually restoring the test I see why I deleted this: I've changed the behavior to just log a warning -- if a user sets http-port without prometheus or health check, we now tell them that we're ignoring http-port but otherwise keep going.

enocom:
I had to rebase v2 onto main with this PR open. As a result, there are a lot of commits here that already exist on v2. But the diff is correct and will cleanly apply onto v2 now.

cmd/root.go
```diff
@@ -225,18 +230,18 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 		return newBadCommandError("cannot specify --credentials-file and --gcloud-auth flags at the same time")
 	}
 
-	if userHasSet("http-port") && !userHasSet("prometheus-namespace") {
-		return newBadCommandError("cannot specify --http-port without --prometheus-namespace")
+	if userHasSet("http-port") && !userHasSet("prometheus-namespace") && !userHasSet("health-check") {
```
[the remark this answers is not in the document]
enocom: I think the two boolean statements are synonymous. In effect, we want this: if a user sets http port and does not set either related flag, we should log the error.

cmd/root.go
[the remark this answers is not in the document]
kurtisvg: So what I was trying to suggest was that the need for an http server is probably orthogonal to which specific httpPorts have been specified. 

So something like this:

```go
if cmd.prometheusNamespace != "" {
    cmd.needHttpServer = true
    // other health check stuff
}
if cmd.healthCheck {
    cmd.needHttpServer = true
    // other health check stuff
}

if cmd.needHttpServer {
     // start http server
}
```

(Consider this a nit)

kurtisvg (approved):

PR #1273, opened 2022-07-20
docs: update README for v2

enocom:

PR #1283, opened 2022-07-29
feat: improve help command output

enocom:


enocom:
Compare with `curl --help all` -- lots of short and to the point help information that doesn't wrap on small screens.

We have more to share for these flags, but I'm tried to condense help commands to the bare minimum without losing meaning.

cmd/root.go
```diff
@@ -100,16 +100,60 @@ func WithDialer(d cloudsql.Dialer) Option {
 	}
 }
 
+var longHelp = `
+The Cloud SQL Auth proxy provides IAM-based authorization and TLS 1.3 encryption
+when connecting to Cloud SQL instances. For every provided instance connection
+name, the proxy creates a local listener that will proxy traffic to and from your
+Cloud SQL instance using an encrypted connection based on an ephemeral certificate
+that is renewed by the proxy every hour.
```
kurtisvg, line 108:
```suggestion
The Cloud SQL Auth proxy is a utility for providing more secure connections to your Cloud SQL instances. It provides IAM Authorization (providing you control over who can or cannot connect to your instance) and TLS 1.3 encryption without having to manage certificates. 
```

cmd/root.go
```diff
@@ -100,16 +100,60 @@ func WithDialer(d cloudsql.Dialer) Option {
...
+
+With the proxy, clients need not manage SSL certificates manually and get the
+benefit of IAM authorization of all client connections.
```
kurtisvg, line 111: nit: would remove

cmd/root.go
```diff
@@ -100,16 +100,60 @@ func WithDialer(d cloudsql.Dialer) Option {
...
+
+To start the proxy, you will need your instance connection name which may be found
```
kurtisvg, line 113:
```suggestion
To start the proxy, you will need your instance connection name, which may be found
```

cmd/root.go
```diff
@@ -143,62 +187,57 @@ any client SSL certificates.`,
...
 	cmd.RunE = func(*cobra.Command, []string) error { return runSignalWrapper(c) }
 
+	// Override Cobra's default messages.
+	cmd.PersistentFlags().BoolP("help", "h", false, "Display help information for cloudsql-proxy")
+	cmd.PersistentFlags().BoolP("version", "v", false, "Print the cloudsql-proxy version")
+
 	// Global-only flags
 	cmd.PersistentFlags().StringVarP(&c.conf.Token, "token", "t", "",
-		"Bearer token used for authorization.")
+		"Use bearer token for authentication.")
```
kurtisvg, line 196: nit: Let's sidestep authZ vs authN here for now (same for the others) 
```suggestion
		"Use bearer token as a source of IAM credentials.")
```

kurtisvg (changes requested):

cmd/root.go
enocom: Slightly revised the suggestion and expanded the suggestion with this:

```
The Cloud SQL Auth proxy is a utility for ensuring secure connections to your                                                  
Cloud SQL instances. It provides IAM authorization, allowing you to control who                                                
can connect to your instance through IAM permissions, and TLS 1.3 encryption,                                                  
without having to manage certificates.                                                                                         
                                                                                                                               
For every provided instance connection name, the proxy creates:                                                                
                                                                                                                               
- a listener that mimics a database running locally, and                                                                       
- an encrypted connection using TLS 1.3 back to your Cloud SQL instance.                                                       
                                                                                                                               
The proxy uses an ephemeral certificate to establish a secure connection to your                                               
Cloud SQL instance. The proxy will refresh those certificates on an hourly                                                     
basis. Existing client connections are unaffected by the refresh cycle.                                                        
                                                                                                                               
NOTE: The proxy does not configure the network. You MUST ensure the proxy can                                                  
reach your Cloud SQL instance, either by deploying it in a VPC that has access                                                 
to your Private IP instance, or by configuring Public IP. 
```

cmd/root.go
enocom: Done.

cmd/root.go
enocom: Done.

cmd/root.go
```diff
@@ -100,16 +100,71 @@ func WithDialer(d cloudsql.Dialer) Option {
 	}
 }
 
+var longHelp = `
+The Cloud SQL Auth proxy is a utility for ensuring secure connections to your
+Cloud SQL instances. It provides IAM authorization, allowing you to control who
```
hessjcg, line 105: The proxy handles authorization to the database using Google Cloud IAM permissions. The proxy configures TLS 1.3 encryption for connections to the database so that your application does not need to manage certificates.

cmd/root.go
```diff
@@ -100,16 +100,71 @@ func WithDialer(d cloudsql.Dialer) Option {
...
+reach your Cloud SQL instance, either by deploying it in a VPC that has access
+to your Private IP instance, or by configuring Public IP.
+
+To start the proxy, you will need your instance connection name, which may be found
+in the Cloud SQL instance overview page or by using gcloud with the following
+command:
+
+    gcloud sql instances describe INSTANCE --format='value(connectionName)'
+
+Starting the proxy will look like, for example:
```
hessjcg, line 128: Use the following command to start the proxy. For example, if database connection name is `my-project:us-central:1-my-db-server`:

cmd/root.go
```diff
@@ -100,16 +100,71 @@ func WithDialer(d cloudsql.Dialer) Option {
...
+
+By default, the proxy will determine the database engine and start a listener
+on localhost using the default database engine's port, i.e., MySQL is 3306,
+Postgres is 5432, SQL Server is 1433. If multiple instances are specified which
+all use the same database engine, the first will be started on the default
+database port and subsequent instances will be incremented from there (e.g.,
+3306, 3307, 3308, etc). To disable this behavior (and reduce startup time), use
+the --port flag. All subsequent listeners will increment from the provided value.
+
+All listeners use localhost. To override this behavior, use the --address flag.
```
hessjcg, line 140: All socket listeners use the localhost network interface. To override this behavior, use the --address flag to choose the network interface.

cmd/root.go
```diff
@@ -143,62 +198,57 @@ any client SSL certificates.`,
...
 	cmd.PersistentFlags().Uint64Var(&c.conf.MaxConnections, "max-connections", 0,
-		`Limits the number of connections by refusing any additional connections.
-When this flag is not set, there is no limit.`)
+		"Limit the number of connections. Default is no limit.")
 	cmd.PersistentFlags().DurationVar(&c.conf.WaitOnClose, "max-sigterm-delay", 0,
-		`Maximum amount of time to wait after for any open connections
-to close after receiving a TERM signal. The proxy will shut
-down when the number of open connections reaches 0 or when
-the maximum time has passed. Defaults to 0s.`)
+		"Maximum time to wait for connections to close after receiving a TERM signal.")
```
hessjcg, line 217: "Maximum time to wait in seconds..."

cmd/root.go
```diff
@@ -143,62 +198,57 @@ any client SSL certificates.`,
...
 	cmd.PersistentFlags().StringVar(&c.telemetryProject, "telemetry-project", "",
-		"Enable Cloud Monitoring and Cloud Trace integration with the provided project ID.")
+		"Enable Cloud Monitoring and Cloud Trace with the provided project ID.")
 	cmd.PersistentFlags().BoolVar(&c.disableTraces, "disable-traces", false,
```
hessjcg, line 220: It would probably be clearer if we make the telemetry `traces` and `metrics` semantically positive default true rather than negative. (This would need to be in a different PR, obviously)

```
--telemetry-project=name   # project=name, traces on, metrics on
--telemetry-project=name --enable-traces=false  #project=name, traces off, metrics on
--telemetry-project=name --enable-metrics=false  #project=name, traces on, metrics off
```

cmd/root.go
```diff
@@ -143,62 +198,57 @@ any client SSL certificates.`,
...
-		"Disable Cloud Trace integration (used with telemetry-project)")
+		"Disable Cloud Trace integration (used with --telemetry-project)")
 	cmd.PersistentFlags().IntVar(&c.telemetryTracingSampleRate, "telemetry-sample-rate", 10_000,
-		"Configure the denominator of the probabilistic sample rate of traces sent to Cloud Trace\n(e.g., 10,000 traces 1/10,000 calls).")
+		"Configure the sample rate of traces sent to Cloud Trace. A smaller number means more traces.")
```
hessjcg, line 223: "Set the Cloud Trace sample rate. A number 1 and 10000, smaller means more traces."

cmd/root.go
```diff
@@ -143,62 +198,57 @@ any client SSL certificates.`,
...
-		"Prefix to use for Cloud Monitoring metrics.")
+		"Prefix for Cloud Monitoring metrics.")
 	cmd.PersistentFlags().BoolVar(&c.prometheus, "prometheus", false,
-		"Enable Prometheus HTTP endpoint /metrics")
+		"Enable Prometheus HTTP endpoint /metrics on localhost")
 	cmd.PersistentFlags().StringVar(&c.prometheusNamespace, "prometheus-namespace", "",
 		"Use the provided Prometheus namespace for metrics")
 	cmd.PersistentFlags().StringVar(&c.httpPort, "http-port", "9090",
-		"Port for the Prometheus server to use")
+		"Port for Prometheus and/or health check server")
```
hessjcg, line 233: Port for Prometheus and health check server."

cmd/root.go
```diff
@@ -143,62 +198,57 @@ any client SSL certificates.`,
...
 	cmd.PersistentFlags().BoolVar(&c.healthCheck, "health-check", false,
-		`Enables HTTP endpoints /startup, /liveness, and /readiness
-that report on the proxy's health. Endpoints are available on localhost
-only. Uses the port specified by the http-port flag.`)
+		"Enables HTTP endpoints /startup, /liveness, and /readiness on localhost.")
```
hessjcg, line 235: "Enable health check endpoints..."

hessjcg (changes requested):

cmd/root.go
enocom: Getting the wording right on this one is key. I think we start with a high level sentence (proxy as a utility for secure connections), and then expand with details (IAM auth, TLS 1.3, etc.). Because there's a user authentication step, people easily confuse the proxy's authorization with logging in to the database. I'm taking the wording from here: https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/1273/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5.

cmd/root.go
enocom: It's a question of whether `--enable-metrics=false` reads better than `--disable-metrics`. I find the `enable` and `false` words to require a careful reading.

```
--telemetry-project=name   # project=name, traces on, metrics on
--telemetry-project=name --disable-traces  #project=name, traces off, metrics on
--telemetry-project=name --disable-metrics  #project=name, traces on, metrics off
```

hessjcg (approved):

cmd/root.go
hessjcg: Oh, yeah. I see. This is good enough for now. I think we can improve on it 

This is how it is handled in  [`curl` man page](https://man7.org/linux/man-pages/man1/curl.1.html) 

```
       In general, all boolean options are enabled with --option and yet
       again disabled with --no-option. That is, you use the exact same
       option name but prefix it with "no-". However, in this list we
       mostly only list and show the --option version of them. (This
       concept with --no options was added in 7.19.0. Previously most
       options were toggled on/off through repeated use of the same
       command line option.)
```

jackwotherspoon (approved):

kurtisvg (approved):

cmd/root.go
```diff
@@ -100,16 +100,73 @@ func WithDialer(d cloudsql.Dialer) Option {
...
+To start the proxy, you will need your instance connection name, which may be found
+in the Cloud SQL instance overview page or by using gcloud with the following
+command:
+
+    gcloud sql instances describe INSTANCE --format='value(connectionName)'
+
+For example, if your instance connection name is "my-project:us-central1:my-db-server",
+starting the proxy will be:
+
+    ./cloudsql-proxy my-project:us-central1:my-db-server
```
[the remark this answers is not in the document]
enocom: We need to make a pass through the entire repo and settle on whatever we're calling this binary. I'll send that as a separate PR.

PR #1311, opened 2022-08-12
(DO NOT REVIEW) feat: add initial support for FUSE

enocom:
This PR adds:

The next PR(s) will include:


enocom:
We're going to have to skip macOS builds given the restrictions with adding kernal extensions and install fuse on the ubuntu image. I'll work on that next.

enocom:
This has some problems and isn't ready for review after all.

PR #1312, opened 2022-08-13
feat: add support for README in FUSE mode

enocom:
This PR adds:

There are four FUSE libraries for Go:

bazil.org/fuse -- a solid library, pure Go implementation, but Linux only.
github.com/jacobsa/fuse -- a fork of the library above. Linux and Darwin are supported.
github.com/hanwen/go-fuse -- pure Go implementation, Linux and Darwin are supported.
github.com/winfsp/cgofuse -- broad OS support (Linux, macOS, Windows), but requires CGO.
What we want:

Broad OS support -- especially Linux and macOS.
No CGO -- we compile cross-platform and static binaries. CGO would introduce a lot of complexity into our build process that we'd rather avoid.
Well maintained and active project.
By the criteria above, that leaves us with:

github.com/jacobsa/fuse and
github.com/hanwen/go-fuse

Option 2 has better performance, so we are moving forward with go-fuse. 

The next PR(s) will include:


internal/proxy/fuse.go
```diff
@@ -0,0 +1,73 @@
...
+
+const readmeText = `
+When programs attempt to open files in this directory, a remote connection to
+the Cloud SQL instance of the same name will be established.
+
+That is, running:
+
+	mysql -u root -S "/path/to/this/directory/project:region:instance-2"
+	-or-
+	psql "host=/path/to/this/directory/project:region:instance-2 dbname=mydb user=myuser"
```
kurtisvg, line 39: Should this mention the suffix postgres will add automatically?

internal/proxy/fuse.go
```diff
@@ -0,0 +1,73 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package proxy
+
+import (
+	"context"
+	"syscall"
+
+	"github.com/hanwen/go-fuse/v2/fs"
```
kurtisvg, line 21: How did we land on go-fuse?

kurtisvg (changes requested):

internal/proxy/fuse.go
enocom: It's what we used on v1 to gain support for macOS. I see that https://github.com/winfsp/cgofuse supports Windows in addition to macOS, and so I'll replace the implementation here as an experiment. A wider set of supported OS's seems better.

internal/proxy/fuse.go
enocom: Yes, that's a good point. Will do.

internal/proxy/fuse.go
enocom: I've revisited this decision and here's what I found:

There are four FUSE libraries for Go:

1. bazil.org/fuse -- a solid library, pure Go implementation, but Linux only. 
2. github.com/jacobsa/fuse -- a fork of the library above. Linux and Darwin are supported. 
3. github.com/hanwen/go-fuse -- pure Go implementation, Linx and Darwin are supported.
4. github.com/winfsp/cgofuse -- broad OS support (Linux, macOS, Windows), but requires CGO.

What we want:
1. Broad OS support -- especially Linux and macOS.
2. No CGO -- we compile cross-platform and static binaries. CGO would introduce a lot of complexity into our build process that we'd rather avoid.
3. Well maintained and active project.

By the criteria above, that leaves us with:

1. github.com/jacobsa/fuse and
2. github.com/hanwen/go-fuse

Option 2 seems more popular based on GitHub stars, but Option 1 seems more active and has a possibly better API.

I'd like to try re-implementing this PR with jacobsa/fuse to see if it's objectively better than go-fuse.

enocom: Reading through https://github.com/jacobsa/fuse/issues/78, I think for now we're good to stick with go-fuse. The fuse interface isn't exported, so we're in a position to rewrite the implementation in the future.

kurtisvg (approved):

PR #1319, opened 2022-08-22
chore: Update Dockerfiles and artifacts to build the v2 proxy

hessjcg:
The are updated to build the correct executable for the v2 proxy.


enocom (approved):

enocom:
Since we're fixing build stuff, might auditing everything in `.build`? Especially https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/v2/.build/gcs_upload.yaml.

PR #1345, opened 2022-08-30
chore: update v1 test config

enocom:


jackwotherspoon (changes requested):
Can we remove the Cloud SDK step from v1 builds? I don't think they need the Cloud SDK installed if I recall correctly?
https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/1345/files#diff-e06364554a6dd27166f6b3f2a377f7963d8554901280865922e59f56a584dccdR50-R51

jackwotherspoon (approved):

PR #1365, opened 2022-08-31
fix: support configuration of HTTP server address

### LINKED ISSUE #1359: [v2] prometheus and health checks listen on localhost instead of anyhost
## Bug Description

When started with `"--prometheus` and `--health-check`, the HTTP server listens for requests on localhost, but it should listen on anyhost. Otherwise, it cannon be accessed using the Pod IP (from Prometheus, kubelet, and Istio).

## How to reproduce

  1. Run the Pod with `--health-check`
  2. Create an ephemeral container with `debug` and run `curl localhost:9090/readiness` and `curl <pod IP>:9090/readiness`. The latter won't work.

## Environment

- GKE 1.23 with COS containerd nodes
- Cloud SQL Proxy version: docker container built from the v2.0.0.preview.0 tag, targeting Linux amd64

EDIT: I can open a pull request if you want. I tested it, and it worked by replacing `localhost` with `0.0.0.0` in the HTTP server.

enocom:
In Kubernetes, the convention is to bind HTTP probes and Prometheus endpoints to
0.0.0.0 (both lo and eth0). Since people might want to run this code on a GCE
VM, default to localhost, but otherwise support binding to both interfaces.

Fixes #1359.


enocom:
This probably should be changed such that Prometheus's endpoint is available on all interfaces, but the remaining endpoints are localhost only.

cmd/root_test.go
```diff
@@ -595,7 +595,12 @@ func TestPrometheusMetricsEndpoint(t *testing.T) {
 	// Keep the test output quiet
 	c.SilenceUsage = true
 	c.SilenceErrors = true
-	c.SetArgs([]string{"--prometheus", "my-project:my-region:my-instance"})
+	c.SetArgs([]string{
+		"--prometheus",
+		"--prometheus-address", "localhost",
```
hessjcg, line 600: You need a test where this isn't localhost. That's the default value

cmd/root.go
```diff
@@ -517,42 +527,17 @@ func runSignalWrapper(cmd *Command) error {
...
 		mux.HandleFunc("/readiness", hc.HandleReadiness)
 		mux.HandleFunc("/liveness", hc.HandleLiveness)
 		notify = hc.NotifyStarted
-	}
-
-	// Start the HTTP server if anything requiring HTTP is specified.
-	if needsHTTPServer {
-		server := &http.Server{
+		srv := &http.Server{
 			Addr:    fmt.Sprintf("localhost:%s", cmd.httpPort),
```
hessjcg, line 537: Should we be putting 0.0.0.0 here?

cmd/root.go
```diff
@@ -78,6 +78,8 @@ type Command struct {
 	telemetryPrefix            string
 	prometheus                 bool
 	prometheusNamespace        string
+	prometheusAddress          string
+	prometheusPort             string
```
daquinoaldo, line 82: [minor] Since it’s prometheusPort, what about healthCheckPort instead of httpPort?

cmd/root.go
```diff
@@ -78,6 +78,8 @@ type Command struct {
 	telemetryPrefix            string
 	prometheus                 bool
 	prometheusNamespace        string
+	prometheusAddress          string
```
daquinoaldo, line 81: We also need httpAddress, right? Otherwise, the probe will fail (am I missing something?).

enocom:
After going back and forth on this, I think the convention is to bind HTTP probes to 0.0.0.0 (both lo and eth0) in a kubernetes context. Since people might want to run this on a GCE VM, I'll leave the default to localhost, but otherwise support binding to both interfaces.

See [the docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes) where it says that the kublet defaults to the pod ID for HTTP probes,

daquinoaldo (approved):
Sounds perfect! 🙌

hessjcg (approved):
LGTM

PR #1366, opened 2022-08-31
fix: pass dial options when checking connections

### LINKED ISSUE #1360: [v2][healthcheck] instance does not have IP of type \"PUBLIC\" when running with `--private-ip`
## Bug Description

I get the following error on `/readiness` although I set the `--private-ip` flag.
```
Readiness failed: Config error: instance does not have IP of type \"PUBLIC\"
```

However, the connection works. It's just the `/readiness` endpoint that is failing.

## How to reproduce

  1. Run with the `--private-ip` flag specifying an instance with no public IP
  2. `curl localhost:9090/readiness`

## Environment

1. GKE v1.23 with Workload Identity
3. Cloud SQL Proxy version: docker container built from the v2.0.0.preview.0 tag, targeting Linux amd64

enocom:
Fixes #1360.


hessjcg (approved):
Lgtm

PR #1370, opened 2022-09-01
Allow for project & region discovery

davidcollom:
## Change Description

We have a usecase in which we have multiple CloudSQL instances across various regions, in which we never want to have cross region communications.
Currently when using `-dir` and `-project` we have an all or nothing approach.

With the changes proposed here, we are then able to scope project & region as a limiting factor, only creating sockets for those CloudSQL instances within the region in which cloudsql-proxy is running within.

## Checklist

### DIFF
--- a/cmd/cloud_sql_proxy/cloud_sql_proxy.go
+++ b/cmd/cloud_sql_proxy/cloud_sql_proxy.go
@@ -239,10 +239,11 @@ Connection:
      the Proxy.
 
   -projects
-    To direct the proxy to allow connections to all instances in specific
-    projects, set the projects parameter:
+    To direct the proxy to allow connections to all (or region) instances in
+    specific projects, Set the projects parameter:
 
         -projects=my-project
+        -projects=my-project:my-region
 
   -fuse
     If your local environment has FUSE installed, you can specify the -fuse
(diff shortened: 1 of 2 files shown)


enocom:
Thanks for the PR @davidcollom. We normally like to start stuff like this with an issue, but it looks like a small change.

In v2 we purposely decided not to port the `projects` flag given its chance of user error (e.g., woops, I didn't want to enable a connection for this instance). Out of curiosity why don't you use explicit instance connection names to solve your use case?

davidcollom:
Hey @enocom I noticed the issue first once I'd already created the code 🤦‍♂️.

We don't want to specifically name the instances as we dynamically scale up/down read-replicas along with instances as we scale applications within that region.

Additionally we'd like to not restart the proxy when we do this as our applications as this could impact multiple applications at a time and/or additional upstream failures (queries in flight etc)

enocom:
Would you mind filing an issue so we can talk about it there? I'd like to land support for what you're requesting on v2 ideally. 

Note we have an issue for supporting dynamic config which might be what you're looking for here: https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/1045.

Either way, right now as written this change only works when the proxy starts up by connecting to instances in the same region.

PR #1373, opened 2022-09-03
feat: add support for FUSE connections

enocom:
This commit also ensures that closing the proxy.Client blocks until all
listeners are closed.

### DIFF
--- a/internal/proxy/fuse_test.go
+++ b/internal/proxy/fuse_test.go
@@ -20,13 +20,15 @@ package proxy_test
 import (
 	"context"
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
-	"github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/internal/log"
+	"github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/cloudsql"
 	"github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/internal/proxy"
+	"github.com/hanwen/go-fuse/v2/fs"
 )
 
 func randTmpDir(t interface {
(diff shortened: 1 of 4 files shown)


internal/proxy/fuse_test.go
```diff
@@ -110,13 +101,212 @@ func TestREADME(t *testing.T) {
...
+			fuseTempDir:  fuseTempDir,
+		},
+		{
+			desc:         "postgres connections create a directory with a special file",
+			wantInstance: "proj:region:pg",
+			socketPath:   filepath.Join(fuseDir, "proj:region:pg", ".s.PGSQL.5432"),
+			fuseTempDir:  fuseTempDir,
+		},
+		{
+			desc:         "connecting creates intermediate temp directories",
```
kurtisvg, line 153: Should we have some failures cases in here?

What if we dial an instance that doesn't exist? what happens?

internal/proxy/proxy.go
```diff
@@ -34,6 +36,44 @@ import (
 	"golang.org/x/oauth2"
 )
 
+var (
+	// Instance connection name is the format <PROJECT>:<REGION>:<INSTANCE>
+	// Additionally, we have to support legacy "domain-scoped" projects (e.g. "google.com:PROJECT")
+	connNameRegex = regexp.MustCompile("([^:]+(:[^:]+)?):([^:]+):([^:]+)")
```
kurtisvg, line 42: Should we consider just exposing these from the Go connector?

kurtisvg (approved):

internal/proxy/fuse_test.go
kurtisvg: I found failure cases below, closing this.

internal/proxy/proxy.go
enocom: Yes, I'll do that.

PR #1388, opened 2022-09-08
chore: configure pipefail to catch failures

enocom:
By explicitly configuring `shell: bash` in test step it enables `set -eo pipefail` ([documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference)) which means errors in left side of pipe will no longer be suppressed by the success of the right side of the pipe.
```yaml
go test -race -v ./... | tee test_results.txt
```

Additionally, with new `pipefail` being set, if line that runs the test `go test -race -v ./... | tee test_results.txt` fails (i.e. a test fails) than this causes an exit code and for the line below it `go-junit-report ...` not to be run which in turn disables flakybot from working as it requires XML sponge log.

By moving the convert XML functionality to its own step in the workflow we can make sure it always runs no matter what happens with the tests.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -119,6 +119,7 @@ jobs:
           SQLSERVER_DB: '${{ steps.secrets.outputs.SQLSERVER_DB }}'
           TMPDIR: "/tmp"
           TMP: '${{ runner.temp }}'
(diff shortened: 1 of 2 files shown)


enocom:
Closing this since @jackwotherspoon did all the work and will re-open a matching PR.

PR #1393, opened 2022-09-08
chore: adjust release-please configuration to properly handle v1

hessjcg:
Apply explicit configuration to the v1 branch so that it uses 'simple' strategy and 'CHANGELOG.md' for changes.

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -16,9 +16,7 @@ handleGHRelease: true
 packageName: cloud-sql-proxy
 releaseType: simple
 versionFile: 'cmd/version.txt'
-changelogPath: 'CHANGELOG.md'
 branches:
 - branch: v1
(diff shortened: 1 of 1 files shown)


.github/release-please.yml
```diff
@@ -16,6 +16,9 @@ handleGHRelease: true
 packageName: cloud-sql-proxy
 releaseType: simple
 versionFile: 'cmd/version.txt'
+changelogPath: 'CHANGELOG.md'
```
enocom, line 19: This and the one below are probably unnecessary. Perhaps just adding releaseType will do the trick as it has no default.

enocom (changes requested):

.github/release-please.yml
hessjcg: ok, we'll try without the explicit setting.

enocom (approved):

PR #1405, opened 2022-09-13
fix: set write permissions for group and other

### LINKED ISSUE #1403: Permission denied when trying to query mysql with unix socket and credential file.
We upgrading to the last version of the cloud-sql-proxy, but we can't get it to work right, when we trying to query to mysql instance get the **[SQLSTATE[HY000] [2002] Permission denied]** error.

## Bug Description

We use the latest version of cloud-sql-proxy installed like this: 

```
go install github.com/GoogleCloudPlatform/cloud-sql-proxy/v2@latest
```

for start the instance we use credential file, we also create a user with **'cloudsqlproxy~%'** hostname to allow connect to the instance.

```
./cloud-sql-proxy -u /cloud-sql/ -c /cloud-sql-service-account PROJECT_NAME:us-central1:INSTANCE_NAME
```

We have a successfully start like this:

```
2022/09/12 15:46:27 Authorizing with the credentials file at "/cloud-sql-service-account"
2022/09/12 15:46:28 [PROJECT_NAME:us-central1:INSTANCE_NAME] Listening on /cloud-sql/PROJECT_NAME:us-central1::INSTANCE_NAME
2022/09/12 15:46:28 The proxy has started successfully and is ready for new connections!
```

But when the application tries to query mysql for any information using the unix socket, we get a error:

```
[2022-09-12 15:47:52] testing.ERROR: SQLSTATE[HY000] [2002] Permission denied (SQL: ...)
```

We tried with oldest cloud-sql-proxy versions and this problem doesn't happens, like:

```
go install github.com/GoogleCloudPlatform/cloudsql-proxy/cmd/cloud_sql_proxy@v1.31.2
```

## Environment

1. OS type and version: alpine-3.15
2. Cloud SQL Proxy version (`./cloud-sql-proxy --version`): cloud-sql-proxy version 2.0.0-preview.1

enocom:
Fixes #1403.

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -619,12 +619,19 @@ func newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst Inst
 			}
 			address = UnixAddress(address, ".s.PGSQL.5432")
 		}
+
 	}
 
 	lc := net.ListenConfig{KeepAlive: 30 * time.Second}
 	ln, err := lc.Listen(ctx, network, address)
 	if err != nil {
 		return nil, err
+	}
+	// Change file permisions to allow access for user, group, and other.
+	if network == "unix" {
+		// Best effort. If this call fails, group and other won't have write
+		// access.
+		_ = os.Chmod(address, 0777)
 	}
 	opts := conf.DialOptions(inst)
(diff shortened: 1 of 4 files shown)


hessjcg (approved):
LGTM

kurtisvg (approved):

PR #1408, opened 2022-09-13
fix: restore Go 1.16 support

### LINKED ISSUE #1407: Cloud functions (Gen 1) with 1.32.0 do not deploy.
## Bug Description

Updated version from `1.31.2` to `1.32.0`
This brought in `golang.org/x/net v0.0.0-20220909164309-bea034e7d591`

Test ran locally without issue, but failed to deploy, error shown in stack trace.

I suspect that this may be due to Gen 1's Go version being pegged at 1.16.7

If I drop back to 1.31.2 then I have no problem with deployment.

## Example code (or command)

```
gcloud functions deploy ReprocessingDispatcher \
--region=${REGION} \
--project=${PROJ} \
--timeout=500s \
--runtime go116 \
--max-instances=1 \
--set-env-vars ${ENV_VARS},GCF=1,PAGE_SIZE=5 \
--trigger-http
```

## Stacktrace
```
ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Build failed: # go.uber.org/atomic
vendor/go.uber.org/atomic/error.go:55:12: x.v.CompareAndSwap undefined (type Value has no field or method CompareAndSwap)
vendor/go.uber.org/atomic/error.go:61:24: x.v.Swap undefined (type Value has no field or method Swap)
vendor/go.uber.org/atomic/string.go:58:12: x.v.CompareAndSwap undefined (type Value has no field or method CompareAndSwap)
vendor/go.uber.org/atomic/string.go:64:12: x.v.Swap undefined (type Value has no field or method Swap); Error ID: 1a663392
```

## How to reproduce

  1. Update mod.go from `github.com/GoogleCloudPlatform/cloudsql-proxy v1.31.2` to `github.com/GoogleCloudPlatform/cloudsql-proxy v1.32.0`.
  2. Attempt to deploy.

## Environment

1. OS type and version: Google Cloud Function Gen 1.
2. Cloud SQL Proxy version (`./cloud-sql-proxy --version`): 1.32.0

enocom:
Update v1 deps only and not transitive deps with:

```
go get $(go list -f '{{if not (or .Main .Indirect)}}{{.Path}}{{end}}' -m all)
```

Fixes #1407.


enocom:
I haven't reproduced the build failure from #1407 locally yet and would like to do so before merging this, to confirm this is in fact the correct fix.

hessjcg (approved):
LGTM. I'll update the release script.

hessjcg (approved):
LGTM

PR #1435, opened 2022-09-22
docs: add disaster recovery example

ruyadorno:
## Change Description

Adds a bash script example that describes how to use **Secret Manager** along with `cloud-sql-proxy` in order to create a disaster recovery mechanism for replacing a primary **Cloud SQL** instance with a replica when the primary instance goes down.

Also adds a `cloud_sql_proxy_secret_manager_failover` tag that can be used for linking this example from **Cloud SQL** docs.

cc @enocom @kurtisvg


examples/disaster-recovery/README.md
```diff
@@ -0,0 +1,49 @@
...
+2. Trigger a promotion of an existing read replica to become a primary
+3. Update a Secret Manager secret with the name of the current primary
+
+## Restart Auth proxy when secret changes
+
+This option uses a wrapper script around the Cloud SQL Auth proxy to detect
+when the secret has changed, and restart the proxy with the new value. This
+could be done in many languages, but here’s an example using bash:
+
+```sh
```
kurtisvg, line 17: Can we include a link to the file that contains this snippet? e.g. :https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/2a9c8d8cb24e2c84e43620dab333677191d1dbd7/examples/k8s-sidecar/README.md?plain=1#L24

examples/disaster-recovery/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Get the latest version of the secret and start the proxy
+INSTANCE=$(gcloud secrets versions access "latest" --secret="$SECRET_ID")
+cloud_sql_proxy -instances="$INSTANCE"=tcp:"$PORT" &
+PID=$!
+
+# Every 5s, get the latest version of the secret. If it's changed, restart the
+# proxy with the new value.
+while true; do
+    sleep 5
```
kurtisvg, line 31: nit: Should sleep be configurable? Some users might want to adjust

examples/disaster-recovery/README.md
ruyadorno: good call 👍

examples/disaster-recovery/README.md
ruyadorno: I'll make it so 👍 I believe it helps having all the configurable params up there at the top 😊

examples/disaster-recovery/failover.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Get the latest version of the secret and start the proxy
+INSTANCE=$(gcloud secrets versions access "latest" --secret="$SECRET_ID")
+cloud_sql_proxy -instances="$INSTANCE"=tcp:"$PORT" &
+PID=$!
+
+# Every 5s, get the latest version of the secret. If it's changed, restart the
+# proxy with the new value.
+while true; do
+    sleep 5
```
enocom, line 29: Let's match the README making this configurable.

enocom (approved):

Issue #1438, opened 2022-09-26
Proxy does not build on FreeBSD as of preview.1 release


mhoran:
As with  #1390, the build fails on FreeBSD after the addition of the FUSE dependency. The build fails as follows:

```
# github.com/hanwen/go-fuse/v2/fuse
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/types.go:550:2: undefined: Attr
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/types.go:578:2: undefined: Attr
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:224:41: undefined: GetAttrIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:225:41: undefined: SetAttrIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:228:39: undefined: MknodIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:252:42: undefined: SetXAttrIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:258:40: undefined: CreateIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:260:38: undefined: ReadIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:269:39: undefined: WriteIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:278:41: undefined: ReadIn
go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.1.0/fuse/api.go:278:41: too many errors
```

enocom:
Thanks for the report @mhoran. I'm surprised that go-fuse doesn't support FreeBSD. Nonetheless, we'll put this behind a build constraint in the next release.

enocom:
We will get this out in the next release.

PR #1442, opened 2022-09-26
fix: restore openbsd and freebsd support

enocom:
Related to #1438.


internal/proxy/unix.go
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+
+//go:build !windows
+// +build !windows
+
+package proxy
+
+import "path/filepath"
+
+// UnixAddress is defined as a function to distinguish between Linux-based
```
hessjcg, line 22: nit: s/Linux-based/Unix-based/

hessjcg (approved):
Only one small nit.

PR #1450, opened 2022-09-30
feat: Downscope OAuth2 token included in ephemeral certificate

### LINKED ISSUE #1446: Downscope OAuth2 token included in ephemeral certificate
The token included in the ephemeral cert should have just a https://www.googleapis.com/auth/sqlservice.login scope attached to it.

enocom:
In the past, when using Auto IAM AuthN, the token embedded into the ephemeral certificate would have a broad scope. This commit ensures that in most cases, the embeddeded token has only a sqlservice.login scope.

When users provide a token to launch the proxy, the proxy will use that token with whatever scopes are attached.

Fixes #1446.


hessjcg (approved):
LGTM

PR #1460, opened 2022-10-07
feat: add support for impersonation

### LINKED ISSUE #417: Support for --impersonate-service-account
It would be nice, if Cloud SQL Proxy supports the `--impersonate-service-account` flag similar to the corresponding flag in `gcloud`. This would help to use impersonation out-of-the-box instead of long-running and possibly non-personalized service-account credentials.

enocom:
Note: the CLI flag matches the gcloud UI.

Fixes #417

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -82,6 +82,13 @@ type Command struct {
 	healthCheck                bool
 	httpAddress                string
 	httpPort                   string
+
+	// impersonationChain is a comma separated list of one or more service
+	// accounts. The last entry in the chain is the impersonation target. Any
+	// additional service accounts before the target are delegates. The
+	// roles/iam.serviceAccountTokenCreator must be configured for each account
+	// that will be impersonated.
+	impersonationChain string
 }
 
 // Option is a function that configures a Command.
@@ -253,6 +260,9 @@ https://cloud.google.com/storage/docs/requester-pays`)
 	cmd.PersistentFlags().StringVar(&c.conf.FUSETempDir, "fuse-tmp-dir",
 		filepath.Join(os.TempDir(), "csql-tmp"),
 		"Temp dir for Unix sockets created with FUSE")
+	cmd.PersistentFlags().StringVar(&c.impersonationChain, "impersonate-service-account", "",
+		`Comma separated list of service accounts to impersonate. Last value
+is the target account.`)
 
 	// Global and per instance flags
 	cmd.PersistentFlags().StringVarP(&c.conf.Addr, "address", "a", "127.0.0.1",
@@ -338,12 +348,25 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 	if userHasSet("sqladmin-api-endpoint") && conf.APIEndpointURL != "" {
 		_, err := url.Parse(conf.APIEndpointURL)
 		if err != nil {
-			return newBadCommandError(fmt.Sprintf("the value provided for --sqladmin-api-endpoint is not a valid URL, %v", conf.APIEndpointURL))
+			return newBadCommandError(fmt.Sprintf(
+				"the value provided for --sqladmin-api-endpoint is not a valid URL, %v",
+				conf.APIEndpointURL,
+			))
 		}
 
 		// add a trailing '/' if omitted
 		if !strings.HasSuffix(conf.APIEndpointURL, "/") {
 			conf.APIEndpointURL = conf.APIEndpointURL + "/"
+		}
+	}
+
+	if cmd.impersonationChain != "" {
+		accts := strings.Split(cmd.impersonationChain, ",")
+		conf.ImpersonateTarget = accts[0]
+		// Assign delegates if the chain is more than one account.
+		if l := len(accts); l > 1 {
+			conf.ImpersonateTarget = accts[l-1]
+			conf.ImpersonateDelegates = accts[:l-1]
 		}
 	}
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -362,11 +362,14 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 
 	if cmd.impersonationChain != "" {
 		accts := strings.Split(cmd.impersonationChain, ",")
-		l := len(accts)
-		conf.ImpersonateTarget = accts[l-1]
-		// Assign delegates if the chain is more than one account.
-		if l > 1 {
-			conf.ImpersonateDelegates = accts[:l-1]
+		conf.ImpersonateTarget = accts[0]
+		// Assign delegates if the chain is more than one account. Delegation
+		// goes from last back towards target, e.g., With sa1,sa2,sa3, sa3
+		// delegates to sa2, which impersonates the target sa1.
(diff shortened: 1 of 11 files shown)


enocom:
~361 lines of this PR are just cleaning up our e2e authentication tests to be table tests, and adding a matching impersonation test for each authentication type.

enocom:
Also, I've updated our CI pipeline to support impersonation as well.

hessjcg (approved):
LGTM

cmd/root.go
```diff
@@ -347,6 +360,16 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 		}
 	}
 
+	if cmd.impersonationChain != "" {
+		accts := strings.Split(cmd.impersonationChain, ",")
```
kurtisvg, line 364: Can this ever be an empty slice?

cmd/root.go
```diff
@@ -347,6 +360,16 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
...
+		conf.ImpersonateTarget = accts[0]
+		// Assign delegates if the chain is more than one account.
+		if l := len(accts); l > 1 {
+			conf.ImpersonateTarget = accts[l-1]
+			conf.ImpersonateDelegates = accts[:l-1]
+		}
```
kurtisvg, line 370: nit:
```suggestion
                 l := len(accts)
		conf.ImpersonateTarget = accts[l-1]
		// Assign delegates if the chain is more than one account.
		if l > 1 {
			conf.ImpersonateDelegates = accts[:l-1]
		}
```

internal/proxy/proxy.go
```diff
@@ -187,38 +199,89 @@ func (c *Config) DialOptions(i InstanceConnConfig) []cloudsqlconn.DialOption {
...
+		default:
+			l.Infof("Impersonating service account with Application Default Credentials")
+		}
+		ts, err := impersonate.CredentialsTokenSource(
+			context.Background(),
+			impersonate.CredentialsConfig{
+				TargetPrincipal: c.ImpersonateTarget,
+				Delegates:       c.ImpersonateDelegates,
+				Scopes: []string{
+					sqladmin.CloudPlatformScope,
```
kurtisvg, line 235: Do we need CloudPlatform here?

internal/proxy/proxy.go
```diff
@@ -187,38 +199,89 @@ func (c *Config) DialOptions(i InstanceConnConfig) []cloudsqlconn.DialOption {
 	return opts
 }
 
-// DialerOptions builds appropriate list of options from the Config
-// values for use by cloudsqlconn.NewClient()
-func (c *Config) DialerOptions(l cloudsql.Logger) ([]cloudsqlconn.Option, error) {
-	opts := []cloudsqlconn.Option{
-		cloudsqlconn.WithUserAgent(c.UserAgent),
+func (c *Config) credentialsOpt(l cloudsql.Logger) (cloudsqlconn.Option, error) {
```
kurtisvg, line 202: This doesn't look like it modifies c in anyway. Would it be better to take c as an arg instead?

kurtisvg (changes requested):

cmd/root.go
```diff
@@ -253,6 +260,9 @@ https://cloud.google.com/storage/docs/requester-pays`)
 	cmd.PersistentFlags().StringVar(&c.conf.FUSETempDir, "fuse-tmp-dir",
 		filepath.Join(os.TempDir(), "csql-tmp"),
 		"Temp dir for Unix sockets created with FUSE")
+	cmd.PersistentFlags().StringVar(&c.impersonationChain, "impersonate-service-account", "",
+		`Comma separated list of service accounts to impersonate. Last value
+is the target account.`)
```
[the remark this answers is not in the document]
enocom: Good question. I checked the gcloud docs and the Impersonate docs again. Both mention service accounts only and not IAM identities. Cf. https://pkg.go.dev/google.golang.org/api/impersonate#CredentialsConfig.

cmd/root.go
enocom: Only if a user passed this: `--impersonate-service-account=,`.

internal/proxy/proxy.go
enocom: Let me check on that -- we did for the e2e tests using the token.

internal/proxy/proxy.go
enocom: Yes, I'm in favor of that idea.

enocom: Looking at this again, I see we're reading a bunch of data from `c`. So even though we're not modifying it, it's still common to define `c` as a receiver. Passing it as an argument would basically be a de-sugared version of the same code here.

cmd/root.go
enocom: Done.

internal/proxy/proxy.go
enocom: We don't need it here. It's the token source that gets the impersonated token that needs this scope. Removed it here, and removed the SQL Admin scope from the token source in the tests.

enocom:
I need to verify the delegation logic here. I've matched gcloud's behavior, but need to check if the Go library does the same (back to front).

internal/proxy/proxy.go
kurtisvg: downgrade to nit: but it does seem like a receiver indicates we are doing something to `c`, where a read only parameter makes clear in the intent is only to create something from c.

tests/connection_test.go
```diff
@@ -35,7 +34,9 @@ const connTestTimeout = time.Minute
 // and then unsets GOOGLE_APPLICATION_CREDENTIALS. It returns a cleanup function
 // that restores the original setup.
 func removeAuthEnvVar(t *testing.T) (*oauth2.Token, string, func()) {
-	ts, err := google.DefaultTokenSource(context.Background(), sqladmin.SqlserviceAdminScope)
+	ts, err := google.DefaultTokenSource(context.Background(),
+		"https://www.googleapis.com/auth/cloud-platform",
```
kurtisvg, line 38: Why do we need cloud-platform here?

kurtisvg (approved):

internal/proxy/proxy.go
enocom: Let me try that in a separate PR and we can apply the pattern across the board if we like it.

tests/connection_test.go
enocom: Short answer: the impersonation fails without this scope. Longer answer: let me figure out if we can limit this scope still further.

enocom: After some digging, I found that this is the required OAuth2 scope to call generateAccessToken.


See:
- https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
- https://github.com/googleapis/google-api-go-client/blob/main/iamcredentials/v1/iamcredentials-gen.go#L85

internal/proxy/proxy.go
enocom: https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/1480

PR #1461, opened 2022-10-07
chore: improve golangci config

enocom:


internal/proxy/alignment_test.go
```diff
@@ -23,6 +23,8 @@ func TestClientUsesSyncAtomicAlignment(t *testing.T) {
 	// The sync/atomic pkg has a bug that requires the developer to guarantee
 	// 64-bit alignment when using 64-bit functions on 32-bit systems.
 	c := &Client{}
+	// linter can't see the read below of c.connCount, so just make it happy
```
hessjcg, line 26: Can you figure out how to make the linter ignore this line using a comment instead of doing a workaround? I'm sure there will be more than one case where we need an exception to the lint rules.

hessjcg (changes requested):
See if you can figure out how to add a code comment as an exception to the finding. Otherwise looks great.

internal/proxy/alignment_test.go
enocom: Sure. It's just `//nolint:linter-name`. Ideally, we don't pepper these around the codebase and prefer Go-based solutions as possible. But since this is just a test, a lint directive is probably just as well.

hessjcg (approved):
LGTM. 
I think the guideline is to only use `//nolint:...` when the linter is throwing a false positive. 

In this case, the linter is misinterpreting the program flow. There will be other cases where the linter can't account for external context, i.e. we are forced into less-than-optimal coding practices by an external library.

PR #1472, opened 2022-10-11
chore: add govulncheck to CI

enocom:


enocom:
We need to do a couple of additional things here:

1. Add flakybot so that failures result in open issues
2. Add PR label configuration

PR #1496, opened 2022-10-19
feat: add support for min ready instances

### LINKED ISSUE #1375: Readiness check supports configuring minimum number of ready instances
## Feature Description
- Allow specify a number of least necessary connected cloud-sql replica for unhealthy condition.

## Alternatives Considered

- Create new cloud-sql-proxy for each cloud-sql (read) replica and make special script for checking ready cloud-sql-proxies.
  - it need to up a lot of other cloud-sql-proxy instances.
  - it mess up readiness probe definitions.

## Additional Context
- If you are using multiple cloud-sql replicas, you can use `query retry` strategies in your application code.
  - 1. try to query for 1st instance in the (randomized) list.
  - 2. if it doesn't work, retry the query in 2nd instance in the list. (continue to last of the list)
  - 3. when correct response has come, you can use the values.
- Such query strategy is making our application robust for downing some replicas. (it will avoid a read replica will be a single point of failure)
- But, current cloud-sql-proxy's readiness check will be unhealthy if single node was down.
- In Kubernetes, this behavior results the Pod's down if single cloud-sql replica has down the pod cannot handle all requests regardless our application can response it (by using another replica.)
  - This means, the retry strategy is not working with cloud-sql-proxy and a single read replica will be a single point of failure.
- If there is this feature, we can use this for avoiding such system down.

enocom:
Fixes #1375.


kurtisvg:
Would this make more sense as a query parameter on the endpoint? Might be less invasive if the change is limited there as well

enocom:
Great idea. I love it.

internal/healthcheck/healthcheck.go
```diff
@@ -88,8 +89,24 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
+		}
+	}
+
+	n, err := c.proxy.CheckConnections(ctx)
+	if minReady > n {
+		c.logger.Errorf("[Health Check] min-ready was greater than registered instances")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if err != nil && !ready(err, minReady, n) {
```
hessjcg, line 109: This is kind of a strange condition statement. It took me a few minutes to figure out why it's ok to return an error if err != nil and minReady == 0. 

I'd move the `err != nil` condition inside ready() so that the logic reads more like a story and less like a CS 101 puzzle.

internal/healthcheck/healthcheck.go
```diff
@@ -88,8 +89,24 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
 		return
 	}
 
-	err := c.proxy.CheckConnections(ctx)
-	if err != nil {
+	var minReady int
```
hessjcg, line 92: It is confusing that when the `min-ready` request url param is not set or has an error, then minReady = 0. If minReady = 0. It would be clearer if minReady = n by default.

internal/healthcheck/healthcheck_test.go
```diff
@@ -239,3 +257,69 @@ func TestHandleReadinessWithConnectionProblems(t *testing.T) {
...
+			desc:       "when min ready is bogus",
+			minReady:   "bogus",
+			wantStatus: http.StatusServiceUnavailable,
+		},
+		{
+			desc:       "when min ready is greater than registered instances",
+			minReady:   "100",
+			wantStatus: http.StatusBadRequest,
+		},
+	}
```
hessjcg, line 292: How about a test case for when. min-ready is blank.

hessjcg (changes requested):
Needs a unit test for when min-ready is not set. Optional: the ready() logic could be simplified and made clearer.

internal/healthcheck/healthcheck.go
```diff
@@ -102,12 +117,18 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
-func ready(err error, minReady uint64) bool {
+func ready(err error, minReady, total int) bool {
 	mErr, ok := err.(proxy.MultiErr)
 	if !ok {
 		return false
 	}
-	return minReady != 0 && minReady <= uint64(len(mErr))
+	if minReady == 0 {
+		return false
+	}
```
kurtisvg, line 127: Why a special case for minReady? Shouldn't areReady always >= minReady if min is 0?

internal/healthcheck/healthcheck.go
```diff
@@ -90,8 +89,24 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
+		n, err := strconv.Atoi(v)
+		// If the query value parsed correctly, set it to minReady. Otherwise,
+		// ignore it.
+		if err == nil {
+			minReady = n
+		}
+	}
+
+	n, err := c.proxy.CheckConnections(ctx)
+	if minReady > n {
```
kurtisvg, line 104: Between this line and ready I feel like I'm confused. Is n the number of connections or the number of instances?

kurtisvg (changes requested):

internal/healthcheck/healthcheck.go
enocom: I'll remove this.

internal/healthcheck/healthcheck.go
```diff
@@ -90,8 +89,24 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
+		c.logger.Errorf("[Health Check] min-ready was greater than registered instances")
```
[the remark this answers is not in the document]
enocom: This was meant to be a validation of input in case a user provided a value of min-ready greater than the number of registered instances. But its value seems low and will be handled by `ready`, so I'll remove it.

internal/healthcheck/healthcheck.go
```diff
@@ -90,8 +89,24 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
 	}
 
-	err := c.proxy.CheckConnections(ctx)
-	if err != nil && !ready(err, c.minReady) {
+	var minReady int
+	q := req.URL.Query()
+	if v := q.Get("min-ready"); v != "" {
+		n, err := strconv.Atoi(v)
+		// If the query value parsed correctly, set it to minReady. Otherwise,
+		// ignore it.
```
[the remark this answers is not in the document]
enocom: Yes, good idea.

internal/healthcheck/healthcheck_test.go
enocom: Good point. I'll add one.

internal/healthcheck/healthcheck.go
enocom: That seems reasonable. I'll update this.

internal/healthcheck/healthcheck.go
enocom: Checking for 0 is necessary, because it means the min-ready query param was not set and so any error is a failure for the check. I'll add a comment.

internal/healthcheck/healthcheck.go
enocom: I've updated this slightly to be more readable.

internal/healthcheck/healthcheck.go
enocom: I've cleaned this up to make it more clear.

kurtisvg (changes requested):

internal/healthcheck/healthcheck.go
```diff
@@ -88,8 +89,20 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
 
-	err := c.proxy.CheckConnections(ctx)
-	if err != nil {
+	var minReady int
+	q := req.URL.Query()
+	if v := q.Get("min-ready"); v != "" {
+		n, err := strconv.Atoi(v)
+		if err != nil {
+			c.logger.Errorf("[Health Check] min-ready %q was not a valid integer", v)
+			w.WriteHeader(http.StatusServiceUnavailable)
```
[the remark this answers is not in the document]
enocom: Done and done.

internal/healthcheck/healthcheck.go
enocom: Added support for 0, which always returns a 200. Also, if a caller specifies more than the number of registered instances, returns a 503.

enocom: And after writing that, I've changed my mind. Min-ready must be greater than zero. Otherwise it's a 400.

internal/healthcheck/healthcheck_test.go
```diff
@@ -239,3 +257,85 @@ func TestHandleReadinessWithConnectionProblems(t *testing.T) {
...
+			dialer:     &errorDialer{},
+		},
+		{
+			desc:       "when min ready is greater than the number of instances",
+			minReady:   "3",
+			wantStatus: http.StatusBadRequest,
+			dialer:     &fakeDialer{},
+		},
+		{
+			desc:       "when min ready is bogus",
```
hessjcg, line 299: Thanks for adding bad data test cases

hessjcg (approved):
Looks good. The logic is correct. I provided a comment on the logic in ready() I'm curious for my own learning what you think.

internal/healthcheck/healthcheck.go
```diff
@@ -88,18 +89,64 @@ func (c *Check) HandleReadiness(w http.ResponseWriter, _ *http.Request) {
...
+		w.Write([]byte(rErr.Error()))
 		return
 	}
 
 	w.WriteHeader(http.StatusOK)
 	w.Write([]byte("ok"))
 }
 
+func ready(err error, minReady *int, total int) (int, error) {
+	// If err is nil, then the proxy is ready.
```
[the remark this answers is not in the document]
enocom: I agree getting this logic as plain and simple as possible is hard. Your revision here reads a little better given its consolidation of status codes as much as possible. I think we can push this a little further and separate the happy path from the error path. I'll push another commit.

kurtisvg (approved):
Looks much better.

PR #1500, opened 2022-10-21
feat: deprecate proxy drivers

enocom:


proxy/dialers/mysql/hook.go
```diff
@@ -37,6 +37,10 @@ func init() {
 //
 // The returned *sql.DB may be valid even if there's also an error returned
 // (e.g. if there was a transient connection error).
+//
+// Deprecated: Dial has been replaced by the Cloud SQL Go connector which has
+// better support for configuring the dialer's behavior.
+// See cloud.google.com/go/cloudsqlconn/mysql/mysql.RegisterDriver instead.
```
kurtisvg, line 43: This seems unlikely to be seen unless someone reads the code - should we consider adding a 1 time log warning or something?

kurtisvg (approved):

proxy/dialers/mysql/hook.go
enocom: It would be nice to put this front and center, but since we don't know how people are setting up their logging system, the best we could do is log to stdout which would be an intrusive change. I think we'll have to advertise this change in the new release.

PR #1520, opened 2022-10-31
fix: impersonated user uses downscoped token

### LINKED ISSUE #1519: Impersonated user should use IAM AuthN downscoped token
When impersonating a service account with Auto IAM AuthN, the proxy should configure the Go connector with an IAM Login token source, downscoped to login only.

enocom:
When an impersonated user logs in with Auto IAM AuthN, the login token is downscoped to support login only.

Fixes #1519.


enocom:
I need to add our impersonated user as an IAM user in the postgres database for these tests to pass.

kurtisvg (approved):

PR #1523, opened 2022-11-01
docs: update to latest preview

enocom:


kurtisvg (approved):
This should make it into the release guide so we don't forget it again.

PR #1542, opened 2022-11-11
fix: add runtime version to user agent if present

shubha-rajan:
Add proxy operator version to user agent string if `CSQL_PROXY_RUNTIME` environment variable is set

Fixes https://github.com/GoogleCloudPlatform/cloud-sql-proxy-operator/issues/67


hessjcg (approved):
LGTM

cmd/root.go
```diff
@@ -52,7 +52,16 @@ var (
...
 func init() {
 	versionString = strings.TrimSpace(versionString)
-	userAgent = "cloud-sql-proxy/" + versionString
+	userAgent = getUserAgentString()
+}
+
+func getUserAgentString() string {
+	userAgentString := "cloud-sql-proxy/" + versionString
+	operatorVersion, isSet := os.LookupEnv("CLOUD_SQL_PROXY_OPERATOR_VERSION")
+	if isSet {
```
hessjcg, line 61: if !isSet return?

cmd/root_test.go
```diff
@@ -81,6 +82,18 @@ func invokeProxyCommand(args []string) (*Command, error) {
 	return c, err
 }
 
+func Test_UserAgentWithOperatorVersion(t *testing.T) {
+	os.Setenv("CLOUD_SQL_PROXY_OPERATOR_VERSION", "0.0.1")
+    defer os.Unsetenv("CLOUD_SQL_PROXY_OPERATOR_VERSION")
```
enocom, line 87: This needs to be run through go-fmt.

cmd/root_test.go
```diff
@@ -81,6 +82,18 @@ func invokeProxyCommand(args []string) (*Command, error) {
...
+
+	expected := "cloud-sql-proxy-operator/0.0.1"
```
enocom, line 89: `want` is conventional here

cmd/root_test.go
```diff
@@ -81,6 +82,18 @@ func invokeProxyCommand(args []string) (*Command, error) {
...
+	userAgentString := getUserAgentString()
```
enocom, line 90: `got` is a common name for variables here.

enocom:
We should make this generic such that other runtimes can use it too.

See https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/acb57cb9ce803062651a4b4764e9181237d84a23/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L439-L454 where we already do this in the Java connector.

cmd/root.go
```diff
@@ -52,7 +52,20 @@ var (
...
+	userAgent = userAgentString()
+}
+
+func userAgentString() string {
+	ua := "cloud-sql-proxy/" + versionString
+	runtime, ok := os.LookupEnv("CLOUD_SQL_PROXY_RUNTIME")
+	if !ok {
+		return ua
+	}
+	rv, ok := os.LookupEnv("CLOUD_SQL_PROXY_RUNTIME_VERSION")
```
enocom, line 64: Shall we just use `CLOUD_SQL_PROXY_RUNTIME` and let the runtime include version?

cmd/root_test.go
```diff
@@ -81,6 +82,20 @@ func invokeProxyCommand(args []string) (*Command, error) {
 	return c, err
 }
 
+func Test_UserAgentWithOperatorVersion(t *testing.T) {
```
enocom, line 85: We've been omitting underscores in test names. Let's remove the underscore here.

cmd/root_test.go
```diff
@@ -81,6 +82,20 @@ func invokeProxyCommand(args []string) (*Command, error) {
...
+	os.Setenv("CLOUD_SQL_PROXY_RUNTIME", "cloud-sql-proxy-operator")
+	defer os.Unsetenv("CLOUD_SQL_PROXY_RUNTIME")
+	os.Setenv("CLOUD_SQL_PROXY_RUNTIME_VERSION", "0.0.1")
+	defer os.Unsetenv("CLOUD_SQL_PROXY_OPERATOR_VERSION")
+
+	want := "cloud-sql-proxy-operator/0.0.1"
+	got := userAgentString()
+	if !strings.Contains(got, want) {
```
enocom, line 93: It would be nice if we could make this assertion more exact.

enocom (approved):

cmd/root.go
```diff
@@ -52,7 +52,17 @@ var (
 
 func init() {
 	versionString = strings.TrimSpace(versionString)
-	userAgent = "cloud-sql-proxy/" + versionString
+	userAgent = userAgentString()
+}
+
+func userAgentString() string {
+	ua := "cloud-sql-proxy/" + versionString
+	runtime, ok := os.LookupEnv("CLOUD_SQL_PROXY_RUNTIME")
```
enocom, line 60: @kurtisvg pointed out that some runtimes (e.g., Cloud run where they'll use this programatically) might not be able to set an environment variable.

Let's make this a CLI flag proper (which will also support an environment variable).

I'm in favor of a single flag rather than a repeatable flag.

cmd/root.go
```diff
@@ -344,6 +345,8 @@ func NewCommand(opts ...Option) *Command {
 	pflags.BoolP("version", "v", false, "Print the cloud-sql-proxy version")
 
 	// Global-only flags
+	pflags.StringVar(&c.runtime, "runtime", "",
+		"Runtime and version, e.g. cloud-sql-proxy-operator/0.0.1")
```
enocom, line 349: Let's add something like "(for internal use only)".

enocom (approved):

PR #1623, opened 2023-01-25
feat: Add unix-socket-path to instance command arguments.

### LINKED ISSUE #1573: Add support for specifying the path to Unix socket per instance
## Bug Description
If you set the `unix-socket` parameter on a database instance, it should be the full path to the instance. In Proxy V2 preview 3, the instance name is appended to the value set for unix-socket. This is confusing.

## Expected Behavior
```
cloud-sql-proxy --unix-socket /var/csql myproject:us-central1:mydb
# listens on socket /var/csql/myproject:us-central1:mydb/.s.PGSQL.5432

cloud-sql-proxy myproject:us-central1:mydb?unix-socket=/var/mydb
# listens on socket /var/mydb/.s.PGSQL.5432
```

## Actual Behavior
```
# OK
cloud-sql-proxy --unix-socket /var/csql myproject:us-central1:mydb
# listens on socket /var/csql/myproject:us-central1:mydb/.s.PGSQL.5432

# BROKEN
cloud-sql-proxy myproject:us-central1:mydb?unix-socket=/var/mydb
# listens on socket /var/mydb/myproject:us-central1:mydb/.s.PGSQL.5432
```

## The Fix
The behavior of the existing `unix-socket` parameter defined globally or per-instance will remain unchanged. When unix-socket is set, it will  create a socket of the form "<unix-socket>/<instance name>" 

A new parameter `unix-socket-path`  will create paths of the form: "<unix-socket-path>". This new parameter may only be set per-instance. 

Postgres clients expect a socket address where the last element in the path is `.s.PGSQL.5432`. For Postgres database instances only, when unix-socket-path is set, the proxy will ensure that the last element of the path is `.s.PGSQL.5432`, adding  special path element if necessary. 

When both `unix-socket-path` and `unix-socket` are set, `unix-socket-path` will take precedence. 

If the socket path for one instance conflicts with the socket path for another instance, the proxy will print an error message and exit with an error code.

### Examples

| Database    |    Proxy CLI Arguments        |   Socket Path    | 
|---------|------------------------------------------------------------------------|--------------------------------------|
| mysql or mssql     |  --unix-socket /var/csql myproject:us-central1:mysqldb | /var/csql/myproject:us-central1:mydb |
| mysql or mssql    |  myproject:us-central1:mysqldb?unix-socket=/var/csql | /var/csql/myproject:us-central1:mydb |
| mysql or mssql    |  myproject:us-central1:mysqldb?unix-socket-path=/var/csql | /var/csql |
| mysql or mssql    |  --unix-socket /var/csql myproject:us-central1:mysqldb?unix-socket-path=/var/special | /var/special |
| pg          |  --unix-socket /var/csql myproject:us-central1:mysqldb | /var/csql/myproject:us-central1:mydb/.s.PGSQL.5432 |
| pg          |    myproject:us-central1:mysqldb?unix-socket=/var/csql  | /var/csql/myproject:us-central1:mydb/.s.PGSQL.5432 |
| pg           |   myproject:us-central1:mysqldb?unix-socket-path=/var/csql  | /var/csql/.s.PGSQL.5432 |
| pg           |   myproject:us-central1:mysqldb?unix-socket-path=/var/csql/.s.PGSQL.5432 | /var/csql/.s.PGSQL.5432 |
| pg    |  --unix-socket /var/csql myproject:us-central1:mysqldb?unix-socket-path=/var/special | /var/special/.s.PSQL.5432 |

hessjcg:
## Change Description

This adds a new parameter `unix-socket-path` to instances to specify the precise unix socket path. 

For example, this command line: 

```
$ cloud-sql-proxy my-project:region:mydb?unix-socket-path=/var/cloudsql/mydb 
```

Will open a unix socket to the database at the path `/var/cloudsql/mydb`.

This is different than the existing parameter `unix-socket` which automatically appends the instance name to the path. 

## Checklist

- [x] Open issue: #1573.

## Relevant issues:

- Fixes #1573

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -173,11 +173,11 @@ Instance Level Configuration
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
     When necessary, you may specify the full path to a Unix socket. Set the
-    unix-socket-path query parameter to the absolute path of the unix socket for
+    unix-socket-path query parameter to the absolute path of the Unix socket for
     the database instance. The parent directory of the unix-socket-path must 
     exist when the proxy starts or else socket creation will fail. For Postgres
     instances, the proxy will ensure that the last path element is 
-		'.s.PGSQL.5432' appending it if necessary. For example,
+    '.s.PGSQL.5432' appending it if necessary. For example,
 
         ./cloud-sql-proxy \
           'my-project:us-central1:my-db-server?unix-socket-path=/path/to/socket'
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -172,16 +172,15 @@ Instance Level Configuration
     	    my-project:us-central1:my-db-server \
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
-Instance Level Configuration Parameters:
+    When necessary, you may specify the full path to a Unix socket. Set the
+    unix-socket-path query parameter to the absolute path of the unix socket for
+    the database instance. The parent directory of the unix-socket-path must 
+    exist when the proxy starts or else socket creation will fail. For Postgres
+    instances, the proxy will ensure that the last path element is 
+		'.s.PGSQL.5432' appending it if necessary. For example,
 
-    There is a parameter that may only be set at the instance, and does not have
-    a corresponding global parameter.
-
-    - unix-socket-path - UnixSocketPath is the path where a Unix socket will be
-    created for a Cloud SQL instance. If set on a Postgres instance, the proxy
-    will ensure that the last path element is '.s.PGSQL.5432', appending this
-		path element if necessary. unix-socket-path takes precedence over
-    the global parameter --unix-socket when both are set.
+        ./cloud-sql-proxy \
+          'my-project:us-central1:my-db-server?unix-socket-path=/path/to/socket'
 
 Health checks
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -178,7 +178,7 @@ Instance Level Configuration Parameters:
     a corresponding global parameter.
 
     - unix-socket-path - UnixSocketPath is the path where a Unix socket will be
-    created for a Cloud SQL instance. If this is a Postgres database, the proxy
+    created for a Cloud SQL instance. If set on a Postgres instance, the proxy
     will ensure that the last path element is '.s.PGSQL.5432', appending this
 		path element if necessary. unix-socket-path takes precedence over
     the global parameter --unix-socket when both are set.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -543,12 +543,22 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 			a, aok := q["address"]
 			p, pok := q["port"]
 			u, uok := q["unix-socket"]
+			up, upok := q["unix-socket-path"]
 
 			if aok && uok {
 				return newBadCommandError("cannot specify both address and unix-socket query params")
 			}
 			if pok && uok {
 				return newBadCommandError("cannot specify both port and unix-socket query params")
+			}
+			if aok && upok {
+				return newBadCommandError("cannot specify both address and unix-socket-path query params")
+			}
+			if pok && upok {
+				return newBadCommandError("cannot specify both port and unix-socket-path query params")
+			}
+			if uok && upok {
+				return newBadCommandError("cannot specify both unix-socket-path and unix-socket query params")
 			}
 
 			if aok {
@@ -583,6 +593,13 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 					return newBadCommandError(fmt.Sprintf("unix query param should be only one value: %q", a))
 				}
 				ic.UnixSocket = u[0]
+			}
+
+			if upok {
+				if len(up) != 1 {
+					return newBadCommandError(fmt.Sprintf("unix-socket-path query param should be only one value: %q", a))
+				}
+				ic.UnixSocketPath = up[0]
 			}
 
 			ic.IAMAuthN, err = parseBoolOpt(q, "auto-iam-authn")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -172,6 +172,17 @@ Instance Level Configuration
     	    my-project:us-central1:my-db-server \
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
+Instance Level Configuration Parameters:
+
+    There is a parameter that may only be set at the instance, and does not have
+    a corresponding global parameter.
+
+    - unix-socket-path - UnixSocketPath is the path where a Unix socket will be
+    created for a Cloud SQL instance. If this is a Postgres database, the proxy
+    will ensure that the last path element is '.s.PGSQL.5432', appending this
+		path element if necessary. unix-socket-path takes precedence over
+    the global parameter --unix-socket when both are set.
+
 Health checks
 
     When enabling the --health-check flag, the proxy will start an HTTP server
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -288,10 +288,10 @@ func TestNewCommandArguments(t *testing.T) {
 		},
 		{
 			desc: "using the unix socket path query param",
-			args: []string{"proj:region:inst?unix-socket-path=/path/to/dir/"},
+			args: []string{"proj:region:inst?unix-socket-path=/path/to/file"},
 			want: withDefaults(&proxy.Config{
 				Instances: []proxy.InstanceConnConfig{{
-					UnixSocketPath: "/path/to/dir/",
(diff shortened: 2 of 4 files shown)


cmd/root.go
```diff
@@ -543,13 +543,23 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
...
 			u, uok := q["unix-socket"]
+			up, upok := q["unix-socket-path"]
 
 			if aok && uok {
 				return newBadCommandError("cannot specify both address and unix-socket query params")
 			}
 			if pok && uok {
 				return newBadCommandError("cannot specify both port and unix-socket query params")
 			}
+			if aok && upok {
```
enocom, line 554: Let's add unit tests for these error cases.

cmd/root_test.go
```diff
@@ -286,6 +286,15 @@ func TestNewCommandArguments(t *testing.T) {
 				}},
 			}),
 		},
+		{
+			desc: "using the unix socket path query param",
+			args: []string{"proj:region:inst?unix-socket-path=/path/to/dir/"},
```
enocom, line 291: nit: remove the trailing slash to make it clear we're not passing a directory.

internal/proxy/proxy.go
```diff
@@ -83,9 +84,13 @@ type InstanceConnConfig struct {
 	// Port is the port on which to bind a listener for the instance.
 	Port int
 	// UnixSocket is the directory where a Unix socket will be created,
-	// connected to the Cloud SQL instance. If set, takes precedence over Addr
-	// and Port.
+	// connected to the Cloud SQL instance. The full path to the socket will be
+	// UnixSocket + "/" + Name. If set, takes precedence over Addr and Port.
```
enocom, line 88: On windows, the forward slash isn't in the name. Let's either remove the sentence or update it to be accurate, e.g., The socket will be created in the UnixSocket directory with Name. When connecting to Postgres, Name will be a directory, etc etc.

internal/proxy/proxy.go
```diff
@@ -83,9 +84,13 @@ type InstanceConnConfig struct {
...
 	UnixSocket string
+	// UnixSocketPath is the directory where a Unix socket will be created,
+	// connected to the Cloud SQL instance. The full path to the socket will be
+	// UnixSocketPath. If set, takes precedence over UnixSocket, Addr and Port.
```
enocom, line 92: Let's add the postgres caveat too.

internal/proxy/proxy.go
```diff
@@ -734,7 +741,47 @@ func newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst Inst
 		}
 
 		address = net.JoinHostPort(a, fmt.Sprint(np))
-	} else {
+
+	case inst.UnixSocketPath != "" && strings.HasPrefix(version, "POSTGRES"):
```
enocom, line 745: This case statement and the one below both duplicate code in the default path. How can we avoid that while keeping the intention clear?

internal/proxy/proxy_test.go
```diff
@@ -257,6 +261,39 @@ func TestClientInitialization(t *testing.T) {
 				filepath.Join(testDir, pg, ".s.PGSQL.5432"),
 			},
 		},
+		{
+			desc: "with a Unix socket path per instance",
```
enocom, line 265: Shall we add a unit test that shows how UnixSocketPath takes precedence over UnixSocket?

internal/proxy/proxy.go
enocom: As one option, leaving the if/else in place makes it clear that we're doing either TCP or Unix. Then we could introduce a function e.g., `address, err := unixSocketPath(dir, version, address, inst.Name)` that give you more room to gracefully handle all the variation.

enocom (changes requested):

cmd/root.go
hessjcg: Done.

cmd/root_test.go
hessjcg: Done.

internal/proxy/proxy.go
hessjcg: Done.

internal/proxy/proxy_test.go
hessjcg: Done.

internal/proxy/proxy.go
hessjcg: Good idea. Done.

cmd/root.go
```diff
@@ -172,6 +172,17 @@ Instance Level Configuration
     	    my-project:us-central1:my-db-server \
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
+Instance Level Configuration Parameters:
+
+    There is a parameter that may only be set at the instance, and does not have
+    a corresponding global parameter.
+
+    - unix-socket-path - UnixSocketPath is the path where a Unix socket will be
```
enocom, line 180: Let's put this paragraph under "Instance Level Configuration" and not introduce another section.

internal/proxy/proxy.go
```diff
@@ -775,6 +767,58 @@ func newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst Inst
...
+		// the parent directory of the unix socket
+		dir string
+	)
+
+	if inst.UnixSocketPath != "" {
+		// When UnixSocketPath is set
+		address = inst.UnixSocketPath
+
+		// If UnixSocketPath ends .s.PGSQL.5432, remove it for consistency
+		if strings.HasPrefix(version, "POSTGRES") && path.Base(address) == ".s.PGSQL.5432" {
```
enocom, line 786: What happens if "address" doesn't exist?

internal/proxy/proxy.go
```diff
@@ -775,6 +767,58 @@ func newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst Inst
...
+
+	// if base directory does not exist, fail
+	if _, err := os.Stat(dir); err != nil {
+		return "", err
+	}
+
+	// When setting up a listener for Postgres, create address as a
+	// directory, and use the Postgres-specific socket name
+	// .s.PGSQL.5432.
+	if strings.HasPrefix(version, "POSTGRES") {
```
enocom, line 808: Should we pull this into a function and re-use it above?

cmd/root.go
```diff
@@ -172,6 +172,17 @@ Instance Level Configuration
     	    my-project:us-central1:my-db-server \
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
+Instance Level Configuration Parameters:
```
enocom, line 175: Let's leave this section off and immediately go into the particular parameter here. For example,

When necessary, you may specify the full path to a Unix socket using the `unix-socket-path` query parameter. The value of the parameter should be an absolute path to where you want the socket to be created. Parent directories must exist, or the socket creation will fail. For Postgres instances, the proxy will ensure that the last path element is '.s.PGSQL.5432', appending it if necessary. For example,

./cloud-sql-proxy \
    	    'my-project:us-central1:my-db-server?unix-socket-path=/path/to/socket'

enocom (changes requested):
Minor rewrite of the docs.

internal/proxy/proxy.go
hessjcg: We return an error on line 803 if the parent directory for `address` doesn't exist.

internal/proxy/proxy.go
hessjcg: I'm not sure it needs to be a function. I've refactored it into a parameter instead so that the check `strings.HasPrefix(version, "POSTGRES")` only appears once.

cmd/root.go
hessjcg: Updated with a few minor edits.

cmd/root.go
```diff
@@ -172,6 +172,16 @@ Instance Level Configuration
     	    my-project:us-central1:my-db-server \
     	    'my-project:us-central1:my-other-server?address=0.0.0.0&port=7000'
 
+    When necessary, you may specify the full path to a Unix socket. Set the
+    unix-socket-path query parameter to the absolute path of the unix socket for
```
enocom, line 176: s/unix/Unix/

cmd/root.go
```diff
@@ -172,6 +172,16 @@ Instance Level Configuration
...
+    the database instance. The parent directory of the unix-socket-path must 
+    exist when the proxy starts or else socket creation will fail. For Postgres
+    instances, the proxy will ensure that the last path element is 
+		'.s.PGSQL.5432' appending it if necessary. For example,
```
enocom, line 180: let's get rid of the whitespace here.

enocom (approved):

cmd/root.go
hessjcg: fixed.

cmd/root.go
hessjcg: fixed.

PR #1624, opened 2023-01-26
feat: add support for quitquitquit endpoint

### LINKED ISSUE #828: Provide a way to shutdown the process from another container
## Feature Description
It is a common practice to run database migrations in an init container and wait for the migration job to finish. However, when the proxy never shuts down, init script stalls. Would be good to provide a way to shut down the proxy gracefully. For example an endpoint for other containers to call when finished.

## Alternatives Considered
Not sure really. If Kubernetes had a native way to shutdown the whole pod and deliver the signal to the container, that could work too. But I think an endpoint could be easier and more straightforward.

## Additional Context
An example tutorial on how to run migrations with init containers, kubernetes jobs and k8s-wait-for: https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-8-running-database-migrations-using-jobs-and-init-containers/

enocom:
Fixes #828.


cmd/root.go
```diff
@@ -254,16 +254,20 @@ Configuration using environment variables
...
+    To change the port, use the --admin-port flag.
 
-    The admin server includes Go's pprof tool and is available at
+    When --debug is passed, the admin server enables Go's profiler available at
     /debug/pprof/.
 
     See the documentation on pprof for details on how to use the
     profiler at https://pkg.go.dev/net/http/pprof.
 
+    When --quitquitquit is passed, the admin server adds an endpoint for
```
hessjcg, line 267: "...the admin server exits gracefully when it receives a POST request at /quitquitquit. When..."

cmd/root.go
```diff
@@ -799,3 +799,38 @@ func runSignalWrapper(cmd *Command) error {
...
 	return err
 }
+
+func quitquitquit(shutdownCh chan<- error) http.HandlerFunc {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		if req.Method != http.MethodPost {
+			rw.WriteHeader(400)
+			return
+		}
+		shutdownCh <- errQuitQuitQuit
```
hessjcg, line 809: What if /quitquitquit was called twice in rapid succession? Would this still exit cleanly?

hessjcg (approved):
Looks good. I suggested alternate wording for the --quitquitquit flag's help.

cmd/root.go
enocom: That's better.

cmd/root.go
enocom: Good point. I'll wrap this in a sync.Once.

hessjcg (approved):
LGTM

PR #1637, opened 2023-02-06
fix: ensure separate token source with auto-iam-authn

enocom:
When a caller passes the --token flag with --auto-iam-authn, the Proxy will now require a separate --login-token. This changes ensures a token with SQL Admin scope does not make its way into the ephemeral certificate. Instead, the value passed to --login-token will be used and should have only the sqlservice.login scope.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -520,8 +520,6 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
 		return newBadCommandError("cannot specify --login-token without --token and --auto-iam-authn")
 	}
 
-	// if login token is set, but token is not error
-
 	if userHasSet("http-port") && !userHasSet("prometheus") && !userHasSet("health-check") {
 		cmd.logger.Infof("Ignoring --http-port because --prometheus or --health-check was not set")
 	}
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -509,6 +511,17 @@ func parseConfig(cmd *Command, conf *proxy.Config, args []string) error {
...
+	// When using token with auto-iam-authn, login-token must also be set.
+	// All three are required together.
+	if conf.IAMAuthN && conf.Token != "" && conf.LoginToken == "" {
+		return newBadCommandError("cannot specify --auto-iam-authn and --token without --login-token")
+	}
+	if conf.LoginToken != "" && (conf.Token == "" || !conf.IAMAuthN) {
+		return newBadCommandError("cannot specify --login-token without --token and --auto-iam-authn")
+	}
+
+	// if login token is set, but token is not error
```
jackwotherspoon, line 523: is something missing here?

jackwotherspoon (changes requested):

cmd/root.go
enocom: Yes, I forgot to delete this comment. The case should be covered above.

jackwotherspoon (approved):

PR #1735, opened 2023-04-06
feat: add support for Auto IP

### LINKED ISSUE #1732: Add support for "Auto IP"
### Feature Description

https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/options.go#L254

This ensures customers who depended on the v1 behavior can still benefit from it in v2.

### Sample code

```
// sample code here
```

### Alternatives Considered

_No response_

### Additional Details

_No response_

enocom:
This commit restores a legacy behavior of the v1 Proxy, this time behind the flag --auto-ip. When the flag is enabled, the Proxy will first attempt to use a public IP. If no public IP is present, the Proxy will second attempt to use a private IP. If neither is present, the Proxy errors as expected.

This new flag is primarily to support legacy behavior and should not be used in production otherwise.

Fixes #1732.


enocom:
We need to add this option to the FUSE path as well. Currently FUSE assumes public IP always.

jackwotherspoon (approved):

PR #1759, opened 2023-04-19
doc: Improve examples for k8s readiness probes.

### LINKED ISSUE #1757: cloudsql proxy examples for k8s-sidecar contain a race condition
### Description

The examples provided assume that `cloud-sql-proxy` will always startup before your application container does. This might not always be true since your application could startup before `cloud-sql-proxy` does in which case you'd be unable to connect to your Cloud SQL instance.

### Potential Solution

I suggest adding an init container to the documentation that checks that cloud sql proxy is running before executing your application.

### Additional Details

_No response_

hessjcg:
This reorganizes and adds explanation to the startup and readiness probes. 

Also, this updates our recommendation to only use a startup and liveness probe, but not a readiness probe under most circumstances. 

Fixes #1757

### DIFF
--- a/examples/k8s-health-check/proxy_with_http_health_check.yaml
+++ b/examples/k8s-health-check/proxy_with_http_health_check.yaml
@@ -142,18 +142,9 @@ spec:
           # The /readiness probe returns OK when the proxy can establish
           # a new connections to its databases.
           #
-          # We do not recommend adding a readiness probe to the proxy sidecar
-          # container because it may cause unnecessary interruption to the
-          # application.
-          #
-          # This probe will fail when (1) the proxy used all its available
-          # concurrent connections to a database or (2) the network connection
-          # to the database is interrupted. These are transient states
-          # that resolve within a few seconds. However, if the readiness check
-          # fails too many times, k8s will kill the pod application container
-          # and the proxy container to be shut down. This is an undesirable
-          # result if the application is using a database connection
-          # pool with built-in resilience to transient failed connections.
+          # Please use the readiness probe to the proxy sidecar with caution.
+          # An improperly configured readiness probe can cause unnecessary
+          # interruption to the application. See README.md for more detail.
           httpGet:
             path: /readiness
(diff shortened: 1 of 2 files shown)


examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -120,23 +139,31 @@ spec:
           # restarted.
           failureThreshold: 5
         readinessProbe:
+          # The /readiness probe returns OK when the proxy can establish
+          # a new connections to its databases.
+          #
+          # We do not recommend adding a readiness probe to the proxy sidecar
```
enocom, line 145: If we don't recommend adding a readiness probe, should we just remove its current implementation and instead just check that the internal state is valid or something similar?

hessjcg: I think it's OK to leave it in. A minority of customers will still want to use it: probably customers installing the proxy alongside existing legacy applications that don't have health checks.

hessjcg: Also, the readiness check is helpful for our own e2e tests - because it ensures that the proxy can connect to the database server.

enocom: Let's lighten the recommendation to something like, "Note: a readiness probe should be used carefully to ensure no unnecessary interruption to the application." And then give examples of how max connections will affect the readiness probe. WDYT?

hessjcg: Ok, good idea. I'll put most of that nuance in README.md instead of a comment in this yaml file.

enocom: That seems fine. I just don't want to be recommending against a feature we built. If we're recommending against it, we should be fixing that.

enocom (approved):
Looks good.

PR #1761, opened 2023-04-20
doc: Example now has a better security posture.

hessjcg:
This adds additional controls on the security context for pod containers to address these common
k8s security best practices:

Run as a non-root user
Mount container's root filesystem as read only
Restrict Container from acquiring additional privileges


jackwotherspoon:
FYI: the proper conventional commit is `docs:` and not `doc:` (not that it really matters) 😄

enocom (approved):

PR #1763, opened 2023-04-20
chore: add testing.Short() to health check tests

enocom:
Also, use non-reserved ports in test to ensure the build runs on dev machines without issue (port 5000 is reserved on macOS).


jackwotherspoon (approved):
LGTM, I had been meaning to get around to this myself 😄

PR #1791, opened 2023-05-14
chore: rename bullseye docker image to cloud-sql-proxy

ngergs:
This has been omitted during the gce-proxy -> cloud-sql-proxy version rename for v2. Adjusted the build config so that future releases are released under the documented docker image name already used for the other tags.

Already published versions would need to be adjusted manually by someone with the respective permissions if desired.


jackwotherspoon:
@ngergs Great catch! Thanks for this PR updating the bullseye build to the proper naming convention, much appreciated!

I will go back and look at re-tagging previous builds to the proper tags 😄

jackwotherspoon (approved):

PR #1805, opened 2023-05-22
feat: allow connections during shutdown

### LINKED ISSUE #1640: Support forcing the proxy to accept new connections while max-sigterm-delay is in effect
## Problem Case
Consider the following scenario when running the proxy as a sidecar in k8s:

1. `term_timeout` or `max-sigterm-delay` is set for the proxy container to 5 minutes
2. Pod scales down
3. The app container AND the proxy get a sigterm (cause they are in the same pod)
4. App's cleanup routine requires it to make _new_ connections to the DB after getting a sigterm
5. The proxy will refuse _new_ connections after getting a sigterm and will shutdown early if there are no open connections (which for this app there occasionally will be)
6. This race condition shuts down the proxy before the app can access the DB and causes the app to fail to finish its sigterm cleanup routine

It doesn't matter what you set the term_timeout to because the app needs to initiate new connections after both it and the proxy get a sigterm

## Alternatives Considered
Hacky scripty entrypoint command that swallows sigterm and forces the proxy to remain up for a static amount of time to like the app container call the new quitquitquit endpoint

## Proposed solution
What I suggest is this: The new quitquitquit supports an idle_timeout setting in the request. So the process could work like this:

1. `term_timeout` or `max-sigterm-delay` is set for the proxy process to 5 minutes
2. Pod scales down
3. The app container AND the proxy get a sigterm (cause they are in the same pod)
5. App's cleanup routine immediately hits the `quitquitquit` endpoint of the proxy and sets `idle_timeout` to 30 seconds
6. `term_timeout` is still in effect but the proxy will allow _new_ connections for 30 seconds
7. The app is now able to make a new DB connection to finish its cleanup routine
8. After 30 secs there are no new connections and the proxy shuts down early
9. Yay

Would it be easy to implement an arbitrary user set timer in the `quitquitquit` method?

enocom:
When the proxy starts its shut down, it will now accept additional connections when the max-sigterm-delay flag is passed. As before, if any connections are still open after the delay has elapsed, the Proxy will report an error.

Fixes #1640.

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -612,9 +612,10 @@ func (m MultiErr) Error() string {
 // Close triggers the proxyClient to shut down.
 func (c *Client) Close() error {
 	mnts := c.mnts
-
 	var mErr MultiErr
 
+	// If FUSE is enabled, unmount it and save a reference to any existing
+	// socket mounts.
 	if c.fuseDir != "" {
 		if err := c.unmountFUSE(); err != nil {
 			mErr = append(mErr, err)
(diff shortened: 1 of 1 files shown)


albaraakefaya:
This fix is not working on later versions, I tried 2.11.3 and I can confirm that is not working in it
Can you check please?

enocom:
#1640 has been re-opened. Would you mind chiming in there with your use case @albaraakefaya?

PR #1813, opened 2023-05-25
docs: update Ruby version, tutorial notes for multi-containers

glasnt:
Changes: 

* Update Ruby version from 2.7.4 (EOL) to 3.2.2 (latest)
* Add explicit rack server (probably due to major Ruby version update)

Doc updates:
* Removed referenced to "2nd gen" Cloud SQL (not mentioned in docs or UI)
* Added optional note to create a user, or use default "postgres"
* Added API enabling (Cloud Build and Cloud Run prompted in CLI, Cloud SQL Admin isn't, so added all three)
* Added optional note around using Secret Manager rather than hardcoding password as an environment variable, linking to docs
* Added note around how the default service account will have permission to connect to Cloud SQL, but explicit permission required for custom service accounts
* Fixed filename on `gcloud run` command

Other improvements that can be made (but require more work):

* Container Registry is deprecated, need to update to use Artifact Registry (https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr)
* Could use Cloud Buildpacks for Ruby instead of Dockerfile, but native gem `pg` building had issues, would need further investigation. 

cc @shubha-rajan for review

Happy to also do the same changes to GoogleCloudPlatform/alloydb-auth-proxy once this is merged.


examples/multi-container/ruby/README.md
```diff
@@ -7,14 +7,24 @@ existing application which connects to a Cloud SQL instance.
...
 1. If you haven't already, [create a project](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project).
 
-1. Create a 2nd Gen Cloud SQL Postgres Instance by following these
+1. [Enable the APIs](https://console.cloud.google.com/flows/enableapi?apiid=run.googleapis.com,sqladmin.googleapis.com,run.googleapis.com) that will be used during this tutorial:
+
+  * Cloud SQL Admin
+  * Cloud Build
+  * Cloud Run
+  
+3. Create a Cloud SQL Postgres Instance by following these
```
jackwotherspoon, line 16: nit
```suggestion
1. Create a Cloud SQL Postgres Instance by following these
```

examples/multi-container/ruby/Gemfile
```diff
@@ -18,3 +18,4 @@ gem 'json'
 gem 'pg'
 gem 'sequel'
 gem 'sinatra'
+gem 'thin'
```
jackwotherspoon, line 21: nit newline at end of file
```suggestion
gem 'thin'

```

jackwotherspoon:
Thanks @glasnt! These look like great improvements!

jackwotherspoon (approved):

jackwotherspoon:
Thanks for this @glasnt ! If you would be so kind to port these same changes over to the [alloydb-auth-proxy](https://github.com/GoogleCloudPlatform/alloydb-auth-proxy/tree/main/examples/multi-container/ruby) too that would be greatly appreciated 😃

PR #1833, opened 2023-06-06
fix: Windows service stop

nick-oconnor:
When the Windows service stops, it immediately spawns a command line instance with the same arguments. This rebinds to the same port(s) and effectively prevents the service from stopping or restarting.


nick-oconnor:
CC: @enocom

enocom (approved):
Thanks!

PR #1856, opened 2023-06-24
chore: use `t.TempDir` to create temporary test directory

Juneezee:
This pull request replaces `os.MkdirTemp` with `t.TempDir` in tests. We can use the `t.TempDir` function from the `testing` package to create temporary directory. The directory created by `t.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	dir, err := os.MkdirTemp("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(dir)

	// now
	dir := t.TempDir()
}
```

### DIFF
--- a/internal/proxy/fuse_test.go
+++ b/internal/proxy/fuse_test.go
@@ -30,16 +30,6 @@ import (
 	"github.com/hanwen/go-fuse/v2/fs"
 )
 
-func randTmpDir(t interface {
-	Fatalf(format string, args ...interface{})
-}) string {
-	name, err := os.MkdirTemp("", "*")
-	if err != nil {
-		t.Fatalf("failed to create tmp dir: %v", err)
-	}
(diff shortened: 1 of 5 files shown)


hessjcg (approved):
Thank you for the contribution. This Looks good to me. Nice simplification of the test cases.

enocom (changes requested):
All the integration tests for unix sockets are failing because the temp dir is too long. 

Looks like t.TempDir always prepends the test name to the directory name which probably makes it infeasible for use here. See https://docs.studygolang.com/src/testing/testing.go?s=31998:32031#L962.

I'm not sure we can accept this, but if you have other ideas I'd be curious to hear them.

PR #1870, opened 2023-07-11
feat: add support for exit zero on sigterm

### LINKED ISSUE #1803: SIGTERM with no active connections should result in 0 exit code
### Feature Description

We have an issue where airflow detects non-zero exit code as a problem in KubernetesPodOperator. Digging deeper to the recent changes in V2 it seems like we moved from 0 to 143 in the case of sigterm.

However this behaviour is curious when all connections etc are gracefully terminated within the time period. There is really no issue so shutdown should result in 0 (imho).

Whats the reasoning behind having `143` always?

### Sample code

_No response_

### Alternatives Considered

I was reading https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/1529#issuecomment-1308118348 and would like to avoid `/quitquitquit`. And I also cant really figure out why using the endpoin is 'better' than SIGTERM?

### Additional Details

_No response_

htigran:
This PR is to add a new argument `exit-zero-on-sigterm` to exit with 0 exit code when `SIGTERM` is received.

This is needed because since v2 cloud-sql-proxy started returning 143 code which is causing problems with Airflow KPO and other applications on k8s.
When cloud-sql-proxy runs in the same pod with other containers and when the main container exits successfully k8s sends `SIGTERM` to cloud-sql-proxy. Then cloud-sql-proxy is returning 143 in this scenario, and airflow KPO thinks that there was an error and marks the job as failed.

Fixes #1803.


enocom:
If you'll sign the CLA, we can get this merged.

htigran:
Hi @enocom, just to let you know that I have signed the CLA. Thanks

enocom (approved):
Thanks for this!

PR #1911, opened 2023-08-15
build: add more tags to builds

jackwotherspoon:
Adding additional tags to docker images.

Currently we only tag builds with a single tag such as `2.6.1-alpine`

This PR will result in additional tags such as `alpine`, `2.6-alpine`, and `2-alpine`.

The benefit is users who want to always pull the latest alpine build of the Proxy can now just pull the `alpine` tag etc.

### DIFF
--- a/.build/release_artifacts.sh
+++ b/.build/release_artifacts.sh
@@ -58,11 +58,27 @@ done
 
 tag_latest() {
     local new_version=$1
+    local minor_version=${new_version%.*}
+    local major_version=${new_version%.*.*}
     for registry in "gcr.io" "us.gcr.io" "eu.gcr.io" "asia.gcr.io"
     do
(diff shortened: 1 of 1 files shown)


.build/release_artifacts.sh
```diff
@@ -58,11 +58,27 @@ done
 
 tag_latest() {
     local new_version=$1
+    local minor_version=${new_version%.*}
```
enocom, line 63: nit: let's make this variable `major_minor_version`

enocom: Or `version_without_patch`.

enocom: Also a link to SO or a bash docs page would help if we ever want to change this.

enocom (approved):

PR #1937, opened 2023-09-01
chore: improve logging with additional test coverage

annafang-google:
Let proxy continue to run even instances specified are inactive.
Tested these scenarios:
- inactive instances without TCP or Unix socket options
- inactive instances with TCP port
- inactive instances with Unix socket

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -510,29 +510,19 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			switch err.(type) {
-			// this is not the inactive instance case. Fail the proxy
-			// so that the user could fix the proxy command error.
-			// for example: two instances with the same port will fail
-			// in the listen call.
-			case *net.OpError:
-				for _, m := range mnts {
-					mErr := m.Close()
-					if mErr != nil {
-						l.Errorf("[%v] failed to close mount: %v", m.inst, mErr)
-					}
+			for _, m := range mnts {
+				mErr := m.Close()
+				if mErr != nil {
+					l.Errorf("failed to close mount: %v", mErr)
 				}
-				return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
 			}
-		} else {
-			l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
-			mnts = append(mnts, m)
+			return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
 		}
+
+		l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
+		mnts = append(mnts, m)
 	}
 	c.mnts = mnts
-	if len(mnts) == 0 {
-		return nil, fmt.Errorf("No active instance to mount")
-	}
 	return c, nil
 }
 
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -513,6 +513,8 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 			switch err.(type) {
 			// this is not the inactive instance case. Fail the proxy
 			// so that the user could fix the proxy command error.
+			// for example: two instances with the same port will fail
+			// in the listen call.
 			case *net.OpError:
 				for _, m := range mnts {
 					mErr := m.Close()
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -510,20 +510,22 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			c.logger.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
-			if networkType(conf, inst) == "unix" {
-				continue
-			}
 			switch err.(type) {
 			// this is not the inactive instance case. Fail the proxy
 			// so that the user could fix the proxy command error.
 			case *net.OpError:
+				for _, m := range mnts {
+					mErr := m.Close()
+					if mErr != nil {
+						l.Errorf("[%v] failed to close mount: %v", m.inst, mErr)
+					}
+				}
 				return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
 			}
+		} else {
+			l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
+			mnts = append(mnts, m)
 		}
-
-		l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
-		mnts = append(mnts, m)
 	}
 	c.mnts = mnts
 	if len(mnts) == 0 {
@@ -795,8 +797,10 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 			np = pc.nextPort()
 		default:
 			version, err := c.dialer.EngineVersion(ctx, inst.Name)
+			// Exit if the port is not specified for inactive instance
 			if err != nil {
-				c.logger.Errorf("could not resolve version for %q: %v", inst.Name, err)
+				c.logger.Errorf("[%v] could not resolve instance version: %v", inst.Name, err)
+				return nil, err
 			}
 			np = pc.nextDBPort(version)
 		}
@@ -807,14 +811,13 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 
 		version, err := c.dialer.EngineVersion(ctx, inst.Name)
 		if err != nil {
-			c.logger.Errorf("could not resolve version for %q: %v", inst.Name, err)
+			c.logger.Errorf("[%v] could not resolve instance version: %v", inst.Name, err)
 			return nil, err
 		}
 
 		address, err = newUnixSocketMount(inst, conf.UnixSocket, strings.HasPrefix(version, "POSTGRES"))
 		if err != nil {
-			c.logger.Errorf("could not mount unix socket %q for %q: %v",
-				conf.UnixSocket, inst.Name, err)
+			c.logger.Errorf("[%v] could not mount unix socket %q: %v", inst.Name, conf.UnixSocket, err)
 			return nil, err
 		}
 	}
@@ -822,6 +825,7 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 	lc := net.ListenConfig{KeepAlive: 30 * time.Second}
 	ln, err := lc.Listen(ctx, network, address)
 	if err != nil {
+		c.logger.Errorf("[%v] could not listen to address %v: %v", inst.Name, address, err)
 		return nil, err
 	}
 	// Change file permissions to allow access for user, group, and other.
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -510,19 +510,25 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			for _, m := range mnts {
-				mErr := m.Close()
-				if mErr != nil {
-					l.Errorf("failed to close mount: %v", mErr)
-				}
+			c.logger.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
+			if networkType(conf, inst) == "unix" {
+				continue
 			}
-			return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
+			switch err.(type) {
+			// this is not the inactive instance case. Fail the proxy
+			// so that the user could fix the proxy command error.
+			case *net.OpError:
+				return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
+			}
 		}
 
 		l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
 		mnts = append(mnts, m)
 	}
 	c.mnts = mnts
+	if len(mnts) == 0 {
+		return nil, fmt.Errorf("No active instance to mount")
+	}
 	return c, nil
 }
 
@@ -747,6 +753,14 @@ type socketMount struct {
 	dialOpts []cloudsqlconn.DialOption
 }
 
+func networkType(conf *Config, inst InstanceConnConfig) string {
+	if (conf.UnixSocket == "" && inst.UnixSocket == "" && inst.UnixSocketPath == "") ||
+		(inst.Addr != "" || inst.Port != 0) {
+		return "tcp"
+	}
+	return "unix"
+}
+
 func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfig, inst InstanceConnConfig) (*socketMount, error) {
 	var (
 		// network is one of "tcp" or "unix"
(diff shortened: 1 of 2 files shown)


internal/proxy/proxy.go
```diff
@@ -510,13 +510,7 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			for _, m := range mnts {
-				mErr := m.Close()
-				if mErr != nil {
-					l.Errorf("failed to close mount: %v", mErr)
-				}
-			}
-			return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
+			fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
```
enocom, line 513: We need to either log this error, or return it. Right now, this is creating an error value and discarding it.

enocom: If there is only one instance, should we still fail the proxy?

internal/proxy/proxy.go
```diff
@@ -784,7 +778,6 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 			version, err := c.dialer.EngineVersion(ctx, inst.Name)
 			if err != nil {
 				c.logger.Errorf("could not resolve version for %q: %v", inst.Name, err)
-				return nil, err
```
enocom, line 787: I'm surprised we don't have test failures for removing this line. In any case, we should always be returning the error and only decide if we want to fail the proxy up above where we set up all these socket mounts.

internal/proxy/proxy.go
```diff
@@ -796,12 +789,12 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 		version, err := c.dialer.EngineVersion(ctx, inst.Name)
 		if err != nil {
 			c.logger.Errorf("could not resolve version for %q: %v", inst.Name, err)
-			return nil, err
```
enocom, line 799: Ditto here.

internal/proxy/proxy.go
```diff
@@ -796,12 +789,12 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
...
 		}
 
 		address, err = newUnixSocketMount(inst, conf.UnixSocket, strings.HasPrefix(version, "POSTGRES"))
 		if err != nil {
-			return nil, err
```
enocom, line 804: And ditto here.

internal/proxy/proxy_test.go
```diff
@@ -306,6 +309,39 @@ func TestClientInitialization(t *testing.T) {
 				filepath.Join(testUnixSocketPathPg),
 			},
 		},
+		{
+			desc: "without TCP port or unix socket for non functional instance",
+			in: &proxy.Config{
```
enocom, line 314: The test below is build around wantTCPAddrs or wantUnixAddrs. So as written, this test will never fail. To ensure we have the behavior we want here, we should add a functioning socket with a non functioning one and ensure we get what we expect with wantTCPAddrs or wantUnixAddrs.

internal/proxy/proxy_test.go
```diff
@@ -306,6 +309,39 @@ func TestClientInitialization(t *testing.T) {
...
+			desc: "with unix socket for non functional instance",
+			in: &proxy.Config{
+				Instances: []proxy.InstanceConnConfig{
+					{
+						Name:           "proj:region:pgfakeserver",
+						UnixSocketPath: testUnixSocketPathPg,
+					},
+				},
+			},
+			wantUnixAddrs: []string{
```
enocom, line 341: Should the fake instance result in a Unix socket?

enocom (changes requested):

internal/proxy/proxy.go
annafang-google: Replaced it with logger.Errorf(). Thanks!
Considering the scenario where this one instance was accidentally stopped, having the proxy run will save a minor effort of restarting proxy after the instance is activated. It's not a huge saving, but the behaviour could be consistent.

internal/proxy/proxy.go
annafang-google: Agreed in principle. In this particular case, if we want to support auto reconnect after the server instance is back alive, we probably need the listening [socket](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/internal/proxy/proxy.go#L809) after this code block?

internal/proxy/proxy.go
annafang-google: Same reason as line#787?

internal/proxy/proxy_test.go
annafang-google: The fake dialer's EngineVersion() added a fakeserver [case ](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/1937/files#diff-9888dbc934bcc930fe8fb9fe169d1eb08c8fb65d0a838442534e5beecfddb3ceR87) to mimic the failed instance case. EngineVersion() call will fail in the NewClient() call. What the test does is to make sure that NewClient() will proceed upon the failure of EngineVersion().

internal/proxy/proxy_test.go
annafang-google: In the case of the auto-reconnecting the instance, it might be good to keep the socket?

enocom:
A couple of high level thoughts:

1. It is possible to specify a port to allow the Proxy to run with inactive instances. In fact, we recommend production workloads do this anyway to speed up the initial startup (and avoid waiting on the metadata call)
2. It is not possible to do this with Unix sockets presently because we must know if an instance is postgres or not, and right now the Go Connector will fail if an instance isn't up ([source](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/internal/cloudsql/refresh.go#L309)). We could split that call into two parts with some work, but perhaps the value of this is less high.
3. It's probably best to just skip failed Unix sockets with an error message and call it a day given we already support this feature otherwise.

WDYT?

annafang-google:
> 

Makes sense. Updated the PR that fails the unix path if the instance is not active.

internal/proxy/proxy.go
```diff
@@ -747,6 +747,14 @@ type socketMount struct {
 	dialOpts []cloudsqlconn.DialOption
 }
 
+func getNetworkType(conf *Config, inst InstanceConnConfig) string {
```
enocom, line 750: Let's make this `networkType`. Cf. https://go.dev/doc/effective_go#Getters

internal/proxy/proxy.go
```diff
@@ -510,20 +510,20 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
 			}
-			return nil, fmt.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
 		}
 
 		l.Infof("[%s] Listening on %s", inst.Name, m.Addr())
 		mnts = append(mnts, m)
 	}
 	c.mnts = mnts
-	return c, nil
+	if len(mnts) > 0 {
```
enocom, line 523: Let's swap this with the error below, so the error path is indented.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

internal/proxy/proxy_test.go
```diff
@@ -306,6 +309,42 @@ func TestClientInitialization(t *testing.T) {
...
+					{Name: "proj:region:fakeserver", Port: 50000},
+				},
+			},
+			wantTCPAddrs: []string{
+				"127.0.0.1:50000",
+			},
+		},
+	}
+
+	unixTcs := []struct {
```
enocom, line 333: Shall we move this one test case into a dedicated test below? e.g., TestProxyInitializationWithFailedUnixSocket?

internal/proxy/proxy.go
```diff
@@ -801,6 +807,8 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 
 		address, err = newUnixSocketMount(inst, conf.UnixSocket, strings.HasPrefix(version, "POSTGRES"))
 		if err != nil {
+			c.logger.Errorf("could not mount unix socket %q for %q: %v",
```
enocom, line 810: Let's consolidate all logging up above in NewClient.

internal/proxy/proxy.go
```diff
@@ -784,7 +791,6 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 			version, err := c.dialer.EngineVersion(ctx, inst.Name)
 			if err != nil {
 				c.logger.Errorf("could not resolve version for %q: %v", inst.Name, err)
-				return nil, err
```
enocom, line 787: We need to keep this line. Anything below it would fail if we didn't return early.

internal/proxy/proxy.go
```diff
@@ -510,20 +510,20 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			for _, m := range mnts {
-				mErr := m.Close()
-				if mErr != nil {
-					l.Errorf("failed to close mount: %v", mErr)
-				}
+			c.logger.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
```
enocom, line 513: There are two cases to consider here:

1. If we have only one instance, we want to fail immediately. The Proxy can't do anything useful and so should exit with an error.
2. If there are multiple instances, we can skip over any instance that does not mount and we can log the error. If we leave the loop and there are still no instances mounted (as you've handled below), we still want to fail.

internal/proxy/proxy_test.go
```diff
@@ -306,6 +309,42 @@ func TestClientInitialization(t *testing.T) {
...
+		{
+			desc: "without TCP port or unix socket for non functional instance",
+			in: &proxy.Config{
+				Instances: []proxy.InstanceConnConfig{
+					{Name: "proj:region:fakeserver"},
+				},
+			},
+		},
+		{
+			desc: "with TCP port for non functional instance",
```
enocom, line 321: We probably also want:

1. a test case with a working TCP socket and a failing Unix socket
2. a test case with a working TCP socket and a failing TCP socket (think two TCP sockets both configured to listen on the same port).

internal/proxy/proxy.go
annafang-google: Fixed. Thanks!

internal/proxy/proxy.go
annafang-google: Fixed. Thanks!

internal/proxy/proxy.go
annafang-google: The benefit of not exiting this function is that If the user specifies the instance without TCP port, it will allow the the listen call to assign a port so that the proxy can continue to run despite the EngineVersion failure. The question is, do we want this behaviour?

internal/proxy/proxy_test.go
annafang-google: Added these two cases. Also let proxy exit if the tcp socket spec is wrong.

internal/proxy/proxy.go
```diff
@@ -510,19 +510,25 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			for _, m := range mnts {
-				mErr := m.Close()
```
enocom, line 514: If we return with an error anywhere below, we should still clean up any open sockets. So this code needs to be moved into a function and called in our two error exits below.

internal/proxy/proxy.go
```diff
@@ -510,19 +510,25 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
...
-					l.Errorf("failed to close mount: %v", mErr)
-				}
+			c.logger.Errorf("[%v] Unable to mount socket: %v", inst.Name, err)
+			if networkType(conf, inst) == "unix" {
+				continue
+			}
+			switch err.(type) {
+			// this is not the inactive instance case. Fail the proxy
+			// so that the user could fix the proxy command error.
+			case *net.OpError:
```
enocom, line 520: In what cases does this error get returned? Could we add a comment clarifying that?

internal/proxy/proxy.go
enocom: I see that we're not doing that consistently above. We should confirm that we're not logging this error twice, though.

internal/proxy/proxy.go
enocom: I think we want to push users to specify a port globally (e.g., --port=5000) or on a per-instance basis.

The behavior we want to add in this PR, is to skip failed instanced for Unix sockets only. This way we still come up even if one of the instances is instance. But this behavior only applies for Unix sockets (which you've captured above).

internal/proxy/proxy.go
enocom: Since this isn't Unix related, I think we shouldn't add it.

enocom (commented):
Key point: let's add support here for skipped over failed Unix sockets. Otherwise, we should require users specify a port for TCP sockets to avoid adding any additional complexity here.

internal/proxy/proxy.go
annafang-google: Added the comments. This is to handle the case where listen call could fail.

internal/proxy/proxy_test.go
```diff
@@ -711,3 +725,86 @@ func TestRunConnectionCheck(t *testing.T) {
...
+					{Name: mysql, Port: 60000},
+				},
+			},
+			wantSuccess: false,
+		},
+	}
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			_, err := proxy.NewClient(ctx, &fakeDialer{}, testLogger, tc.in)
+			if tc.wantSuccess != (err == nil) {
```
enocom, line 805: Could this be:

``` golang
if tc.wantSuccess && err != nil {
// ...
}
```

internal/proxy/proxy.go
```diff
@@ -510,19 +510,29 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
-			for _, m := range mnts {
-				mErr := m.Close()
-				if mErr != nil {
-					l.Errorf("failed to close mount: %v", mErr)
+			switch err.(type) {
```
enocom, line 513: Instead of switching on error type, could we just continue if the socket type is a Unix socket (like what we had before)?

internal/proxy/proxy.go
annafang-google: As discussed offline, removed the NewClient logs and kept logs at the newSocketMount level.

internal/proxy/proxy.go
annafang-google: Cleanup code has been added.

internal/proxy/proxy_test.go
annafang-google: For the proposed change, if tc.wantSuccess is false (we want NewClient() to fail), the branch will not be executed regardless what the err is. This will miss reporting the failing case when err is nil and we want NewClient() to fail.

internal/proxy/proxy.go
annafang-google: It will miss the case when the type is TCP socket and there is no active instance. Correct me if I'm wrong, my understanding from the discussion offline:
- Inactive instance with unix socket should fail
- Inactive instance with TCP socket with no port number configured should fail.

annafang-google:
As discussed offline, keep the logic to fail the proxy if the proxy is unable to connect all instances. Retain the logging to keep it consistent. 
The recommended way to keep the proxy running is to specify the port number in the configuration for TCP case. The unix socket case however, because the proxy has no way to identify if the db is postgres (for inactive instance), the code will fail the proxy in this case.

enocom (approved):

PR #1947, opened 2023-09-13
feat: /quitquitquit api now responds to HTTP GET and POST requests.

### LINKED ISSUE #1946: Allow HTTP GET to /quitquitquit admin api for k8s compatibility
### Feature Description

Kubernetes pod [lifecycle handler](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#LifecycleHandler) allows the kubernetes api to make an HTTP GET request to a pod after the pod starts up or shuts down. This could be used to gracefully terminate the proxy when another container on the pod exits. 

However, the proxy will only respond to HTTP POST requests to /quitquitquit. K8s lifecycle handler cannot make a POST request. 

So in order to use the lifecycle handler, the proxy /quitquitquit endpoint should respond to both GET and POST request.

hessjcg:
The /quitquitquit endpoint used to return a 400 error when it received a request other than an HTTP POST. Now, it 
will shut down the proxy if it receives either a GET or a POST request. This will make it possible for kubernetes
pod lifecycle handlers to gracefully signal the proxy to shut down.
 
Fixes #1946

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1307,7 +1307,7 @@ func TestPProfServer(t *testing.T) {
 	}
 }
 
-func TestQuitQuitQuit(t *testing.T) {
+func TestQuitQuitQuitHttpPost(t *testing.T) {
 	c := NewCommand(WithDialer(&spyDialer{}))
 	c.SilenceUsage = true
 	c.SilenceErrors = true
@@ -1320,7 +1320,7 @@ func TestQuitQuitQuit(t *testing.T) {
 		err := c.ExecuteContext(ctx)
 		errCh <- err
 	}()
-	resp, err := tryDial("GET", "http://localhost:9192/quitquitquit")
+	resp, err := tryDial("HEAD", "http://localhost:9192/quitquitquit")
 	if err != nil {
 		t.Fatalf("failed to dial endpoint: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


cmd/root_test.go
```diff
@@ -1307,7 +1307,7 @@ func TestPProfServer(t *testing.T) {
 	}
 }
 
-func TestQuitQuitQuit(t *testing.T) {
+func TestQuitQuitQuitHttpPost(t *testing.T) {
```
enocom, line 1310: HTTP

cmd/root_test.go
```diff
@@ -1348,6 +1348,41 @@ func TestQuitQuitQuit(t *testing.T) {
 	}
 }
 
+func TestQuitQuitQuitHttpGet(t *testing.T) {
+	c := NewCommand(WithDialer(&spyDialer{}))
+	c.SilenceUsage = true
+	c.SilenceErrors = true
+	c.SetArgs([]string{"--quitquitquit", "--admin-port", "9192", "my-project:my-region:my-instance"})
```
enocom, line 1355: Let's use 9193 or some other port to avoid conflicts with the test above.

cmd/root_test.go
```diff
@@ -1348,6 +1348,41 @@ func TestQuitQuitQuit(t *testing.T) {
 	}
 }
 
+func TestQuitQuitQuitHttpGet(t *testing.T) {
```
enocom, line 1351: HTTP

cmd/root_test.go
hessjcg: Fixed.

cmd/root_test.go
hessjcg: Fixed.

cmd/root_test.go
hessjcg: Fixed.

cmd/root_test.go
```diff
@@ -1348,6 +1348,41 @@ func TestQuitQuitQuit(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	errCh := make(chan error)
+	go func() {
+		err := c.ExecuteContext(ctx)
+		errCh <- err
+	}()
+
+	resp, err := tryDial("GET", "http://localhost:9192/quitquitquit")
```
enocom, line 1365: Maybe we pull the port out as a constant to avoid this kind of mistake (9192 here, 9193 above)

enocom (approved):

PR #1950, opened 2023-09-14
docs: Update the example healthcheck config 

### LINKED ISSUE #1943: Explicitly open port 9090 on the cloud-sql-proxy sidecar for k8s to hit the health check endpoints
### Description

The [documentation for adding health checks](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/tree/main/examples/k8s-health-check#running-cloud-sql-auth-proxy-with-health-checks-in-kubernetes) to the cloud-sql-proxy sidecar is mostly very complete and well-written, and gave good guidance on the how and why of the whole process.

However, after implementing the example chart as shown, Kubernetes was not able to hit the health checks on the default port 9090, so the pods were constantly being killed and rebuilt, leaving the whole deployment unhealthy.

### Potential Solution

My final solution was to explicitly open port 9090 on the cloud-sql-proxy sidecar:

```
          ports:
            - name: cloud-sql-proxy
              containerPort: 9090
              protocol: TCP
```

I do not have complete knowledge of my Kubernetes setup, so there may be something hidden in the configuration that changed the default behavior such that I experienced my issue above. However, I'd like to see the example code updated to include explicitly opening that port, as it may save many other people a similar headache if they're in my situation.

### Additional Details

Kubernetes version: 1.25

hessjcg:
The configuration example for health checks now matches match the default configuration used by the [cloud-sql-proxy-operator](https://github.com/GoogleCloudPlatform/cloud-sql-proxy-operator). 

Fixes #1943

### DIFF
--- a/examples/k8s-health-check/README.md
+++ b/examples/k8s-health-check/README.md
@@ -54,7 +54,7 @@ livenessProbe:
    # We recommend adding a liveness probe to the proxy sidecar container.
    httpGet:
       path: /liveness
-      port: 9090
+      port: 9801
    # Number of seconds after the container has started before the first probe is scheduled. Defaults to 0.
    # Not necessary when the startup probe is in use.
    initialDelaySeconds: 0
@@ -74,35 +74,40 @@ livenessProbe:
 # We do not recommend adding a readiness probe under most circumstances
 ```
 
-2. Add `--http-address` and `--http-port` (optional) to your
+2. Enable the health checks by setting `--http-address` and `--http-port` (optional) to your
    proxy container configuration under `command: `.
     > [proxy_with_http_health_check.yaml](proxy_with_http_health_check.yaml#L53-L76)
 
 ```yaml
 args:
-  # If connecting from a VPC-native GKE cluster, you can use the
-  # following flag to have the proxy connect over private IP
-  # - "--private-ip"
+# Replace <INSTANCE_CONNECTION_NAME> with the instance connection
+# name in the format: "project_name:region:instance_name"
+# Replace <DB_PORT> with the port that the proxy should open
+# to listen for database connections from the application
+- <INSTANCE_CONNECTION_NAME>?port=<DB_PORT>
 
-  # Enable HTTP health checks
-  - "--health-check"
+env:
+# It can be easier to manage the k8s configuration file when you
+# use environment variables instead of CLI flags. This is the
+# recommended configuration. This configuration is enabled by default
+# when the cloud-sql-proxy-operator configures a proxy image
 
-  # Listen on all addresses so the kubelet can reach the endpoints
-  - "--http-address=0.0.0.0"
+# Enable HTTP healthchecks on port 9801. This enables /liveness,
+# /readiness and /startup health check endpoints. Allow connections
+# listen for connections on any interface (0.0.0.0) so that the
+# k8s management components can reach these endpoints.
+- name: CSQL_PROXY_HEALTH_CHECK
+  value: "true"
+- name: CSQL_PROXY_HTTP_PORT
+  value: "9801"
+- name: CSQL_PROXY_HTTP_ADDRESS
+  value: 0.0.0.0
 
-  # Set the port where the HTTP server listens
-  # - "--http-port=9090"
+# Configure the proxy to exit gracefully when sent a k8s configuration
+# file.
+- name: CSQL_PROXY_EXIT_ZERO_ON_SIGTERM
+  value: "true"
 
-  # Enable structured logging with LogEntry format:
-  - "--structured-logs"
-
-  # This flag specifies where the service account key can be found
-  # Remove this argument if you are using workload identity
-  - "--credentials-file=/secrets/service_account.json"
-
-  # Replace DB_PORT with the port the proxy should listen on
-  - "--port=<DB_PORT>"
-  - "<INSTANCE_CONNECTION_NAME>"
 ```
 
 ### Readiness Health Check Configuration
@@ -145,7 +150,7 @@ seconds, 6 x 30sec = 3 minutes.)
 readinessProbe:
   httpGet:
     path: /readiness
-    port: 9090
+    port: 9801
   initialDelaySeconds: 30
   # 30 sec period x 6 failures = 3 min until the pod is terminated
   periodSeconds: 30
@@ -176,7 +181,7 @@ for more than 1 minute.
     readinessProbe:
         httpGet:
             path: /readiness
-            port: 9090
+            port: 9801
         initialDelaySeconds: 10
         # 5 sec period x 12 failures = 60 sec until the pod is terminated
         periodSeconds: 5
--- a/examples/k8s-health-check/README.md
+++ b/examples/k8s-health-check/README.md
@@ -82,15 +82,18 @@ livenessProbe:
 args:
 # Replace <INSTANCE_CONNECTION_NAME> with the instance connection
 # name in the format: "project_name:region:instance_name"
-# Replace <DB_PORT> with the port that the proxy should open
-# to listen for database connections from the application
-- <INSTANCE_CONNECTION_NAME>?port=<DB_PORT>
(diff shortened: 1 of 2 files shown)


examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -159,6 +159,161 @@ spec:
...
+        - name: CSQL_PROXY_USER_AGENT
+          value: cloud-sql-proxy-operator/unknown
+
+        # Enable structured logging with LogEntry format
+        - name: CSQL_PROXY_STRUCTURED_LOGS
+          value: "true"
+
+          # This flag specifies where the service account key can be found
+          # Remove this argument if you are using workload identity
+        - name: CSQL_CREDENTIALS_FILE
```
enocom, line 221: Do we want to show how to do this? Or do we want to use WI by default?

examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -159,6 +159,161 @@ spec:
...
+        # metrics and health check endpoints.
+        ports:
+        - containerPort: 9801
+          protocol: TCP
+        # Resource configuration depends on an application's requirements. You
+        # should adjust the following values based on what your application
+        # needs. For details, see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+        resources:
+          requests:
+            cpu: "500e-3"
```
enocom, line 302: This doesn't match what we have elsewhere. Is there a reason?

examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -159,6 +159,161 @@ spec:
...
+        securityContext:
+          # The default Cloud SQL Auth Proxy image runs as the
+          # "nonroot" user and group (uid: 65532) by default.
+          runAsNonRoot: true
+          # Use a read-only filesystem
+          readOnlyRootFilesystem: true
+          # Do not allow privilege escalation
+          allowPrivilegeEscalation : false
+        volumeMounts:
+        - name: <YOUR-SA-SECRET-VOLUME>
```
enocom, line 312: Same here -- do we want to highlight key file usage?

examples/k8s-health-check/README.md
```diff
@@ -74,35 +74,40 @@ livenessProbe:
...
-  - "--structured-logs"
-
-  # This flag specifies where the service account key can be found
-  # Remove this argument if you are using workload identity
-  - "--credentials-file=/secrets/service_account.json"
+# Replace <INSTANCE_CONNECTION_NAME> with the instance connection
+# name in the format: "project_name:region:instance_name"
+# Replace <DB_PORT> with the port that the proxy should open
+# to listen for database connections from the application
+- <INSTANCE_CONNECTION_NAME>?port=<DB_PORT>
```
enocom, line 87: Why are we using the query string syntax instead of `--port=<DB_PORT>`?

enocom: Or `CSQL_PROXY_PORT=<DB_PORT>` below?

examples/k8s-health-check/README.md
```diff
@@ -74,35 +74,40 @@ livenessProbe:
...
+
+env:
```
enocom, line 89: If we recommend env vars for Kubernetes, shall we change all our other examples too?

examples/k8s-health-check/README.md
hessjcg: This example, like the operator, sets the port number on the instance instead of setting a
a global starting port number. I prefer to make a direct, unambiguous association between database 
instance and port number. However, if you prefer, I'll switch it to `--port=<DB_PORT>`

examples/k8s-health-check/README.md
hessjcg: I'm on board with that approach. That will be in a different PR though.

examples/k8s-health-check/README.md
enocom: I think the port flag is clear because there aren't more instances involved. It also matches our other examples.

examples/k8s-health-check/README.md
enocom: Let's file an issue to track that work and link to it here then.

hessjcg: Done. #1969

enocom (changes requested):
Let's use the `--port` syntax to be consistent with other examples.

hessjcg:
Fixed.

examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -49,90 +49,104 @@ spec:
...
+
         args:
-          # If connecting from a VPC-native GKE cluster, you can use the
-          # following flag to have the proxy connect over private IP
-          # - "--private-ip"
+        # Replace <INSTANCE_CONNECTION_NAME> with the instance connection
+        # name in the format: "project_name:region:instance_name"
+        # Replace <DB_PORT> with the port that the proxy should open
+        # to listen for database connections from the application
+        - <INSTANCE_CONNECTION_NAME>
```
enocom, line 60: FYI This could be an environment variable too. `CSQL_PROXY_INSTANCE_CONNECTION_NAME=...`

examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -49,90 +49,104 @@ spec:
...
-          - "<INSTANCE_CONNECTION_NAME>"
+        # Enable the admin api server (which only listens for local connections)
+        # and enable the /quitquitquit endpoint. This allows other pods
+        # to shut down the proxy gracefully when they are ready to exit.
+        - name: CSQL_PROXY_QUITQUITQUIT
+          value: "true"
+        - name: CSQL_PROXY_ADMIN_PORT
+          value: "9092"
+        - name: CSQL_PROXY_USER_AGENT
+          value: cloud-sql-proxy-operator/unknown
```
enocom, line 105: I don't think we should add this here -- this is a private implementation detail for our purposes here.

examples/k8s-health-check/proxy_with_http_health_check.yaml
enocom: Not sure if we want to do that, though.

hessjcg: I avoided that in the operator's implementation because  when configuring multiple instances, the envvar was somewhat more confusing than the args.

examples/k8s-health-check/proxy_with_http_health_check.yaml
hessjcg: Good catch, fixed.

examples/k8s-health-check/proxy_with_http_health_check.yaml
```diff
@@ -49,90 +49,100 @@ spec:
       - name: cloud-sql-proxy
         # It is recommended to use the latest version of the Cloud SQL Auth Proxy
         # Make sure to update on a regular schedule!
-        image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.1.0
+        image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.6.0
```
enocom, line 52: 2.7.0

enocom (approved):

examples/k8s-health-check/proxy_with_http_health_check.yaml
w32-blaster: this would be highly appreciated to have. We can't move from 1.x to 2.x because of this missing. I even opened a bugticket through the support: https://issuetracker.google.com/issues/273982124

enocom: @w32-blaster would you mind opening an issue so we can track that here?

PR #1953, opened 2023-09-14
docs: Add kubernetes job example

### LINKED ISSUE #1941: Document using cloudsql auth proxy in kubernetes jobs
### Question

I need to use this in a kubernetes job,
While it works in the side car pattern, the job is stuck in NotReady status since the auth proxy never exit.

any way to trigger an exit code with status code that will make the job succeed?

### Code

_No response_

### Additional Details

_No response_

hessjcg:
This example demonstrates how to use the /quitquitquit endpoint in a kubernetes job to allow job pods
to exit successfully when the main application container exits.

Fixes #1941

### DIFF
--- a/examples/k8s-sidecar/job_with_shutdown_hook.yaml
+++ b/examples/k8s-sidecar/job_with_shutdown_hook.yaml
@@ -49,7 +49,7 @@ spec:
         - -c
         - >
           my_batch_job --host=127.0.0.1 --port=<DB_PORT> --username=<DB_USER> --dbname=<DB_NAME>
-          wget --post-data '' http://localhost:9091/quitquitquit
+          curl http://localhost:9091/quitquitquit
         image: busybox
         imagePullPolicy: IfNotPresent
         resources: {}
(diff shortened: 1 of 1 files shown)


examples/k8s-sidecar/job_with_shutdown_hook.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: my-application
+        # Run your batch job command.
+        # Then, send a HTTTP POST request to the proxy sidecar container's
+        # /quitquitquit api. This will cause the proxy process to exit.
+        command:
+        - sh
+        - -c
+        - >
+          my_batch_job --host=127.0.0.1 --port=<DB_PORT> --username=<DB_USER> --dbname=<DB_NAME>
+          wget --post-data '' http://localhost:9091/quitquitquit
```
enocom, line 52: Does this have to be `--post-data` now that we're supporting GET?

Also curl makes this easier:

``` shell
curl -X POST http://localhost:9091/quitquitquit
```

hessjcg: No, now that #1947 has merged. I'll update this accordingly.

enocom (approved):

hessjcg:
#1969 tracks updating the other docs.

PR #2014, opened 2023-11-02
fix: use proper error instance to write error log

FGtatsuro:
Now, we use cloud-sql-proxy, and encounter an error as follows.

```
[{my-instance-name}] connection aborted - error writing to client: <nil>
```

we checked code, and noticed that nil error instance was used for logging.


jackwotherspoon (approved):
Thanks so much for this @FGtatsuro! Good find 😄

FGtatsuro:
@ jackwotherspoon
Thank you for your quick response! 🙇

PR #2041, opened 2023-11-21
feat: add support for wait command

### LINKED ISSUE #1117: Add a wait-until-proxy-is-up-and-running.sh script
It is a well known issue there is no way to sequence container startups in Kubernetes (e.g. https://github.com/kubernetes/kubernetes/issues/65502). As far as I know, this is the best solution we currently have: https://github.com/istio/istio/issues/11130#issuecomment-645424556

This would be made a lot better if the images provided some `wait-until-proxy-is-up-and-running.sh` script (especially since the default image lacks `sleep`, `curl`, etc.) so that users can easily force Kuberentes to wait until the proxy is up before it starts the next container. Could be as simple as the proxy creating a file when it finishes starting up.

enocom:
To help ensure the Proxy is up and ready, this commit adds a wait command with an optional `--max` flag to set the maximum time to wait.

By default when invoking this command:

```
./cloud-sql-proxy wait
```

The Proxy will wait up to the maximum time for the /startup endpoint to respond. This command requires that the Proxy be started in another process with the HTTP health check enabled. If an alternate health check port or address is used, as in:

```
./cloud-sql-proxy <INSTANCE_CONNECTION_NAME> \
  --http-address 0.0.0.0 \
  --http-port 9191
```

Then the wait command must also be told to use the same custom values:

```
./cloud-sql-proxy wait \
  --http-address 0.0.0.0 \
  --http-port 9191
```

By default the wait command will wait 30 seconds. To alter this value, use:

```
./cloud-sql-proxy wait --max 10s
```

Fixes #1117

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -284,16 +284,16 @@ Waiting for Startup
 
     Sometimes it is necessary to wait for the Proxy to start.
 
-    To help ensure the Proxy is up and ready, the Proxy includes adds a wait
+    To help ensure the Proxy is up and ready, the Proxy includes a wait
     subcommand with an optional --max flag to set the maximum time to wait.
 
-    By default when invoking this command:
+    Invoke the wait command, like this:
 
     ./cloud-sql-proxy wait
 
-    The Proxy will wait up to the maximum time for the /startup endpoint to
-    respond. This command requires that the Proxy be started in another
-    process with the HTTP health check enabled. If an alternate health
+    By default, the Proxy will wait up to the maximum time for the startup
+    endpoint to respond. The wait command requires that the Proxy be started in
+    another process with the HTTP health check enabled. If an alternate health
     check port or address is used, as in:
 
     ./cloud-sql-proxy <INSTANCE_CONNECTION_NAME> \
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -280,8 +280,38 @@ Localhost Admin Server
     /quitquitquit. The admin server exits gracefully when it receives a POST
     request at /quitquitquit.
 
-(*) indicates a flag that may be used as a query parameter
+Waiting for Startup
 
+    Sometimes it is necessary to wait for the Proxy to start.
+
+    To help ensure the Proxy is up and ready, the Proxy includes adds a wait
+    subcommand with an optional --max flag to set the maximum time to wait.
+
+    By default when invoking this command:
+
+    ./cloud-sql-proxy wait
+
+    The Proxy will wait up to the maximum time for the /startup endpoint to
+    respond. This command requires that the Proxy be started in another
+    process with the HTTP health check enabled. If an alternate health
+    check port or address is used, as in:
+
+    ./cloud-sql-proxy <INSTANCE_CONNECTION_NAME> \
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -280,8 +280,38 @@ Localhost Admin Server
     /quitquitquit. The admin server exits gracefully when it receives a POST
     request at /quitquitquit.
 
-(*) indicates a flag that may be used as a query parameter
+Waiting for Startup
+
+    Sometimes it is necessary to wait for the Proxy to start.
+
+    To help ensure the Proxy is up and ready, the Proxy includes adds a wait
```
jackwotherspoon, line 287:
```suggestion
    To help ensure the Proxy is up and ready, the Proxy includes a wait
```

cmd/root.go
```diff
@@ -280,8 +280,38 @@ Localhost Admin Server
...
+    subcommand with an optional --max flag to set the maximum time to wait.
+
+    By default when invoking this command:
```
jackwotherspoon, line 290: This line is slightly confusing ... reads a little funny

jackwotherspoon (changes requested):

jackwotherspoon (approved):
LGTM

hessjcg (approved):
LGTM

vvxxvvxx:
Hi @enocom , is this feature also added in cloud-sql-proxy v1?

jackwotherspoon:
> Hi @enocom , is this feature also added in cloud-sql-proxy v1?

@vvxxvvxx The wait command is not supported in the v1 version of the Cloud SQL Proxy.

We highly recommend upgrading to v2 to take advantage of new features as v1 is mostly in maintenance mode (bug fixes and dep updates only).

You can check out the [migration guide](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/migration-guide.md) for more detailed steps to do so.

If there is anything holding you back from migrating, please let us know 😄

jackwotherspoon:
> hi @enocom could you update the examples or provide an example of how this would work with a poststart hook? I've been struggling to get it to work as described

Hi @mdering! 

Now that k8s supports sidecar configurations I would recommend taking a look at and using the Cloud SQL Proxy as an `initContainers` instead of a `postStart` hook:  Please see https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/128#issuecomment-2264787327

mdering:
hi @jackwotherspoon, thanks for your help! I was interested in using it as a sidecar for jobs and services, is that no longer the pattern you recommend? It seems like the regular application won't start until the `initcontainers` all stop according to the docs, is there something im missing?

devauxbr:
Hi @mdering, the [new official "sidecar container" support in K8S](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/) is indeed a bit misleading, as it requires using `initContainer`but with an additionnal `restartPolicy: Always` to make it a sidecar 😅

enocom:
+1 to what @devauxbr said. @mdering if you're still getting stuck, feel free to open an issue and we can either document this better, or explore other options.

PR #2106, opened 2024-02-08
feat: add support for config file

marklonquist:
This pull request adds functionality to load in configuration through a TOML, YAML or JSON file.

It supports all config options defined in `internal/proxy/proxy.go > Config struct`

Flags will always have first priority and will overwrite any other config set. This is the order of which we prioritize:
`flags > environment variables > configuration files and the defaults set by the tool`


enocom:
Haven't forgotten about this -- I'll get to it this week.

cloud-sql-proxy.config.example.toml
```diff
@@ -0,0 +1,42 @@
+# Comment are supported in toml
```
enocom, line 1: Let's remove this file and add a few more examples to the help command. Also, some kind of entry in the README (however short) is in order.

cmd/config_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+package cmd
+
+import (
+	"os"
+	"testing"
+)
+
+func TestNewCommandWithTomlConfigFile(t *testing.T) {
+	cmd, err := invokeProxyCommand([]string{"--config-file", "../testdata/cloud-sql-proxy.config.test-toml.toml"})
```
enocom, line 23: Let's make all these tests table tests. We'll need to add a setup/cleanup function:

``` golang
# table test data looks like this
var tcs := []struct{
  desc string
  args []string
  setup: func() func()
}{
  desc: "some description here",
  args: []string{ /*...*/ },
  setup: func() { 
    // t.Setenv and t.Cleanup might be another good option
    os.Setenv("CSQL_PROXY_DEBUG", false) 
    return func() { os.Unsetenv("CSQL_PROXY_DEBUG") }
  },
}

for _, tc := range tcs {
  t.Run(tc.desc, func(t *testing.T) {
    cleanup := setup()
    defer cleanup()

    // use tc.args and assert as necessary
  }
}
```

enocom: Also let's move the testdata directory under cmd, so we don't have to go up a directory to reach it (`../testdata` vs `testdata`).

cmd/root.go
```diff
@@ -75,7 +75,8 @@ func semanticVersion() string {
 func Execute() {
 	if err := NewCommand().Execute(); err != nil {
 		exit := 1
-		if terr, ok := err.(*exitError); ok {
+		var terr *exitError
+		if errors.As(err, &terr) {
```
enocom, line 79: 👏

cmd/root.go
```diff
@@ -245,6 +246,28 @@ Configuration using environment variables
...
 
+Configuration using a configuration file
+
+  Instead of using CLI flags, the Proxy may be configured using a configuration
+  file. The configuration file is a TOML, YAML or JSON file with the same keys as the
+  environment variables. The configuration file is specified with the
+  --config-file flag. An invocation of the Proxy using a configuration file
+  would look like the following:
+
+	  ./cloud-sql-proxy --config-file=config.toml
```
enocom, line 257: Let's format this to wrap at 80 characters as much as possible, and move this example just two additional spaces in (aligning with above).

enocom: Ditto below.

cmd/root.go
```diff
@@ -405,31 +400,16 @@ func NewCommand(opts ...Option) *Command {
 	rootCmd.AddCommand(waitCmd)
 
 	rootCmd.Args = func(cmd *cobra.Command, args []string) error {
-		// If args is not already populated, try to read from the environment.
-		if len(args) == 0 {
-			args = instanceFromEnv(args)
-		}
-		// Override CLI based arguments with any programmatic options.
-		for _, o := range opts {
```
enocom, line 413: We need to keep doing this below -- pass opts to loadArgs, and just keep four lines.

cmd/root.go
```diff
@@ -405,31 +400,16 @@ func NewCommand(opts ...Option) *Command {
...
-		cmd.SilenceUsage = true
-		// Errors will be handled by logging from here on.
-		cmd.SilenceErrors = true
-		return nil
+		// Get the instance connection names from flags, the environment or config file
+		return loadArgs(c, args)
+	}
+	rootCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		// Load the configuration file before running the command.
+		// this should ensure that the configuration is loaded in the correct order:
```
enocom, line 408: s/this/This/

cmd/root.go
```diff
@@ -440,6 +420,8 @@ func NewCommand(opts ...Option) *Command {
 	localFlags.BoolP("help", "h", false, "Display help information for cloud-sql-proxy")
 	localFlags.BoolP("version", "v", false, "Print the cloud-sql-proxy version")
 
+	localFlags.StringVar(&c.conf.ConfigFilepath, "config-file", c.conf.ConfigFilepath,
```
enocom, line 423: nit: &c.conf.Filepath seems feasible without the conf config duplication.

cmd/root.go
```diff
@@ -529,30 +511,169 @@ status code.`)
...
+	// The arguments are parsed. Usage is no longer needed.
+	c.SilenceUsage = true
+
+	// Errors will be handled by logging from here on.
+	c.SilenceErrors = true
+
+	return nil
+}
+
+func getViper(c *Command) (*viper.Viper, error) {
```
enocom, line 576: How about `initViper`? Cf. https://go.dev/doc/effective_go#Getters

cmd/root.go
```diff
@@ -529,30 +511,169 @@ status code.`)
...
-	return c
+		if ext != ".toml" && ext != ".yaml" && ext != ".yml" && ext != ".json" {
+			return nil, newBadCommandError(fmt.Sprintf("config file %v has an invalid extension, "+
+				"requires TOML, YAML or JSON format", c.conf.ConfigFilepath))
+		}
+
+		conf := filepath.Base(c.conf.ConfigFilepath)
+		noExt := strings.ReplaceAll(conf, ext, "")
+		v.SetConfigName(noExt)   // argument must be the name of config file without extension
+		v.SetConfigType(ext[1:]) // setting the config type makes the extension optional
```
enocom, line 598: Do we need to do this?

cmd/root.go
```diff
@@ -529,30 +511,169 @@ status code.`)
...
+		v.AddConfigPath(filepath.Dir(c.conf.ConfigFilepath))
+
+		// Attempt to load configuration from a file. If no file is found, assume
+		// configuration is provided by flags or environment variables.
+		if err := v.ReadInConfig(); err != nil {
+			// If the error is a ConfigFileNotFoundError, then ignore it.
+			// Otherwise, report the error to the user.
+			var configFileNotFoundError viper.ConfigFileNotFoundError
```
enocom, line 606: nit: the type tells us what it is, so `cErr`, or something similarly short is better.

testdata/cloud-sql-proxy.config.test-json.json
```diff
@@ -0,0 +1,4 @@
+{
+  "instance_connection_name": "proj:region:inst",
+  "debug": true
+}
```
enocom, line 4: These testdata files all need newlines.

cmd/root.go
```diff
@@ -529,30 +511,169 @@ status code.`)
...
+	}
+	return args
+}
+
+func instanceFromConfigFile(v *viper.Viper) []string {
+	var args []string
+	inst := v.GetString("instance_connection_name")
+
+	if inst == "" {
+		inst = v.GetString("instance_connection_name_0")
```
enocom, line 657: Let's add a test for this behavior, i.e., config file with:

``` toml
instance_connection_name_0 = x:y:z
instance_connection_name_1 = a:b:c
```

enocom (changes requested):
Big picture looks good. Thanks for putting this together.

cmd/root.go
marklonquist: yes - missed that. Reintroduced it in the loadArgs func 🙂

cmd/root.go
marklonquist: added a test for this 🙂

testdata/cloud-sql-proxy.config.test-json.json
marklonquist: indeed. Fixed!

cmd/root.go
marklonquist: yes, renamed to cErr.

cmd/root.go
marklonquist: changed name to initViper.

cmd/root.go
marklonquist: agreed. conf.Config is a bit much. Changed it.

cmd/root.go
marklonquist: fixed the typo!

cmd/root.go
marklonquist: in my editor is looks correct. But tried to replace the indentation with a copy of the above indentation

cmd/config_test.go
marklonquist: done!

cloud-sql-proxy.config.example.toml
marklonquist: removed! Added small info to README and an example in the help command which tells the user that they can also set other flags in the config file

marklonquist:
Pushed changes - so whenever you got time for another review @enocom , it's ready 😄

enocom:
Looking good. Let me make another careful pass and kick the tires on this a bit, and then we'll get it merged.

marklonquist:
thank you for cleaning  it up to match your code-style 👍

enocom:
Going to add a few more details to the README and we'll merge this.

enocom:
Thanks again @marklonquist. This is an awesome addition to the Proxy. 👏 👏 👏

PR #2130, opened 2024-02-26
fix: use kebab case for config file consistently

enocom:
When specifying keys in the configuration file, one must use kebab case. This commit ensures the instance connection names also use kebab case and updates the README to accurately describe the feature.

When configuring the Proxy with a config file, one must use matching flag names and specify any instances with the special key instance-connection-name, or instance-connection-name-i, where i is the zero-based index of the instance.

Follow-on to #2106

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ The configuration file format supports all flags. The key names should match
 the flag names. For example:
 
 ``` toml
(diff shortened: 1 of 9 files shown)


jackwotherspoon (approved):
LGTM, one typo to fix

enocom:
FYI @marklonquist

Issue #1124, opened 2022-03-03
Permissions error reading credentials with running docker container past 1.16


aebrahim:
## Bug Description

I expect that the container will start, but instead it errors out with a permission denied error when attempting to read the credentials file.

## Example code (or command)

For example, this works:
```
>>> docker run -v ~/developer_creds.json:/config gcr.io/cloudsql-docker/gce-proxy:1.16 /cloud_sql_proxy -instances=myproject:us-central-1:db=tcp:0.0.0.0:5432 -credential_file=/config
2022/03/03 19:18:10 current FDs rlimit set to 1048576, wanted limit is 8500. Nothing to do here.
2022/03/03 19:18:11 using credential file for authentication; email=developers@myproject.iam.gserviceaccount.com
2022/03/03 19:18:11 Listening on 0.0.0.0:5432 for myproject:us-central-1:db
2022/03/03 19:18:11 Ready for new connections
```

However, this fails:
```
docker run -v ~/developer_creds.json:/config gcr.io/cloudsql-docker/gce-proxy:1.17 /cloud_sql_proxy -instances=myproject:us-central-1:db=tcp:0.0.0.0:5432 -credential_file=/config
```

## Stacktrace
```
2022/03/03 19:19:32 current FDs rlimit set to 1048576, wanted limit is 8500. Nothing to do here.
2022/03/03 19:19:33 invalid json file "/config": open /config: permission denied
```

## Environment

1. Ubuntu 20.04

kurtisvg:
Hi @aebrahim -- the most likely cause is that the linux user in the container doesn't have permission to access your credential file. You need to make sure [uid 65532](https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/7e0c7081ed129e12d7e682351a15930a294d8392/Dockerfile#L28) can read the file.

enocom:
There are two ways to do this:

First, you can run your container as a user (and group) like so:

 ```
docker run 
  -v $PWD/key.json:/config/key.json \
  --user 65532:65532 \
  gcr.io/cloudsql-docker/gce-proxy:1.17 /cloud_sql_proxy \
  -instances=<INSTANCE_CONNECTION_NAME> \
  -credential_file=/config/key.json
```

Note, your `key.json` will need to be readable by 65532, e.g., `chmod o+r key.json` (not recommended).

Second, just switching to the alpine or buster container requires no special changes:

```
docker run \
[...]

PR #2137, opened 2024-03-11
docs: Improve Docker section in README (#2129)

### LINKED ISSUE #2129: Credentials can't be read because of ownership issue
### Question

Hey! Thanks for the Cloud SQL Auth Proxy; I've used it quite a lot in it's standalone form!

I'm trying to add a service to our `compose.yml` config using the `2-alpine` image. We want to authenticate using a Service Account JSON file which we want to mount to the container when spinning up the service. However, I get this error:

```bash
Attaching to sql-proxy
sql-proxy  | 2024/02/24 00:01:31 Authorizing with the credentials file at "/sa.json"
sql-proxy  | 2024/02/24 00:01:31 The proxy has encountered a terminal error: unable to start: error initializing dialer: Config error: open /sa.json: permission denied (connection name = "n/a")
sql-proxy exited with code 1
```
My SA JSON file is owned by me and a group that I share with other team mates who might want to use the file, though making it work for my user is the first (and most urgent) step.

I saw #1124 which is related, but the fix there doesn't work for me as I can't change the permissions over my credentials file in my environment. There's also [this](https://stackoverflow.com/questions/77856600/troubleshooting-google-cloud-sql-connection-through-auth-proxy-in-docker) SO question where they set up things almost the same as me.

### Code

```Go
proxy:
    image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2-alpine
    container_name: sql-proxy
    platform: linux/amd64
    command: [
      "--port", "12345",
      "--credentials-file", "/sa.json",
      "project:zone:instance-name"
    ]
    ports:
      - 12345:12345
    volumes:
      - type: bind
        source: ./sa.json
        target: /sa.json
```

### Additional Details

This works on my local machine but doesn't on a shared server (Ubuntu 20.04.6 LTS) where I don't have much permissions (though I own the file).

sebastian-correa:
Add a _Working with Docker and the Proxy_ section that explains how to use the proxy via Docker on a local machine, including some comments about permissions over the credentials file.

This PR closes #2129 and is related to #1124.

I'm open to any comments you might have, style or content-wise. Thanks for the opportunity to chime in!

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,47 @@ need a shell or related tools, use the Alpine or Buster images listed above.
 
 [distroless]: https://github.com/GoogleContainerTools/distroless
 
+### Working with Docker and the Proxy
+
+The containers have the proxy as an `ENTRYPOINT` so, to use the proxy from a
+container, all you need to do is specify options using the command, and expose
+the proxy's internal port to the host. For example, you can use:
+
+```shell
+docker run --publish <host-port>:<proxy-port> \
+    gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.8.2 \
+    --address "0.0.0.0" --port <proxy-port> <instance-connection-name>
+```
+
+You'll need the `--address "0.0.0.0"` so that the proxy doesn't only listen for
+connections originating from *within* the container.
+
+You will need to authenticate using one of the methods outlined in the
+[credentials](#credentials) section. If using a credentials file you must mount
+the file and ensure that the non-root user that runs the proxy has *read access*
+to the file. These alternatives might help:
+
+1. Change the group of your local file and add read permissions to the group
+with `chgrp 65532 key.json && chmod g+r key.json`.
+1. If you can't control your file's group, you can directly change the public
+permissions of your file by doing `chmod o+r key.json`.
+
+    > [!WARNING]
+    > This can be insecure because it allows any user in the host system to read
+    > the credential file which they can use to authenticate to services in GCP.
+
+For example, a full command using a JSON credentials file might look like
+
+```shell
+docker run \
+    --publish <host-port>:<proxy-port> \
+    --mount type=bind,source="$(pwd)"/sa.json,target=/config/sa.json \
+    gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.8.2 \
+    --address 0.0.0.0 \
+    --port <proxy-port> \
+    --credentials-file /config/sa.json <instance-connection-name>
+```
+
 ## Running as a Kubernetes Sidecar
 
 See the [example here][sidecar-example] as well as [Connecting from Google


sebastian-correa:
I'll tackle the CLA signing later today.

README.md
```diff
@@ -412,6 +412,30 @@ need a shell or related tools, use the Alpine or Buster images listed above.
...
+2. If you can't control your file's group, you can directly change the public permissions of your file by doing `chmod o+r key.json`.
+
+    > [!WARNING]
+    > This can be insecure because it allows any user in the host system to read the credential file, which they can use to authenticate to services in GCP.
+
+For example, a full command using a JSON credentials file might look like
+
+```bash
+docker run --publish <host-port>:<proxy-port> --mount type=bind,source="$(pwd)"/sa.json,target=/config/sa.json gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.8.2 --address 0.0.0.0 --port <proxy-port> --credentials-file /config/sa.json <instance-connection-name>
+```
```
jackwotherspoon, line 437: @enocom I thing long-lived service account keys are slowly trying to be phased out in documentation for other methods like workload identity etc. Is there a way we can showcase workload identity here instead?

sebastian-correa: We're trying to go with Workload Identity for a GH Action in a project I'm in. Coincidentally, I was going to tackle this tomorrow so, if it works and @enocom thinks it's best, I can give it a shot with the Proxy and change the `README.md`.

enocom: Yeah I'd be curious to hear how that goes. As an aside, we might want to mention here that if you run this in Cloud Shell, or a GCE VM, the auth library will find the token for you, and makes this much simpler.

For on-prem, it's a bit tricky: either WIF or a SA key is necessary.

sebastian-correa: We were able to use Workload Identity by adding this step to our GitHub Action Job:

```yaml
      - id: auth
        name: Auth with Google Cloud
        uses: "google-github-actions/auth@v2"
        with:
          project_id: ${{ vars.PROJECT_ID }}
          workload_identity_provider: ${{ secrets.WORKLOAD_IDENTITY_PROVIDER }}
```

which provides the `steps.auth.outputs.auth_token` output with the token needed to authorize.

It seems the proxy supports either a SA JSON or Application Default Credentials from `gcloud`, so how would this work from within the container? [[ref](https://cloud.google.com/docs/authentication/provide-credentials-adc#local-dev)] All I can think of is requesting a token and injecting it to the container via the `--token` flag, but that sounds a bit more complicated.

sebastian-correa: No pressure but, to clarify, I'm awaiting some guidance on how to use WIF with the proxy container to try it out and document it :).

enocom: Sorry for the slow response here. If you're OK with a short-lived workload (e.g. CI/CD), then using the token flag would probably be best.

To take a step back, are we solving for GitHub Actions here, or trying to be more generic about giving advice on using the Proxy in an on-prem containerized situation?

sebastian-correa: > are we solving for GitHub Actions here, or trying to be more generic about giving advice on using the Proxy in an on-prem containerized situation?

I think "generic advice". At least my usage is local only for now, so I haven't tried this in CI/CD.

enocom: I think for the purpose of this PR, the current context is probably adequate.

enocom (approved):
Mind wrapping the long lines at 80 characters? Then we can get this merged.

sebastian-correa:
> Mind wrapping the long lines at 80 characters? Then we can get this merged.

Done. Thanks for your patience, this PR's tab got lost in the sea of tabs I have in Brave :S.

I wrapped the lines and renamed the commit from _chore_ to _docs_.

enocom:
Thanks again for the README update @sebastian-correa.

PR #2159, opened 2024-03-25
feat: use Google managed base images

enocom:
This commit updates the Bullseye and Buster containers to pull from a Google-managed Debian image.

Context:

Google maintains base images for building its own applications. These images are built from the same source that Docker Hub uses. Therefore, they match the images you would get from Docker Hub.

The advantage of using Google-maintained images is that they are stored on Google Cloud, so you can pull these images directly from your environment without having to traverse networks.

Google updates these images whenever a new version of an official image is released and the new version fixes a known vulnerability reported by Artifact Analysis. For more information on image versions, see the GitHub repository of official images.

See https://cloud.google.com/software-supply-chain-security/docs/base-images


jackwotherspoon (approved):
LGTM

PR #2178, opened 2024-04-08
chore: add govulncheck to PR checks

enocom:


.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,49 @@
+# Copyright 2023 Google LLC
```
jackwotherspoon, line 1:
```suggestion
# Copyright 2024 Google LLC
```

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  push:
+    branches:
+    - 'main'
+  pull_request:
+  pull_request_target:
+    types: [labeled]
+  schedule:
+  - cron:  '0 2 * * *'
```
jackwotherspoon, line 28: do we want all these? Main won't work here for sure

jackwotherspoon (changes requested):

enocom:
Looks like I'm doing this wrong. We already have v1 configured.

See https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/.github/workflows/govulncheck.yaml#L24.

PR #2184, opened 2024-04-15
feat: add support for a lazy refresh

### LINKED ISSUE #2183: Add support for a lazy refresh
This is for exposing https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/772 as a CLI flag.

enocom:
When clients run the Proxy in environments where the CPU may be throttled, the background connection info refresh operation can fail to complete, causing connection errors. This commit introduces an option for a lazy refresh. Connection info is retrieved on an as needed-basis and cached based on the associated certificate's expiration. No background goroutine runs, unlike the default refresh ahead cache.

 Enable it like so:

 ./cloud-sql-proxy <INSTANCE_CONNECTION_NAME> --lazy-refresh

A lazy refresh may result in increased latency (more requests will be subject to waiting for the refresh to complete), but gains in reliability.

Fixes #2183

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/GoogleCloudPlatform/cloud-sql-proxy/v2
 go 1.22
 
 require (
-	cloud.google.com/go/cloudsqlconn v1.8.1
+	cloud.google.com/go/cloudsqlconn v1.8.2-0.20240415201134-931150f492cb
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
 	contrib.go.opencensus.io/exporter/stackdriver v0.13.14
 	github.com/coreos/go-systemd/v22 v22.5.0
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/GoogleCloudPlatform/cloud-sql-proxy/v2
 go 1.22
 
 require (
-	cloud.google.com/go/cloudsqlconn v1.8.1
+	cloud.google.com/go/cloudsqlconn v1.8.2-0.20240415201134-931150f492cb
```
enocom, line 6: TODO: Make this a released version

jackwotherspoon (changes requested):
should we add an example usage to the help message?

README.md
```diff
@@ -365,6 +365,13 @@ debug-logs = true
 
 Run `./cloud-sql-proxy --help` for more details.
 
+### Configuring a Lazy Refresh
+
+The `--lazy-refresh` flag configures the Proxy to retrieve connection info
+lazily and as-needed. Otherwise, no background refresh cycle runs. This setting
+is useful in environments where the CPU may be throttled outside of a request
+context, e.g., Cloud Run.
```
jackwotherspoon, line 373:
```suggestion
context, e.g., Cloud Run, Cloud Functions, etc.
```

jackwotherspoon (approved):
LGTM

PR #2207, opened 2024-05-01
fix: don't depend on downstream in readiness check

### LINKED ISSUE #2083: Readiness probe shouldn't depend on downstream services
Cf. https://github.com/zegl/kube-score/blob/master/README_PROBES.md#readinessprobe

Also, the readiness probe should respond differently after the process received a SIGTERM.

enocom:
The readiness probe should not depend on downstream connections. Doing so can cause unwanted downtime. See [1]. This commit removes all dialing of downstream database servers as a result.

In addition, after the Proxy receives a SIGTERM or SIGINT, the readiness check will now report an unhealthy status to ensure it is removed from circulation before shutdown completes.

[1]: https://github.com/zegl/kube-score/blob/master/README_PROBES.md#readinessprobe

Fixes #2083

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1062,8 +1062,10 @@ func runSignalWrapper(cmd *Command) (err error) {
 	var (
 		needsHTTPServer bool
 		mux             = http.NewServeMux()
-		notify          = func() {}
+		notifyStarted   = func() {}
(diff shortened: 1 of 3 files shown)


jackwotherspoon (approved):
LGTM

PR #2210, opened 2024-05-08
chore: add debug logs around shutdownCh

jackwotherspoon:
Add additional debug log entries around `shutdownCh` to allow for verification that SIGINT/SIGTERM signals are being properly sent.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1025,6 +1025,7 @@ func runSignalWrapper(cmd *Command) (err error) {
 		defer close(startCh)
 		p, err := proxy.NewClient(ctx, cmd.dialer, cmd.logger, cmd.conf)
 		if err != nil {
+			cmd.logger.Debugf("Error starting proxy. Writing error to shutdown channel.")
 			shutdownCh <- fmt.Errorf("unable to start: %v", err)
 			return
 		}
@@ -1129,7 +1130,10 @@ func runSignalWrapper(cmd *Command) (err error) {
 		)
 	}
 
-	go func() { shutdownCh <- p.Serve(ctx, notify) }()
+	go func() { 
+		cmd.logger.Debugf("Starting to listen on shutdown channel.")
+		shutdownCh <- p.Serve(ctx, notify) 
+	}()
 
 	err = <-shutdownCh
 	switch {
@@ -1159,6 +1163,7 @@ func quitquitquit(quitOnce *sync.Once, shutdownCh chan<- error) http.HandlerFunc
 			default:
 				// The write attempt to shutdownCh failed and
 				// the proxy is already exiting.
+				cmd.logger.Debugf("Could not write to shutdown channel. The proxy is already exiting.")
 			}
 		})
 	}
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
@enocom feel free to leave suggestions for additional spots I could add debug log entries and I'll happily add them 😄

enocom:
There are a few other spots. Just adding a logging message directly above these lines will be good enough I think.

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1026

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1130

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1156

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1177

jackwotherspoon:
> There are a few other spots. Just adding a logging message directly above these lines will be good enough I think.
> 
> https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1026
> 
> https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1130
> 
> https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1156
> 
> https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b4076f38c462b9a60e8885bbed2116e4f6363517/cmd/root.go#L1177

The last two don't have access to the logger so I couldn't add them...

cmd/root.go
```diff
@@ -1023,6 +1025,7 @@ func runSignalWrapper(cmd *Command) (err error) {
 		defer close(startCh)
 		p, err := proxy.NewClient(ctx, cmd.dialer, cmd.logger, cmd.conf)
 		if err != nil {
+			cmd.logger.Debugf("Error starting proxy. Writing error to shutdown channel.")
```
enocom, line 1028: nit: cmd.logger.Debugf(Error starting proxy: %v", err)

cmd/root.go
```diff
@@ -1131,7 +1134,10 @@ func runSignalWrapper(cmd *Command) (err error) {
 		)
 	}
 
-	go func() { shutdownCh <- p.Serve(ctx, notifyStarted) }()
+	go func() {
+		cmd.logger.Debugf("Starting to listen on shutdown channel.")
```
enocom, line 1138: By logging the error, we can see what Serve might be complaining about and then infer if the error is reported through the channel based on subsequent log messages.

``` golang
err := <- p.Serve(ctx, notifyStarted)
cmd.logger.Debugf("proxy server error: %v", err)
shutdownCh <- err
```

jackwotherspoon: did `err := p.Serve(ctx, notifyStarted)` instead of `err := <- p.Serve(ctx, notifyStarted)`

enocom: Yep you're correct.

PR #2227, opened 2024-05-23
chore: update README with rendered notices

jackwotherspoon:
Add rendered notices to README for better readability and to allow notices grab readers attention.

Example important notice: 
image

Example note notice
image


enocom (approved):
:clap:

PR #2257, opened 2024-06-26
fix: exit when FUSE errors on accept 

hessjcg:
This change makes FUSE-configured connections behave the same way as normal connections: 
the proxy process should exit when there is an error accepting a connection while listening on 
the local server socket.

### DIFF
--- a/internal/proxy/fuse_test.go
+++ b/internal/proxy/fuse_test.go
@@ -43,24 +43,37 @@ func randTmpDir(t interface {
 // newTestClient is a convenience function for testing that creates a
 // proxy.Client and starts it. The returned cleanup function is also a
 // convenience. Callers may choose to ignore it and manually close the client.
-func newTestClient(t *testing.T, d cloudsql.Dialer, fuseDir, fuseTempDir string) (*proxy.Client, func()) {
+func newTestClient(t *testing.T, d cloudsql.Dialer, fuseDir, fuseTempDir string) (*proxy.Client, chan error, func()) {
 	conf := &proxy.Config{FUSEDir: fuseDir, FUSETempDir: fuseTempDir}
+
+	// This context is only used to call the Cloud SQL API
 	c, err := proxy.NewClient(context.Background(), d, testLogger, conf)
 	if err != nil {
 		t.Fatalf("want error = nil, got = %v", err)
 	}
 
 	ready := make(chan struct{})
-	go c.Serve(context.Background(), func() { close(ready) })
+	servErrCh := make(chan error)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+
+		servErr := c.Serve(ctx, func() { close(ready) })
+		select {
+		case servErrCh <- servErr:
+		case <-ctx.Done():
+		}
+	}()
 	select {
 	case <-ready:
 	case <-time.Tick(5 * time.Second):
 		t.Fatal("failed to Serve")
 	}
-	return c, func() {
+	return c, servErrCh, func() {
 		if cErr := c.Close(); cErr != nil {
 			t.Logf("failed to close client: %v", cErr)
 		}
+		cancel()
 	}
 }
 
@@ -70,7 +83,7 @@ func TestFUSEREADME(t *testing.T) {
 	}
 	dir := randTmpDir(t)
 	d := &fakeDialer{}
-	_, cleanup := newTestClient(t, d, dir, randTmpDir(t))
+	_, _, cleanup := newTestClient(t, d, dir, randTmpDir(t))
 
 	fi, err := os.Stat(dir)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/proxy/fuse_test.go
```diff
@@ -185,13 +190,70 @@ func TestFUSEDialInstance(t *testing.T) {
...
+	// Create a new client
+	d := &fakeDialer{}
+	c, servErrCh, cleanup := newTestClient(t, d, fuseDir, fuseTempDir)
+	defer cleanup()
+
+	// Attempt a successful connection to the client
+	conn := tryDialUnix(t, socketPath)
+	defer conn.Close()
+
+	var got []string
```
enocom, line 212: Is this block of assertions necessary? tryDialUnix should wait until a connection has been established and we have tests elsewhere that ensure the wiring is correct (see above for example).

internal/proxy/fuse_test.go
```diff
@@ -43,21 +44,25 @@ func randTmpDir(t interface {
...
 	if err != nil {
 		t.Fatalf("want error = nil, got = %v", err)
 	}
 
 	ready := make(chan struct{})
-	go c.Serve(context.Background(), func() { close(ready) })
+	servErrCh := make(chan error)
+	go func() {
+		servErr := c.Serve(context.Background(), func() { close(ready) })
+		servErrCh <- servErr
```
enocom, line 58: This will leak goroutines if servErrCh is full.

How about:

``` golang
select {
  case servErrCh <- servErr:
  default:
    // exit background thread
}
```

internal/proxy/fuse_test.go
```diff
@@ -185,13 +190,70 @@ func TestFUSEDialInstance(t *testing.T) {
...
+	}
+
+	// Explicitly close the dialer. This will close all the unix sockets, forcing
+	// the unix socket accept goroutine to exit with an error
+	c.Close()
+
+	// Check that Client.Serve() returned a non-nil error
+	for i := 0; i < 10; i++ {
+		select {
+		case servErr := <-servErrCh:
```
enocom, line 234: I think the point of this assertion is that something come through the channel (other than nil). Do we need to assert on the specific error message?

internal/proxy/fuse_test.go
hessjcg: Fixed.

internal/proxy/fuse_test.go
hessjcg: I removed the `d.dialedInstances()` assertion.

internal/proxy/fuse_test.go
hessjcg: I removed the error string assertion.

internal/proxy/proxy_other.go
```diff
@@ -165,10 +176,16 @@ func (c *Client) serveFuse(ctx context.Context, notify func()) error {
 	}
 	c.fuseServerMu.Lock()
 	c.fuseServer = srv
+	c.fuseExitCh = make(chan error)
 	c.fuseServerMu.Unlock()
 	notify()
-	<-ctx.Done()
-	return ctx.Err()
+	select {
+	case cErr := <-c.fuseExitCh:
```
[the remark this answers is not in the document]
hessjcg: Updated to reuse `err`.

internal/proxy/fuse_test.go
```diff
@@ -43,21 +43,29 @@ func randTmpDir(t interface {
...
 	c, err := proxy.NewClient(context.Background(), d, testLogger, conf)
 	if err != nil {
 		t.Fatalf("want error = nil, got = %v", err)
 	}
 
 	ready := make(chan struct{})
-	go c.Serve(context.Background(), func() { close(ready) })
+	servErrCh := make(chan error)
+	go func() {
+		servErr := c.Serve(context.Background(), func() { close(ready) })
```
enocom, line 56: My suggestion seems to have broken your test. Maybe it's better to block on the error channel or a context that you can cancel at the end of the test run.

hessjcg: There must be some subtle timing difference in the ubuntu implementation of channels. I put the test back to how it was and now it passes. I think its OK for this goroutine to block until `servErrCh` can accept an error.

enocom:
``` golang
ctx, cancel := context.WithCancel(context.Background())
defer cancel()
```

go func() {
  servErr := c.Server(ctx, func() { close(ready) })
  select {
    case servErrCh <- serrErr:
    case <-ctx.Done():
  }
}()
```

Comment (hessjcg) [reply internal/proxy/fuse_test.go]: Fixed.

File: internal/proxy/fuse_test.go [hunk 6]
```diff
@@ -43,24 +43,37 @@ func randTmpDir(t interface {
...
 		t.Fatalf("want error = nil, got = %v", err)
 	}
 
 	ready := make(chan struct{})
-	go c.Serve(context.Background(), func() { close(ready) })
+	servErrCh := make(chan error)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+
```
enocom, line 60: nit: whitespace

enocom (approved):

PR #2271, opened 2024-07-16
feat: add support for bookworm to v1 Proxy

jackwotherspoon:
Drop support for buster and add support for bookworm (debian 12) as 
buster has fallen out of LTS.


enocom (changes requested):
Filename needs to change.

enocom (approved):

PR #2282, opened 2024-08-02
ci: skip integration tests on fork PRs

jackwotherspoon:
Pull request from forks once merged will still be skipped because `github.event.pull_request.head.repo.full_name` will still be that of the fork.

To combat this we run the job on all events that are not a pull_request (aka `schedule` and `push`)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,6 +45,10 @@ jobs:
 
   integration:
     name: integration tests
+    # run integration tests on all builds except pull requests from forks or dependabot
+    if: |
+      github.event_name != 'pull_request' || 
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
@enocom it seems for the Proxy we only have one test suite that runs `unit + integration` tests which means for forks no tests will be run.

I wonder if we should follow the Go Connector and other repos where we have one suite for just unit tests and one suite for unit + integration? This way fork PRs will still have unit tests run...

enocom:
Yes, good point. Since we're having to skip integration tests for forks, it's probably a good idea to split out the unit tests before merging this. WDYT?

enocom (approved):

PR #2283, opened 2024-08-02
ci: add unit test build

jackwotherspoon:
Currently the Proxy only has one tests suite that runs all tests.

Adding a unit tests suite that skips integration tests. 

This will allow for forked PRs to run only unit tests.

Related to #2282

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -163,8 +163,6 @@ jobs:
           go-version: 1.22
       - name: Checkout code
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-      - name: Enable fuse config
-        run: sudo sed -i 's/#user_allow_other/user_allow_other/g' /etc/fuse.conf
       - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -152,3 +152,21 @@ jobs:
...
+    name: unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+        with:
+          go-version: 1.22
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Enable fuse config
```
enocom, line 166: I believe this fuse tests don't run with the `-short` flag (or at least they shouldn't). So this should be unnecessary.

enocom (approved):

.github/workflows/tests.yaml
jackwotherspoon: Let us see, removed it

jackwotherspoon: Yup looks good

Issue #2296, opened 2024-09-12
TLS Handshake failures with Go 1.23


dan-lind:
### Bug Description

When attempting to connect through the proxy, I'm getting connection reset by peer

Starting the proxy is fine
```
2024/09/12 08:58:14 Authorizing with Application Default Credentials
...
2024/09/12 08:58:15 [redacted-prod:europe-west1:redacted-live-773ed67c] Listening on /tmp/redacted-prod:europe-west1:redacted-live-773ed67c
2024/09/12 08:58:15 [redacted-prod:europe-west1:redacted-live-1c3b6da9] Listening on /tmp/redacted-prod:europe-west1:redacted-live-1c3b6da9
2024/09/12 08:58:15 [redacted-prod:europe-west1:redacted-live-6aab2946] Listening on /tmp/redacted-prod:europe-west1:redacted-live-6aab2946
2024/09/12 08:58:15 [redacted-prod:europe-west1:redacted-live-23a55252] Listening on /tmp/redacted-prod:europe-west1:redacted-live-23a55252
2024/09/12 08:58:15 The proxy has started successfully and is ready for new connections!
2024/09/12 08:58:47 [redacted-prod:europe-west1:redacted-live-156c5af9] Accepted connection from
```

but when I attempt to make the actual connection, I get

```
2024/09/12 08:58:47 [redacted-up-prod:europe-west1:redacted-live-6aab2946] failed to connect to instance: Dial error: handshake failed (connection name = "redacted-up-prod:europe-west1:redacted-live-6aab2946"): read tcp 172.22.164.170:53244->redacted:3307: read: connection reset by peer
```

In the client (Datagrip 2024.2.1) I see the error `Socket is not connected.`
On the CloudSQL side, I can also see a warning log in Logs explorer like `1893634 [Warning] [MY-000000] [Server] Failed to connect to mysql: 1893634. Server-side error number: 1159`

### Stacktrace

```bash
Output from tcpdump


tcpdump: data link type PKTAP
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on any, link-type PKTAP (Apple DLT_PKTAP), snapshot length 524288 bytes
08:58:47.153440 IP 172.22.164.170.53244 > redacted.3307: Flags [S], seq 2094967633, win 65535, options [mss 1360,nop,wscale 6,nop,nop,TS val 1730041868 ecr 0,sackOK,eol], length 0
[...]

PR #2303, opened 2024-10-07
Revert "feat: bump to Go 1.23 (#2287)"

TokieSan:
This reverts commit fd6bd49242c884508f641c754eb5cec5d28ac665. 

Solves #2296 since Go 1.23 has issues with TLS fallbacks & the release notes didn't cover any security related concerns or modifications that would affect commits after the bump.


jackwotherspoon:
Thanks @TokieSan for this PR. However, I think for now we will go forward with recommending users pin to an older version while we continue investigating the root cause (Go issue vs Cloud SQL server side issue). 

If the investigation points at needing to revert the Go version change I will re-open this in the future.

Thanks!

PR #2353, opened 2024-12-18
chore: add "Configuring IAM Database Authentication" to README

jackwotherspoon:
Adding section to README for IAM database authentication.

Added warning highlighting nuances that can trip users up.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,40 @@ query param:
 > ./cloud-sql-proxy --unix-socket C:\cloudsql myproject:my-region:mysql
 > ```
 
+### Configuring IAM Database Authentication
+
+The Proxy supports [Automatic IAM Database Authentication][iam-auth] for MySQL
+and Postgres instances, allowing IAM principal's to authenticate and connect
+as database users.
+
+Make sure to configure your [Cloud SQL instance to allow IAM authentication][iam-auth-config]
+and to [add your IAM principal as a database user][iam-auth-user].
+
+```shell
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
cc @fragile-ds would love to hear if this change would have helped you 😄

README.md
```diff
@@ -338,6 +338,41 @@ query param:
...
+```
+
+> [!IMPORTANT]
+> 
+> Make sure to run the Proxy as the same IAM principal as the database user
+> you want to log in as. Only the IAM principal that is attached to the
+> [sourced credentials](#credentials) will be able to successfully log in
+> via automatic IAM database authentication.
+>
+> When logging in using an IAM database user, the username must be
```
hessjcg, line 361: Maybe mention the name length restrictions when describing how IAM usernames get truncated.
[mysql - 32 chars](https://dev.mysql.com/doc/refman/8.4/en/user-names.html#:~:text=MySQL%20user%20names%20are%20up,have%20a%20different%20maximum%20length.&text=The%20MySQL%20user%20name%20length,mysql%20database%20does%20not%20work.)
[postgres - 63 chars](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#:~:text=The%20system%20uses%20no%20more,identifier%20length%20is%2063%20bytes.)

hessjcg (approved):

README.md
jackwotherspoon: Good call, done!

PR #2363, opened 2025-01-15
feat: enable Proxy to be started with DNS names

jackwotherspoon:
Add default Go Connector Option [`WithDNSResolver`](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/dcd433d098bbf0b1cdd23037a01770bd72485302/options.go#L269) to the Cloud SQL Proxy.

This Option allows the Proxy to be started with either an instance connection name or a DNS name.

The `WithDNSResolver` Option configures the underlying Cloud SQL Go Connector
to first attempt to parse value as an instance connection name, if it is not a valid
instance connection name then attempt to parse as DNS name.


jackwotherspoon:
Integration test will be added as a fast follow, waiting for backend changes propagate

jackwotherspoon:
Added integration test for customer CAS instance with instance connection name and domain name

hessjcg (approved):
Nice test.

PR #2373, opened 2025-01-21
fix: Handle the errors when the listener socket is closed gracefully.

### LINKED ISSUE #2372: This may result in an exit code error status
When the proxy receives a SIGTERM, it will close this server socket. This can result in the server.Accept() returning a non-nil error with the message
"proxy server error: accept tcp" or "use of closed network connection".  This error will be put into the error channel, causing the process end with exit code 2, instead of the appropriate 143 (TERMINATED) or 0 (SUCCESS).

hessjcg:
The accept loop for socket connections now will properly handle the error case where the listener socket
is closed gracefully.  

Fixes #2372


enocom:
Nice find! This probably deserves a test, if only of the private method, so that future maintainers don't trip over this again.

hessjcg:
I'm converting this back to a draft, because after some testing, the errors returned from `l.Accept()` on socket close are not exactly as documented.

PR #2395, opened 2025-02-25
chore: remove Go toolchain from v1 Proxy

jackwotherspoon:
No need for toolchain in Go 1.23


jackwotherspoon:
Obsolete due to #2400

PR #2398, opened 2025-02-27
chore: add Supported Go Versions to README.md

jackwotherspoon:
Adding Supported Go Versions policy to README.


jackwotherspoon:
Don't actually need this

PR #2401, opened 2025-03-06
docs: update wait command help message

sjeandeaux:
Without --health-check or --prometheus, the server http does not start

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -332,8 +332,8 @@ Configuration
 
   By default, the Proxy will wait up to the maximum time for the startup
   endpoint to respond. The wait command requires that the Proxy be started in
-  another process with the HTTP health check or prometheus enabled. If an alternate health
-  check port or address is used, as in:
+  another process with the HTTP health check or Prometheus enabled. If an
+  alternate health check port or address is used, as in:
 
   ./cloud-sql-proxy <INSTANCE_CONNECTION_NAME> \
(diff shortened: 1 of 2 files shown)


jackwotherspoon:
@sjeandeaux Thanks for the PR! 😄 

All the docs in our `docs/` folder are auto-generated from our help messages.

So instead of updating the docs you will want to update the help message directly...

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/bfb63f543b1112306ee62e5210cb609578aeef04/cmd/root.go#L331-L340

cmd/root.go
```diff
@@ -332,12 +332,13 @@ Configuration
 
   By default, the Proxy will wait up to the maximum time for the startup
   endpoint to respond. The wait command requires that the Proxy be started in
-  another process with the HTTP health check enabled. If an alternate health
+  another process with the HTTP health check or prometheus enabled. If an alternate health
   check port or address is used, as in:
```
jackwotherspoon, line 336: nit: keep line length under 80 chars and capitalize Prometheus

```suggestion
  another process with the HTTP health check or Prometheus enabled. If an
  alternate health check port or address is used, as in:
```

jackwotherspoon (changes requested):
One small nit, otherwise LGTM 👍

jackwotherspoon (approved):

PR #2403, opened 2025-03-08
Add note about the new executable name

bseib:
It is a subtle difference to notice in all the examples that the executable name has changed from using underscores to hyphens. It would be helpful to the reader to have this explicitly pointed out, the same way that the other changes have been spelled out.

(Yes, it is obvious and self-evident when you are typing the old/new command on the command line that `cloud_sql_proxy` does not exist. But I lost a good bit time pinpointing this problem when it was inside a k8s container that had been converted from v1 to v2, even when I got all the new flag changes correct! 😀)

### DIFF
--- a/migration-guide.md
+++ b/migration-guide.md
@@ -34,9 +34,25 @@ and then attempt to use a private IP. In v2, the Proxy now defaults to public
 IP without trying private IP. If you want to use private IP, you must pass
 either the `--private-ip` flag or the query parameter. See the README for details.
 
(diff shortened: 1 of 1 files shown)


jackwotherspoon (approved):

jackwotherspoon:
Thanks again @bseib 😄 

If you have any other general feedback about the Proxy or our docs, please let me know!

We definitely value feedback and will make sure to create docs/feature bugs for any enhancements or suggestions.

bseib:
Thanks for the quick review and merge @jackwotherspoon !

PR #2419, opened 2025-04-02
chore: remove reference to min-ready

dmarkwat:
Addresses a gap in the [documented behavior](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/b6ca9c52ca41dfd0ceaf5dce104a533410e6dfe0/cmd/root.go#L179-L185) for `/readiness` checks which states:
> Optionally supports a min-ready query param (e.g., /readiness?min-ready=3) where the Proxy will return a 200 status if the Proxy can connect successfully to at least min-ready number of instances. If min-ready exceeds the number of registered instances, returns a 400.

This is a proposal to close that gap. One deficiency is clear, though, which is that it opens connections as part of the readiness endpoint which may slow its response time in certain connection conditions. Curious if there's suggestions to address this, but ultimately the documented feature isn't implemented and needs to be to properly support HA cloudsql configurations.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -179,10 +179,7 @@ Health checks
   - /readiness: Returns 200 status when the Proxy has started, has available
   connections if max connections have been set with the --max-connections
   flag, and when the Proxy can connect to all registered instances. Otherwise,
-  returns a 503 status. Optionally supports a min-ready query param (e.g.,
-  /readiness?min-ready=3) where the Proxy will return a 200 status if the
-  Proxy can connect successfully to at least min-ready number of instances. If
-  min-ready exceeds the number of registered instances, returns a 400.
+  returns a 503 status.
 
   - /liveness: Always returns 200 status. If this endpoint is not responding,
   the Proxy is in a bad state and should be restarted.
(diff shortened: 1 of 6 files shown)


enocom:
We actually removed this deliberately and didn't update the docs :facepalm: 

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/2207

So this is could be a documentation PR.

dmarkwat:
Fair. Unfortunate, but fair.

Perhaps you can lend some insight then into how best to handle this particular situation:

We are configuring HA databases and the idea is that when one of them is down, the other should still be accessible -- bc HA, of course. Will readiness still report as operable if one of the replicas configured is down? What will happen int he event all are down? Will readiness still report OK?

The doc is mum on this which is why we pursued this. We'd seen behavior where when one was down readiness no longer said 200/OK and the serving pods were pulled from their k8s Services, but this may have been isolated to cloud sql v1.

enocom:
Would you mind opening an issue for this so we can help other people discover it?

Short answer: configure readiness probes through your app rather than the Proxy. These probes might then traverse the proxy in reaching to the backing instance. This way you can configure whatever behavior makes sense for the problem you're trying to solve.

dmarkwat:
Opened https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/2420

I'll revert my changes and update the doc in this PR instead.

enocom:
Thank you!

jackwotherspoon (approved):
LGTM 👍  Thanks @dmarkwat 👏

PR #2421, opened 2025-04-04
refactor: replace zap with slog

### LINKED ISSUE #2120: Replace zap with log/sloq
### Feature Description

Now that structured logging has landed in the standard library, we can shed an external dependency and just use log/slog.

### Sample code

```
// sample code here
```

### Alternatives Considered

_No response_

### Additional Details

_No response_

danielandersson:
This PR replaces the third-party Zap logging library with Go's standard library `slog` package for structured logging. This change reduced the number of dependencies while maintaining the same logging functionality and format.

Fixes #2120

### Changes
- Removed dependencies on `go.uber.org/zap` and `go.uber.org/zap/zapcore`
- Added usage of Go's standard library `log/slog` package
- Simplified the `StdLogger` structure by combining `infoLog` and `debugLog` into a single `stdLog`
- Refactored the `StructuredLogger` to use `slog.Logger` instead of Zap's `SugaredLogger`
- Modified the `NewStructuredLogger` function to use `slog` handlers and configuration
- Added a new `replaceAttr` function to maintain compatibility with Google Cloud Logging format
- Ensured that log messages continue to adhere to the LogEntry format for Google Cloud Logging


jackwotherspoon:
Thanks @danielandersson 👏 

I will take a look at this in the coming days 😄

jackwotherspoon:
@hessjcg mind taking a look at this one?

Also @enocom you might also want to take a look as a similar change would probably be added to AlloyDB Proxy 😄

internal/log/log.go
```diff
@@ -54,61 +60,73 @@ func (l *StdLogger) Errorf(format string, v ...interface{}) {
...
+}
+
+// replaceAttr remaps default Go logging keys to adhere to LogEntry format
+// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
+func replaceAttr(groups []string, a slog.Attr) slog.Attr {
+	if groups == nil {
+		if a.Key == slog.LevelKey {
+			a.Key = googLvlKey
+			return a
+		} else if a.Key == slog.MessageKey {
```
enocom, line 117: We can remove all these `else` usages, since they immediately follow a return statement.

internal/log/log.go
```diff
@@ -54,61 +60,73 @@ func (l *StdLogger) Errorf(format string, v ...interface{}) {
...
+		stdLog: slog.New(infoHandler),
+		errLog: slog.New(errorHandler),
+	}
+	return l
+}
+
+// replaceAttr remaps default Go logging keys to adhere to LogEntry format
+// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
+func replaceAttr(groups []string, a slog.Attr) slog.Attr {
+	if groups == nil {
```
enocom, line 113: Let's invert this check so we don't have to indent the happy path.

https://google.github.io/styleguide/go/decisions#indent-error-flow

internal/log/log.go
```diff
@@ -15,36 +15,42 @@
...
 	"os"
+	"time"
 
 	"github.com/GoogleCloudPlatform/cloud-sql-proxy/v2/cloudsql"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
+)
+
+const (
+	googLvlKey    = "severity"
```
enocom, line 29: These are only used in the function below. Let's just inline them.

enocom (commented):
Some minor style nits. Otherwise LGTM.

@hessjcg we should manually verify the before and after logs just to be certain this doesn't break the existing structured logging. I think it's correct, but better to double check.

danielandersson:
> Some minor style nits. Otherwise LGTM.
> 
> @hessjcg we should manually verify the before and after logs just to be certain this doesn't break the existing structured logging. I think it's correct, but better to double check.

Thanks for the review, all of your feedback should have been addressed with 0946cd0c922c7e0d24912087bc9469e497f42e78.

hessjcg (approved):
Thank you for this excellent change! Approved.

PR #2452, opened 2025-05-27
feat: add flag for ignoring bad connections

kenkania:
Port `skip-failed-instance-config` from v1 to v2. If set, this will ignore connection errors instead of exiting on startup. See more use case details in associated issue.

fixes https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/2451

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -527,6 +527,9 @@ CPU may be throttled and a background refresh cannot run reliably
 against all specified instances. If an instance is unreachable, the Proxy exits with a failure
 status code.`)
 
+	localFlags.BoolVar(&c.conf.IgnoreFailedConnections, "ignore-failed-connections", false,
+		`If set, the Proxy will not exit on startup when Cloud SQL instances are unreachable.`)
+
 	// Global and per instance flags
 	localFlags.StringVarP(&c.conf.Addr, "address", "a", "127.0.0.1",
 		"(*) Address to bind Cloud SQL instance listeners.")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -527,8 +527,9 @@ CPU may be throttled and a background refresh cannot run reliably
 against all specified instances. If an instance is unreachable, the Proxy exits with a failure
 status code.`)
 
-	localFlags.BoolVar(&c.conf.IgnoreFailedConnections, "ignore-failed-connections", false,
-		`If set, the Proxy will not exit on startup when Cloud SQL instances are unreachable.`)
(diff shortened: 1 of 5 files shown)


cmd/root.go
```diff
@@ -527,6 +527,9 @@ CPU may be throttled and a background refresh cannot run reliably
 against all specified instances. If an instance is unreachable, the Proxy exits with a failure
 status code.`)
 
+	localFlags.BoolVar(&c.conf.IgnoreFailedConnections, "ignore-failed-connections", false,
```
enocom, line 530: Let's make this skip-failed-instance-config, and update https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/migration-guide.md to mention this new flag.

cmd/root.go
```diff
@@ -527,6 +527,9 @@ CPU may be throttled and a background refresh cannot run reliably
...
+		`If set, the Proxy will not exit on startup when Cloud SQL instances are unreachable.`)
```
enocom, line 531: Let's phrase this in the positive:

"If set, the Proxy will skip any instances that cannot be found" or similar.

internal/proxy/proxy.go
```diff
@@ -267,6 +267,10 @@ type Config struct {
 	// RunConnectionTest determines whether the Proxy should attempt a connection
 	// to all specified instances to verify the network path is valid.
 	RunConnectionTest bool
+
+	// IgnoreFailedConnections determines whether the Proxy should ignore failed
+	// connections to Cloud SQL instances instead of exiting on startup.
+	IgnoreFailedConnections bool
```
enocom, line 273: SkipFailedInstanceConfig

internal/proxy/proxy.go
```diff
@@ -546,6 +550,11 @@ func NewClient(ctx context.Context, d cloudsql.Dialer, l cloudsql.Logger, conf *
 	for _, inst := range conf.Instances {
 		m, err := c.newSocketMount(ctx, conf, pc, inst)
 		if err != nil {
+			if conf.IgnoreFailedConnections {
```
enocom, line 553: A test for this block of code would be good too.

kenkania: Done

kenkania:
Thanks for the review! I've addressed all your comments, PTAL.

enocom (approved):
Thanks! @hessjcg or @kgala2 are codeowners and have the final say on this, but LGTM.

kenkania:
@kgala2 can you run the integration tests for this PR? I see the failure says: "Needs /gcbrun from a collaborator"

PR #2494, opened 2025-10-08
feat: Add --override-ip option to specify custom instance IP address

west-david:
This adds support for overriding the IP address that the proxy connects to for Cloud SQL instances. This is useful in advanced scenarios where you need to connect to a specific IP address that differs from the IPs returned by the Cloud SQL Admin API.

Features:
- Global --override-ip flag for all instances
- Per-instance override-ip query parameter
- Support for both IPv4 and IPv6 addresses
- Environment variable support (CSQL_PROXY_OVERRIDE_IP)
- Configuration file support
- Comprehensive validation and conflict detection

The implementation uses WithOneOffDialFunc to intercept the connection and redirect to the specified IP address while preserving the port and certificate verification.

Testing:
- Added 24+ unit and integration tests
- Tests cover flag parsing, query params, env vars, config files
- Validation tests for IP conflicts and invalid inputs
- All existing tests continue to pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,10 +308,32 @@ private IP, use:
 ```
 
 > [!IMPORTANT]
-> 
+>
 > The Proxy does not configure the network. You MUST ensure the Proxy can
 > reach your Cloud SQL instance, either by deploying it in a VPC that has access
 > to your Private IP instance, or by configuring Public IP.
+
+### Overriding the Instance IP Address
+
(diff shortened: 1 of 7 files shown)


kgala2:
Thanks for your contribution, @west-david! To help us understand the motivation for this change, could you please describe the problem you're aiming to solve and why the existing IP address discovery doesn't work for your use case?

PR #2540, opened 2026-01-21
test: change hardcoded ports to avoid ephemeral port collisions on linux

hessjcg:
On one of the developer linux machines, the range for ephemeral ports overlaps with
the hardcoded ports in the proxy_test.go - 50000 to 50100. The tests failed occasionally
because the hardcoded port was in use by another linux process on the same machine.
This changes the hardcoded ports to avoid all ephemeral ports.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -212,7 +212,7 @@ function write_e2e_env(){
     exit 1
   fi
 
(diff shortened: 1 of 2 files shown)


hessjcg:
This is out of date.

PR #2543, opened 2026-01-22
fix: Correctly set the credentials token source

### LINKED ISSUE #2542: Proxy fails to correctly handle Auto IAM Auth when impersonating service accounts
### Bug Description

When the `--impersonate-service-account` flag is set  and the auto-iam-auth flag is only set on the instance (not the global flag), credential impersonation will not be configured correctly for the auto-IAM instances. 

```
cloud_sql_proxy  \
  --impersonate-service-account=service_account@project.iam.gserviceaccount.com
  project:region:instance?auto-iam-auth=true
```

See: https://github.com/GoogleCloudPlatform/cloud-sql-proxy-operator/issues/719

The bug is in credentialsOpt() internal/proxy/proxy.go:355. The proxy checks if the global c.IAMAuthN flag is set, but it does not check if the auto-iam-auth flag is set on individual instances. Thus the dialers are configured without the correct token source for IAM auth.

hessjcg:
WIP: needs  tests

Fixes #2542

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -363,7 +363,20 @@ func credentialsOpt(c Config, l cloudsql.Logger) (cloudsqlconn.Option, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		var iamAuthNEnabled bool
 		if c.IAMAuthN {
+			iamAuthNEnabled = true
+		} else {
+			for _, ic := range c.Instances {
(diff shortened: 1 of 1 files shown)


mruoss:
Hi @hessjcg 
I'd like to support but I guess I can't commit to this branch. I guess I will create a separate PR...

hessjcg:
Replaced by #2544

PR #2544, opened 2026-01-23
fix: Correctly set the credentials token source and the WithIAMAuthN opt

mruoss:
Fixes https://github.com/GoogleCloudPlatform/cloud-sql-proxy/issues/2542
Replaces https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/2543

Additional to https://github.com/GoogleCloudPlatform/cloud-sql-proxy/pull/2543: Test (fails without the code change) and setting options was missing.


hessjcg:
@mruoss Thank you for the fix! Nice work.

PR #2600, opened 2026-05-21
feat: drop support for Bullseye container

panavenue:
Bullseye Debian is EOL already.

Dropping the the support from it. 

https://endoflife.date/debian


nancynh (approved):

enocom (approved):

enocom:
@panavenue Adding a link to the Debian EOL schedule would be good for people who don't know about it.

kgala2 (approved):

PR #2610, opened 2026-06-03
feat: Support Developer Edition connections

hessjcg:
This PR adds support for Developer Edition connections to the Cloud SQL Auth Proxy. It introduces the --sql-data flag and --sqldata-api-endpoint flag to enable and configure these connections. It also supports the sql-data query parameter in instance connection strings.


DeepeshDamodaran (approved):

build.sh
```diff
@@ -227,16 +244,47 @@ function write_e2e_env(){
...
+  else
+    echo "$email"
+  fi
+
+}
+
+function iam_user_mysql() {
+  # Truncate the part after the @
+  local email
+  local pguser
```
panavenue: should this be mysqluser?

cmd/root.go
```diff
@@ -606,6 +609,10 @@ only applicable to Unix sockets)`)
 		"(*) Connect to the private ip address for all instances")
 	localFlags.BoolVar(&c.conf.PSC, "psc", false,
 		"(*) Connect to the PSC endpoint for all instances")
+	localFlags.BoolVar(&c.conf.SQLDataEnabled, "sql-data", false,
+		"Enable SQL Data to tunnel through the Cloud SQL Admin API without"+
+			" needing network access to your public or private IP",
+	)
```
panavenue, line 621: Do we need some validation if user specify `sql-data` and other method ("--private-ip", "--auto-ip") at the same time?

internal/proxy/proxy.go
```diff
@@ -859,6 +877,10 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
 			np = inst.Port
 		case conf.Port != 0:
 			np = pc.nextPort()
+		case conf.SQLDataEnabled || inst.SQLDataEnabled != nil && *inst.SQLDataEnabled:
+			// Only Postgres is supported by the SqlDataService
+			// when more engines are supported, this code will need to change.
```
panavenue: Maybe add a TODO tag here?

hessjcg: Added TODO

internal/proxy/proxy.go
```diff
@@ -873,10 +895,17 @@ func (c *Client) newSocketMount(ctx context.Context, conf *Config, pc *portConfi
...
 		network = "unix"
 
-		version, err := c.dialer.EngineVersion(ctx, inst.Name)
-		if err != nil {
-			c.logger.Errorf("[%v] could not resolve instance version: %v", inst.Name, err)
-			return nil, err
+		var version string
+		switch {
+		case conf.SQLDataEnabled || inst.SQLDataEnabled != nil && *inst.SQLDataEnabled:
+			version = "POSTGRES"
```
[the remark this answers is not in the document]
hessjcg: version is really db engine type + version. It allows the proxy to choose the port to listen on that matches the database engine. For example, if the database instances is a postgres instance, then the proxy will listen for connections on port 5432 by default.

build.sh
hessjcg: Fixed.

cmd/root.go
hessjcg: Yeah, that's a good idea. I added validation and tests.

nicknovitski (approved):

